PR #6, opened 2020-12-02
Build improvements

ndw:
1. Remove old specs
2. Refactor a few ant tasks so that they will work in gradle
3. Fix a few markup errors
4. Add a top-level gradle build that publishes the specs

More work will be needed to automate publication, but this is a start.


michaelhkay (approved):
All looks good. Sorry about the markup errors: not sure how I committed a state that wasn't working. Presumably we can be a lot more radical about deleting pre-3.1 specs. The diagram is XSLT is a bit unfortunate. At one stage in the 3.0 streaming work we were making very extensive use of GraphML diagramming tools; there's just one trivial diagram left but we still have all the tooling in case it's ever needed again.

ndw:
Yes, deleting more of the old specs is a good idea. I simply removed the stuff that looked obvious.

The diagramming stuff isn't a problem, we can make that work in its full generality again if we need to. I just didn't want to spend a lot of time on that right now.

PR #12, opened 2020-12-04
Fix typo in spec reference

ndw:
Spec identifiers are case sensitive.


michaelhkay (commented):
Looks good. I wonder how that worked in 3.1? - the link is there in the published TR. Perhaps it's published on a windows box.

ndw:
That's really weird. I would have sworn you had just added that because it worked on the CI machine yesterday. Weird.

Issue #22, opened 2020-12-17
[XPath] Allowing multiple let clauses in LetExpr and for clauses in ForExpr


rhdunn:
Currently, it is possible to have multiple `SimpleLetBinding`s in a `SimpleLetClause`, but not have multiple `SimpleLetClause`s in a `LetExpr`. The same applies for `ForExpr`.

I propose that this should be possible. In other words, make the following changes to the XPath 4.0 grammar:

    [12] ForExpr ::= SimpleForClause SimpleForClause* "return" ExprSingle
    [15] LetExpr ::= SimpleLetClause SimpleLetClause* "return" ExprSingle

This way, a user can write the following in XPath 4.0:

    let $x := 1.0
    let $y := 2.0
    return $x + $y

in addition to the following in XPath 3.1 and earlier:

    let $x := 1.0, $y := 2.0
    return $x + $y

Note that this does not add any new capability to XPath (in particular, for and let clauses cannot be mixed like in XQuery) as a user can define multiple let or for clauses, it just provides them with an alternative way of expressing that like they can do in XQuery (where it is common to use a series of let clauses instead of let bindings for multiple local variables).

Kind regards,
Reece

michaelhkay:
It's increasingly difficult to justify leaving any XQuery FLWOR constructs out of XPath. The only place I'd draw the line is where the semantics can only be described in terms of "tuple of variable bindings", because (a) that requires internal machinery in an XPath engine that's not otherwise needed, and (b) the only things it enables are things that aren't remotely useful, like being able to order the sequence of tuples independently of any of its input sequences.

rhdunn:
This interacts with the `for member` syntax when that is part of the `SimpleForClause`. Specifically, it means that `SimpleForClause` cannot be expressed as:

    SimpleForClause ::= "for" "member"?  SimpleForBinding (("," | "for")  SimpleForBinding)*

as it would break the semantics of `for member` expressions due to the `for` keyword switching to sequence iteration, but not being able to easily identify that from the parse tree. The separator would also need to be `("," | "for" "member"?)` to support mixing array and sequence for expressions.

As such, this would need to be done on the `ForExpr` as originally proposed (which would continue to work if/when extended to maps):

    ForExpr ::= (SimpleForClause)+ "return" ExprSingle

rhdunn:
Fixed by PR #344.

PR #28, opened 2020-12-17
[XPath] Support multiple clauses in ForExpr and LetExpr.

rhdunn:
This implements the proposal in issue #22 to support multiple clauses in XPath ForExpr and LetExpr, with both the grammar and specification text changes.

### DIFF
--- a/etc/XP40.xml
+++ b/etc/XP40.xml
@@ -4015,6 +4015,12 @@ Components</h3>
    </nt>
    <nt xmlns:xlink="http://www.w3.org/1999/xlink"
        xmlns:e="http://www.w3.org/1999/XSL/Spec/ElementSyntax"
+       def="doc-xpath40-SimpleForClause"
+       xlink:type="simple">
+      <nt def="doc-xpath40-SimpleForClause" xlink:type="simple">SimpleForClause</nt>
+   </nt>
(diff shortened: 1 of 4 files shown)


rhdunn:
The build failure is an out of memory error:

michaelhkay (commented):
Any reason you didn't define the syntax simply as

`ForExpression ::= (SimpleForClause)+ "return" ExprSingle`

Alternatively I think it could also be

`SimpleForClause  ::=  "for"  SimpleForBinding (("," | "for")  SimpleForBinding)*`

Emphasising that there's no semantic distinction between "for" and "," here.




  | SimpleForClause | ::= | "for"  SimpleForBinding (","  SimpleForBinding)*
-- | -- | -- | --

rhdunn:
I simply forgot that `+` existed in EBNF when updating the grammar. I like your alternative proposal better as you are right about the lack of semantic distinction.

I'll update the grammar with your alternative syntax for this.

rhdunn:
This needs to be:

    ForExpression ::= (SimpleForClause)+ "return" ExprSingle

to work with the `member` keyword being on the `SimpleForClause` instead of the `SimpleForBinding`. I've also added a comment on the issue about this.

michaelhkay:
This is now superseded and can be closed with no further action.

PR #38, opened 2020-12-28
Create a schema-for-xslt40.xsd file.

rhdunn:
This uses schema-for-xslt30.xsd as the basis for adding the new elements and attributes in the current XSLT 4.0 draft.


rhdunn:
The build failure looks to be related to:
```
Error: no xpath-functions-40 known!
Error: no xpath-40 known!
Error: no xslt30 known!
```
which is unrelated to the changes I've made here -- I'm getting those errors when building the xslt-40 gradle task locally on the master branch.

rhdunn:
I've figured out what the issue with this is -- another file (xslt-first-cut.xml) is referencing the v3.0 schema file. I'll re-do this without removing the 3.0 schema file.

PR #140, opened 2022-09-14
Make agreed changes to fn:all and fn:som functions (NW testing)

ndw:


ndw:
This was just a test. Closing it so no one puts comments here.

PR #145, opened 2022-09-17
Cleaned up infrastructure

ndw:
I couldn't leave well enough alone. This "finishes" the conversion to a more modern build system, fixes a number of small editorial issues (incorrect bibliographic links, etc.), and suppresses some warnings by default.


ndw:
This won't build correctly as a PR because it changes the build infrastructure, so I'm just going to merge it.

PR #152, opened 2022-09-22
Mike's changes to fn:all and fn:some

ndw:
This is a re-application of Mike's PR onto the latest sources and build environment.


ndw:
Approved [on 27 September](https://qt4cg.org/meeting/minutes/2022/09-27.html)

PR #153, opened 2022-09-26
Explicitly mention subtypes for arrays and maps

line-o:
Reword section 3.6.4 for map and array type tests to explicitly mention sub typing and link to the relevant section.

I believe, it is important to add examples for subtypes of arrays and maps. Can someone advise how and where to add them?

map subtype example (could be added to 3.7.2.4 d.):

```xml
<note>
  <p><code>map(xs:integer, element(fruit))</code> is a subtype of <code>map(xs:anyAtomic, node())</code></p>
</note>
```

array subtype example (could be added to 3.7.2.4 i.):

```xml
<note>
  <p><code>array(xs:positiveInteger)</code> is a subtype of <code>array(xs:decimal)</code></p>
</note>
```

### DIFF
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -4707,9 +4707,13 @@ name.</p>
                      <termref
                         def="dt-map"
                         >map</termref>, and (b) every
-                     entry in <var>M</var> has a key that matches <code>K</code> and an associated value that matches <code>V</code>. For example,
+                     entry in <var>M</var> has a key that is a <termref def="dt-subtype"
+                     >subtype</termref> of <code>K</code> and an associated value that is a <termref def="dt-subtype"
+                     >subtype</termref> of <code>V</code>. For example,
                      <code>map(xs:integer, element(employee))</code> matches a map if all the keys in the map are integers, and all the associated
-                     values are <code>employee</code> elements. Note that a map (like a sequence) carries no intrinsic type information separate
+                     values are <code>employee</code> elements. Map tests are described in more detail in <specref ref="id-itemtype-subtype"/> (4.c to 4.g).</p>
+                  <p>
+                     Note that a map (like a sequence) carries no intrinsic type information separate
                      from the types of its entries, and the type of existing entries in a map does not constrain the type of new entries that can be
                      added to the map.</p>
                   <note>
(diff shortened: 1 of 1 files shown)


michaelhkay (changes requested):
Please avoid adding explicit references to "clause numbers" like 4.c - there's no easy way to preserve such references when new clauses are added. Unfortunately we don't have machinery for doing this (it would be nice to have a `<clauseref ref="..."/>`) and in its absence I would suggest introducing a defined term "This rule and the four rules that follow are referred to as the `<termdef id="dt-map-subtyping-rules"/>`", and then use a termref. Alternatively you might be able to solve the problem by making the top-level list items (for example "Conditions for functions, maps, and arrays") into section headings, making them a bit more granular, and using a specref.

line-o:
Thank you for your feedback @michaelhkay. I will do one of the two suggested edits. Do you have a preference where to add the subtype examples?

michaelhkay:
Examples should be as close as possible to the rule that they elucidate. So if you want an example of how union(A,B) is a subtype of union(A, B, C), then it belongs in 2.c.

It would be nice to have some uniformity; having examples for some of the rules and not others may be appropriate if some of the rules are more obvious than others, but it shouldn't be done purely on a whim.

michaelhkay (commented):
I have done a significant re-vamp of the section on subtyping item types, with a lot more examples and explanations; and I have added cross-references from the definitions of map, array, function,and record types to the section where the subtyping rules are defined. I think that means this PR can be dropped; but please check the new draft.

line-o:
@michaelhkay https://qt4cg.org/specifications/xquery-40/xquery-40.html#id-array-test does not mention subtypes explicitly. 

> The [AnyArrayTest](https://qt4cg.org/specifications/xquery-40/xquery-40.html#doc-xquery40-AnyArrayTest) array(*) matches any array. The [TypedArrayTest](https://qt4cg.org/specifications/xquery-40/xquery-40.html#doc-xquery40-TypedArrayTest) array(X) matches any array in which every array member matches the [SequenceType](https://qt4cg.org/specifications/xquery-40/xquery-40.html#doc-xquery40-SequenceType) X.

Does now have the *hint* at sequence types but I still believe that this should be made more explicit.
Or were your changes just not deployed yet?

michaelhkay:
I (a) added cross-reference to where subtypes are defined and explained, and (b) added a lot more explanation and examples to the section on subtyping. You can see the changes in https://qt4cg.org/pr/202/xquery-40/xpath-40.html - at the end of 3.6.4.4 there is a forwards reference to 3.7.2.8.

line-o:
@michaelhkay Thanks for the link, the changes you made are very good and exhaustive.
I still see the need to either

- update the information in the XQuery spec itself
- cross reference to the XPath spec instead - removing the duplication

michaelhkay:
The text in the XPath and XQuery specs comes from common source, so it should be identical. However, it might be that cross-reference links aren't fully working yet.

ndw:
Formatted versions of the specifications with this PR applied [are available](https://qt4cg.org/#pr-153)

michaelhkay:
This PR has been subsumed by PR #202 so I think it can now be closed with no further action.

line-o:
Thanks for taking this from here and clarify sub typing! @michaelhkay @ndw

PR #166, opened 2022-09-29
Variadicity

michaelhkay:
Two main separate but related features: (a) allow function declarations in XQuery and XSLT to define optional parameters, (b) allow static function calls (including partial function application) to take keyword arguments.


ndw:
Overtaken by #197

PR #180, opened 2022-10-05
Implement function signatures as tables with default values

ndw:


ChristianGruen:
The new layout is much easier to digest; thanks.

PR #193, opened 2022-10-09
Moved these changes into the DeltaXML pipeline; just serialize to HTML 5

ndw:


michaelhkay:
Looking forward to seeing the result.

PR #197, opened 2022-10-10
NW for MK: Variadicity (166)

ndw:
This PR is intended to be technically equivalent to #166.

### DIFF
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -664,11 +664,22 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
     </g:optional>
     <g:string>function</g:string>
     <g:ref name="_Function_QName_or_EQName" unfold="yes"/>   
-    <g:ref name="FunctionSignature"/>
+    <g:ref name="FunctionFamilySignature"/>
     <g:choice name="FunctionDeclBody">
       <g:ref name="FunctionBody" if="xquery40"/>
       <g:ref name="External"/>
     </g:choice>
+  </g:production>
+  
+  <g:production name="FunctionFamilySignature">
+    <g:string>(</g:string>
+    <g:optional name="OptionalFamilyParamList">
+      <g:ref name="FamilyParamList"/>
+    </g:optional>
+    <g:string>)</g:string>
+    <g:optional name="optionFuncFamilyType">
+      <g:ref name="TypeDeclaration"/>
+    </g:optional>
   </g:production>
   
   <g:production name="FunctionSignature">
@@ -679,6 +690,26 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
     <g:string>)</g:string>
     <g:optional name="optionFuncType">
       <g:ref name="TypeDeclaration"/>
+    </g:optional>
+  </g:production>
+  
+  <g:production name="FamilyParamList" if=" xpath40 xquery40 ">
+    <g:ref name="ParamWithDefault"/>
+    <g:zeroOrMore name="FamilyParamListTail">
+      <g:string>,</g:string>
+      <g:ref name="ParamWithDefault"/>
+    </g:zeroOrMore>
+  </g:production>
+  
+  <g:production name="ParamWithDefault" if=" xpath40 xquery40 ">
+    <g:string>$</g:string>
+    <g:ref name="_QName_or_EQName" unfold="yes"/>
+    <g:optional name="OptionalTypeDeclarationForParamWithDefault">
+      <g:ref name="TypeDeclaration"/>
+    </g:optional>
+    <g:optional name="OptionalDefaultForParam">
+      <g:string>:=</g:string>
+      <g:ref name="ExprSingle"/>
     </g:optional>
   </g:production>
 
@@ -1860,9 +1891,9 @@ ErrorVal ::= "$" VarName
   </g:production>
   
   <g:production name="KeywordArgument" if="xpath40 xquery40 xslt40-patterns">
-    <g:ref name="NCName"/>
-    <g:string>:=</g:string>
-    <g:ref name="ExprSingle"/> 
+    <g:ref name="EQName"/>
+    <g:string>:</g:string>
+    <g:ref name="Argument"/> 
   </g:production>
 
   <g:production name="PredicateList" condition="&gt; 0">
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -2484,11 +2484,7 @@ ErrorVal ::= "$" VarName
   <!-- [ start Types + Tests -->
 
   <g:production name="SingleType" >
-    <g:ref name="SimpleTypeName" if=" xslt40-patterns"/>
-    <g:choice if="xpath40 xquery40">
-      <g:ref name="SimpleTypeName"/>
-      <g:ref name="LocalUnionType"/>
-    </g:choice>
+    <g:ref name="SimpleTypeName"/>
     <g:optional name="OptionalOccurrenceIndicator">
       <g:string process-value="yes">?</g:string>
     </g:optional>
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -664,17 +664,17 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
     </g:optional>
     <g:string>function</g:string>
     <g:ref name="_Function_QName_or_EQName" unfold="yes"/>   
-    <g:ref name="FunctionFamilySignature"/>
+    <g:ref name="FunctionSignatureWithDefaults"/>
     <g:choice name="FunctionDeclBody">
       <g:ref name="FunctionBody" if="xquery40"/>
       <g:ref name="External"/>
     </g:choice>
   </g:production>
   
-  <g:production name="FunctionFamilySignature">
+  <g:production name="FunctionSignatureWithDefaults">
     <g:string>(</g:string>
     <g:optional name="OptionalFamilyParamList">
-      <g:ref name="FamilyParamList"/>
+      <g:ref name="ParamListWithDefaults"/>
     </g:optional>
     <g:string>)</g:string>
     <g:optional name="optionFuncFamilyType">
@@ -693,7 +693,7 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
     </g:optional>
   </g:production>
   
-  <g:production name="FamilyParamList" if=" xpath40 xquery40 ">
+  <g:production name="ParamListWithDefaults" if=" xpath40 xquery40 ">
     <g:ref name="ParamWithDefault"/>
     <g:zeroOrMore name="FamilyParamListTail">
       <g:string>,</g:string>
@@ -1892,7 +1892,7 @@ ErrorVal ::= "$" VarName
   
   <g:production name="KeywordArgument" if="xpath40 xquery40 xslt40-patterns">
     <g:ref name="EQName"/>
-    <g:string>:</g:string>
+    <g:string>:=</g:string>
     <g:ref name="Argument"/> 
   </g:production>
 
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -707,8 +707,7 @@ inferred by static type inference as discussed in  <specref
                         >expanded QName</termref> and the overlapping
                      arity ranges (even if the signatures are consistent).</phrase></p>
  
-    <p>The entries in the known static functions define the set of 
-     declared functions that are available to be called from a 
+    <p>The declared functions are available to be called from a 
     <termref
                         def="dt-static-function-call"
                         >static function call</termref>,
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -644,7 +644,7 @@ inferred by static type inference as discussed in  <specref
                      is mapped to an <code>ItemType</code> in the <termref def="dt-item-type-aliases"/> of
                      the <termref def="dt-static-context"/>.</termdef>
                   </p>
-                  <p>Item type aliases allow frequently-used item types, especially complex item types such as
+                  <p>Item type aliases allow frequently used item types, especially complex item types such as
                   record types, to be given simple names, so that the definition of the type is not repeated
                   every time it is used.</p>
                   <note>
@@ -659,8 +659,8 @@ inferred by static type inference as discussed in  <specref
                   <p>
                     
                         <termdef id="dt-statically-known-function-definitions"
-                           term="statically-known function definitions">
-                           <term>Statically-known function definitions.</term> This is a set of 
+                           term="statically known function definitions">
+                           <term>Statically known function definitions.</term> This is a set of 
                            <termref def="dt-function-definition">function definitions</termref>.</termdef>
                   </p>
                   <p><termdef id="dt-function-definition" term="function definition">A <term>function definition</term>
@@ -681,7 +681,7 @@ inferred by static type inference as discussed in  <specref
                         <ulist>
                            <item><p>a parameter name (an <termref def="dt-expanded-qname"/>)</p></item>
                            <item><p>a required type (a <termref def="dt-sequence-type"/>)</p></item>
-                           <item><p>a default value (a <termref def="dt-sequence-type"/>)</p></item>
+                           <item><p>a default value expression (an <term>expression</term>: see <specref ref="id-expressions"/>)</p></item>
                         </ulist>
                      </item>
                      <item><p>A return type (a <termref def="dt-sequence-type"/>)</p></item>
@@ -693,16 +693,29 @@ inferred by static type inference as discussed in  <specref
                   
                   <p>The names of the parameters must be distinct.</p>
                   
-                  <p><termdef id="dt-arity-range" term="arity range">A <termref def="dt-function-definition"/> has an <term>arity range</term>
+                  <p><termdef id="dt-arity-range" term="arity range">A <termref def="dt-function-definition"/> has an <term>arity range</term>,
                      which is a range of consecutive non-negative integers. If the function definition has <var>M</var> required parameters
                      and <var>N</var> optional parameters, then its arity range is from <var>M</var> to <var>M</var>+<var>N</var>
                      inclusive.</termdef></p>
                   <p>The static context may contain several <termref def="dt-function-definition">function definitions</termref> with the
                      same name, but the <termref def="dt-arity-range">arity ranges</termref> of two such function definitions must not 
-                     overlap. For example, it is acceptable to have a function definition with two required parameters and no optional
-                     parameters, and another function definition (with the same name) that has three required parameters and one optional
-                     parameter. It is not acceptable to have a function definition with one required parameter and another (with the same name)
-                     having three optional parameters.</p>
+                     overlap. For example, if two function definitions <var>A</var> and <var>B</var> have the same function name, then:</p>
+                  <ulist>
+                     <item><p>It is acceptable for <var>A</var> to have two required parameters and no optional
+                        parameters, while <var>B</var> has three required parameters and one optional
+                        parameter.</p></item>
+                     <item><p>It is not acceptable for <var>A</var> to have one required parameter while <var>B</var>
+                        has three optional parameters.</p></item>
+                  </ulist>
+                  
+                  <note><p>Implementations must ensure that no two <termref def="dt-function-definition">function definitions</termref> 
+                     have the same <termref def="dt-expanded-qname">expanded QName</termref> and overlapping
+                     arity ranges (even if the signatures are consistent).</p>
+                  
+                  <p>XQuery and XSLT enforce this rule by defining a static error if the rule is violated; but further constraints
+                     may be needed if an API allows external functions to be added to the static context.</p></note>
+                  
+                 
                   
                
                   <p diff="chg" at="A">Function definitions in the static context may originate in a number of ways, for example they may be:</p>
@@ -721,11 +734,7 @@ inferred by static type inference as discussed in  <specref
                   the functions specified in <bibref ref="xpath-functions-40"/>.</termdef></p>
                   
                   
-                  <p>Implementations must ensure that no two declared functions have the same <termref
-                        def="dt-expanded-qname"
-                        >expanded QName</termref> and overlapping
-                     arity ranges (even if the signatures are consistent).</p>
- 
+                  
     <p>The <termref def="dt-function-definition">function definitions</termref> are available for reference from a 
     <termref
                         def="dt-static-function-call"
@@ -738,22 +747,7 @@ inferred by static type inference as discussed in  <specref
 </p>
 
 
-                  <!--<p>The declared functions typically 
-                     include functions from a variety of sources, including
-                     the <phrase role="xpath"><termref def="dt-built-in-function">built-in functions</termref>.</phrase>
-                         <phrase role="xquery"><termref
-                        def="dt-built-in-function"
-                        >built-in functions</termref>, functions declared in the current module (see <specref
-                        ref="FunctionDeclns"/>), module imports (see <specref ref="id-module-import"
-                        />), <termref def="dt-constructor-function"
-                        >constructor functions</termref> for user-defined types (see  <specref
-                        ref="id-constructor-functions"
-                        />), and functions provided by an
-implementation or via an implementation-defined API (see <specref
-                        ref="id-xq-static-context-components"/>).</phrase>
-
-
-                  </p>-->
+                  
 
                </item>
 
@@ -1284,20 +1278,12 @@ the number of items in the sequence obtained by evaluating <code
                <item>
                   <p diff="chg" at="variadicity">
                      <termdef term="dynamically known function definitions" id="dt-dynamically-known-function-definitions">
-                        <term>Dynamically-known function definitions</term>.
+                        <term>Dynamically known function definitions</term>.
         This is a set of <termref def="dt-function-definition">function definitions</termref>. It includes the
                         <termref def="dt-statically-known-function-definitions"/> as a subset, but may include
                         other function definitions that are not known statically.
       </termdef></p>
-                  <!--<ulist>
-                     <item><p>For each <termref def="dt-function-definition"/> in the <termref def="dt-in-scope-function-definitions"/>,
-                     and for each integer <code>A</code> in the <termref def="dt-arity-range"/> of that function definition,
-                        a <termref def="dt-function-item">function item</termref> equivalent to that
-                     obtained using the <termref def="dt-named-function-ref"/> <code>N#A</code>, where <var>N</var>
-                     is the name of the declared function.</p></item>
-                     <item><p>Optionally, an <termref def="dt-implementation-defined"/> set of additional named
-                        <termref def="dt-function-item">function items</termref>.</p></item>
-                  </ulist>-->
+     
                   <p>The function definitions in the dynamic context are used primarily by the <code>fn:function-lookup</code>
                   function.</p>
                   <p>If two function definitions in the <termref def="dt-dynamically-known-function-definitions"/> have the same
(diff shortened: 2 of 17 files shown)


specifications/xslt-40/src/xslt.xml
```diff
@@ -534,6 +534,25 @@
                   </p>
 
 
+               </item>
+               <item>
+                  <p>
+                     <termdef id="dt-declared-function" term="declared function">The term
+                        <term>declared function</term> is defined in <xspecref spec="XP40" ref="static_context"/>. 
+                        In refers to the definition of a function that can be called from within an XPath
```
Arithmeticus, line 542: In refers -> It refers

specifications/xquery-40/src/expressions.xml
```diff
@@ -438,42 +438,23 @@ whitespace normalized according to the rules for the <code>xs:anyURI</code> type
...
-                     <termdef id="dt-function-name-resolver" term="function name resolver">
-                        <term>Function name resolver.</term> This is an algorithm that takes as input
-                        a lexical QName and arity, and delivers as its result an expanded QName.</termdef> 
-                        The expanded QName and arity must uniquely identify a function signature in the
-                        <termref def="dt-known-func-signatures"/>, or a static error is reported.
-                        This algorithm
-                        is used when function names are resolved statically, specifically when they
-                        appear in static function calls or function references. The 
+                     <termdef id="dt-default-function-namespace" term="default function namespace">
+                        <term>Default function namespace.</term> This is an algorithm that takes as input
```
Arithmeticus, line 442: A namespace is not an algorithm. 
Suggested: "This is the namespace URI returned by the algorithm that...."
Is the algorithm anywhere described or defined? Xref.

specifications/xquery-40/src/expressions.xml
```diff
@@ -646,43 +627,100 @@ inferred by static type inference as discussed in  <specref
...
-                        term="statically known function signatures">
-                        <term>Statically known function signatures.</term>
-        This is a mapping from (expanded QName, arity) to 
-        <xtermref
-                           spec="DM31" ref="dt-signature"
-                        />. 
-    </termdef>
-    The entries in this mapping define the set of 
-     functions that are available to be called from a 
+                     <termdef id="dt-declared-functions"
```
Arithmeticus, line 632: Is there precedence in the specs for two paired definitions, one for the plural form of the noun and the other for the singular? On the face of it, the formula "Xs: the set of Xs" is circular, and should be rephrased in terms of the singular version, e.g., "Xs: the set of every X". But why is the plural version needed? 
Further: "a set" -- i.e., arbitrary? Or is it every declared function declared from the static context?

Arithmeticus (changes requested):
The proposal is excellent. I have flagged points that touch primarily on clarity, structure, and overlap. This is also my first attempt at reviewing changes to the specs, and I have focused primarily on questions of consistency. You may find a number of misfires on my part. Just tell my why and I'll try not to repeat the error in future reviews.

specifications/grammar-40/xpath-grammar.xml
```diff
@@ -681,6 +692,26 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
...
+  </g:production>
+  
+  <g:production name="ParamWithDefault" if=" xpath40 xquery40 ">
+    <g:string>$</g:string>
+    <g:ref name="_QName_or_EQName" unfold="yes"/>
+    <g:optional name="OptionalTypeDeclarationForParamWithDefault">
+      <g:ref name="TypeDeclaration"/>
+    </g:optional>
+    <g:optional name="OptionalDefaultForParam">
+      <g:string>:=</g:string>
```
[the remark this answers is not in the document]
rhdunn: > In function calls, the = in := has been deleted (KeywordArgument). Will the discrepancy between : and := confuse users?

This is using the $variable syntax to declare a variable (parameter in this case), so using := here is consistent with other variable declarations and bindings.

Using : in the function calls, that is not using the $variable syntax to assign the parameter (which would conflict with passing a variable of that name to the function). Instead, it is using a syntax akin to map construction, which uses the key : value syntax.

If/when function argument binding extends to map keys (such as for an option map), then the parameter and map key assignment is uniform and consistent with map construction.

specifications/grammar-40/xpath-grammar.xml
```diff
@@ -1860,9 +1891,9 @@ ErrorVal ::= "$" VarName
   </g:production>
   
   <g:production name="KeywordArgument" if="xpath40 xquery40 xslt40-patterns">
-    <g:ref name="NCName"/>
-    <g:string>:=</g:string>
```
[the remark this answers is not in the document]
rhdunn: See my comment on the parameters. -- This is consistent with the way map construction works.

To me, `:=` is for variable assignment which would require a `$name` variable syntax. That would be ambiguous with passing a variable of that name to the function. It would also be less clear as to what is happening, making the code harder to read. Using `:=` with `$name` (if we used `$name` instead of just `name`) would break the Scripting Extension assignment expressions which allow you to do things like `$name := 123` to update the value of a variable.

specifications/xquery-40/src/expressions.xml
rhdunn: We haven't discussed the capability of the default function namespace logic to support this additional complexity. As such, I would expect this to be the same text as the XQuery/XPath specs:
> [Definition: Default function namespace. This is a namespace URI or [absent](https://www.w3.org/TR/xpath-datamodel-31/#dt-absent)DM31. The namespace URI, if present, is used for any unprefixed QName appearing in a position where a function name is expected.] The URI value is whitespace normalized according to the rules for the xs:anyURI type in [Section 3.2.17 anyURI ](https://www.w3.org/TR/xmlschema-2/#anyURI)XS1-2 or [Section 3.3.17 anyURI ](https://www.w3.org/TR/xmlschema11-2/#anyURI)XS11-2.

specifications/xquery-40/src/expressions.xml
```diff
@@ -646,43 +627,100 @@ inferred by static type inference as discussed in  <specref
...
+                        term="declared functions">
+                        <term>Declared functions.</term> This is a set of 
+                        <termref def="dt-declared-function">declared functions</termref>.</termdef></p>
+                  
+                  <p><termdef id="dt-declared-function" term="declared function">A <term>declared function</term>
+                     contains the information needed to evaluate a static function call. It may be derived
+                     from an actual declaration in a host language such as XQuery or XSLT, or it may represent
+                     a built-in function, or it may represent information registered with the 
+                     processor using some <termref def="dt-implementation-defined"/> API.</termdef></p>
```
rhdunn, line 641: > or it may represent information registered with the processor using some API.

I find this part slightly confusing, as in order to define a function via an API you are not just registering information, you are defining the function. Maybe change this to:

> or it may be registered with the processor using some API.

specifications/xquery-40/src/expressions.xml
```diff
@@ -646,43 +627,100 @@ inferred by static type inference as discussed in  <specref
...
+                  
+                  <p>The properties of a <termref def="dt-declared-function"/> include:</p>
+                  
+                  <ulist>
+                     <item><p>The function name, which is an <termref def="dt-expanded-qname"/>.</p></item>
+                     <item><p>A (possibly empty) list of required parameters, each having:</p>
+                        <ulist>
+                           <item><p>a parameter name (an <termref def="dt-expanded-qname"/>)</p></item>
```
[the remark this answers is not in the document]
rhdunn: > delete "parameter"

I think keeping parameter here makes sense as it is qualifying what the name is relating to.

specifications/xslt-40/src/xslt.xml
```diff
@@ -16031,8 +16033,11 @@ and <code>version="1.0"</code> otherwise.</p>
                      <rfc2119>must not</rfc2119> have a <code>select</code> attribute.</p>
 
                <p diff="add" at="A">The static context for evaluating the default value depends on where the
-               relevant expression appears in the stylesheet, in the usual way. The dynamic context is different
-               for different kinds of parameter:</p>
+               relevant expression appears in the stylesheet, in the usual way. Note however that
+               for <elcode>xsl:param</elcode> elements defining <termref def="dt-function-parameter">function parameters</termref>,
+                  the static context does not include variables bound in preceding-sibling <elcode>xsl:param</elcode> elements.</p>
```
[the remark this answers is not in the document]
rhdunn: > edit?: preceding-sibling > sibling

This is correct. -- It is saying that if a previous xsl:param has a default expression that defines some variables, then those variables are not in the static context/scope of the current xsl:param.

rhdunn (changes requested):

specifications/xslt-40/src/xslt.xml
michaelhkay: Yes to "It refers". But it (the term) doesn't refer to a "function" in the sense of "XDM function" or "function item". In 3.1 it was possible to gloss over the distinction, but this is no longer the case; the things in the static context ("declared function" objects) represent families of functions rather than single functions. (But there's a lot of legacy mess in this area and I haven't cleaned it up 100%. The 3.1 spec called this "function signatures" though it clearly contained more than the signature. And then it didn't really use it anyway, the binding rules for static function calls actually went to the dynamic context....

specifications/xquery-40/src/expressions.xml
michaelhkay: Agreed, this probably isn't ideal. Altogether, the terminology for the parts of the static and dynamic context is far from ideal, and it's difficult to sort out. Many of these "parts" are described using plural forms of the things they contain and I followed that convention, but it's not a good convention.

specifications/xquery-40/src/expressions.xml
```diff
@@ -646,43 +627,100 @@ inferred by static type inference as discussed in  <specref
...
+                     </item>
+                     <item><p>A (possibly empty) list of optional parameters, each having:</p>
+                        <ulist>
+                           <item><p>a parameter name (an <termref def="dt-expanded-qname"/>)</p></item>
+                           <item><p>a required type (a <termref def="dt-sequence-type"/>)</p></item>
+                           <item><p>a default value (a <termref def="dt-sequence-type"/>)</p></item>
+                        </ulist>
+                     </item>
+                     <item><p>A return type (a <termref def="dt-sequence-type"/>)</p></item>
+                     <item><p>A (possibly empty) set of <term>function annotations</term></p>
```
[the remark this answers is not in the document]
michaelhkay: XDM says that functions have a set of annotations, but in XPath it's always an empty set.

specifications/xslt-40/src/xslt.xml
michaelhkay: Thanks for this very detailed review. As I work through it, I get to the point where it's hard to respond to every detailed point because changes that need to be made to one comment tend to invalidate subsequent comments. So I think the best approach is probably to attempt a second draft that takes into account as many of the comments as possible, focussing on those that are independent.

cmsmcq:
Three minor queries about declared functions:

* In 2.1.1, the replacement of the definition of *statically known function signatures* with definitions of *declared function* and *declared functions* has as a consequence that while the status quo explicitly says that those function signatures are available for reference from a named function reference, the new text does not.  Is that covered elsewhere?  If not, is it likely to matter?

* Working through the proposal I find myself increasingly uncomfortable that the denotation of *declared function* is not a function.  I wonder if the term *function declaration* would work better?  Or *function family*?

* I note that the description of the default function namespace allows implementation-defined algorithms for searching multiple namespaces and resolving conflicts; this makes me wonder whether we have a compelling reason to forbid such algorithms for resolving overlapping arity ranges.  (I don't think I have a preference, beyond adhering to the goose/gander principle.)

michaelhkay:
@cmsmcq 

> In 2.1.1, the replacement of the definition of statically known function signatures with definitions of declared function and declared functions has as a consequence that while the status quo explicitly says that those function signatures are available for reference from a named function reference, the new text does not. Is that covered elsewhere? If not, is it likely to matter?

I think we used to say it twice, we now only say it once. It's there in 4.4.2.3.

> Working through the proposal I find myself increasingly uncomfortable that the denotation of declared function is not a function. I wonder if the term function declaration would work better? Or function family?

Yes, me too. I was trying to minimise change to the 3.1 spec, but I'm increasingly persuaded that the 3.1 presentation which tries to maintain that there are two kinds of function, static functions and dynamic functions, is wrong. I'm now proposing to call the things in the static context "function definitions". I don't want to call them "function declarations" because that's a specific piece of syntax in XQuery, but I think that the name "function definitions" gets us away from the idea that a declared function is a kind of function.

> I note that the description of the default function namespace allows implementation-defined algorithms for searching multiple namespaces and resolving conflicts; this makes me wonder whether we have a compelling reason to forbid such algorithms for resolving overlapping arity ranges. (I don't think I have a preference, beyond adhering to the goose/gander principle.)

Unfortunately, the base text I was starting from included a speculative enhancement for resolving unprefixed function names, which is unrelated to this proposal. I'm trying to roll that back, but the changes muddy the diff.

specifications/xquery-40/src/expressions.xml
```diff
@@ -7650,51 +7684,58 @@ return $a("A")]]></eg>
...
+                  beyond the <var>A</var>'th parameter, to their respective default values. The evaluation of the expressions
+                  that define these default values occurs in the dynamic context of the named function reference.</p></item>
+               </ulist>
             </p>
+            
+            <note diff="add" at="variadicity"><p>Consider the declared function <code>fn:format-date</code>
+            with an arity range of 2 to 5. The named function reference <code>fn:format-date#3</code>
+            returns a (dynamic) function whose three arguments correspond to the first three arguments
+            of <code>fn:format-date</code>; the remaining two arguments will take their default values.
+               To obtain a 3-argument function that binds to arguments 1, 2, and 5 of <code>fn:format-date</code>,
```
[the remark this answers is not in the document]
michaelhkay: No, what is obtained is an arity-3 function item with the default values for the other two arguments present in its closure.

specifications/grammar-40/xpath-grammar.xml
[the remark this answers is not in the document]
michaelhkay: We decided to use ":=".

ndw:
Formatted versions of the specifications with this PR applied [are available](https://qt4cg.org/#pr-197)

cmsmcq:
Some incomplete and unsystematic notes:

* Is it intended that in section 2.2.1 Static Context of XPath, in the description of **function definition**, both *required type* and *default value* are glossed as "(a sequence type)"?  It seems unexpected enough that I suspect a copy/paste error.
* I think the sentence "It is not acceptable to have a function definition with one required parameter and another (with the same name) having three optional parameters", would do its work better without the parenthesis marks.
* A propos the sentence "Implementations must ensure that no two declared functions have the same [expanded QName](https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-expanded-qname) and overlapping arity ranges (even if the signatures are consistent)."
  * The term "declared function" seems important and is used in several places, but is nowhere defined in the XPath spec.  Does it mean a function definition in the statically known function definitions? ... in the dynamically known function definitions?
  * The requirement that no two declared functions have certain properties entails that we need a clean story about how to tell whether a declared function *f1* is the same declared function as a declared function *f2*, or they are different functions.  (I suspect this need is particularly acute when there is some possibility that the same module definition [for some definition of "same"] might by accident or design be referenced twice under different URIs, but the need is probably more general.  I don't think we want complicated rules, just a clear one.)  
* In section 2.3.4 Consistency constraints, the last bullet item appears to be redundant.  It says "Every [function definition](https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-function-definition) in the [statically-known function definitions](https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-statically-known-function-definitions) of the [static context](https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-static-context) must also be present in the [dynamically known function definitions](https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-dynamically-known-function-definitions) of the [dynamic context](https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-dynamic-context)", which follows from the statement in the definition of the *dynamically known function definitions* property that "It includes the [statically-known function definitions](https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-statically-known-function-definitions) as a subset, but may include other function definitions that are not known statically."  For economy of expression, one of these should probably be deleted; alternatively, one should be marked as a redundant reminder of the normative statement made elsewhere.  (I would make the definition be the normative statement and drop the bullet item or label it a consequence of the definition.)
* In 4.4.1 Function definitions, the sentence "The function is evaluated by executing its implementation, ..." makes my blood run cold.  I do not now have a proposal for alternative wording, but the sudden irruption of an imperative, procedural description into what is otherwise a declarative specification seems out of place.  Perhaps the change could just be to "The function is evaluated using a dynamic context that provides values for all the declared parameters, initialized as described in [4.4.1.2 Evaluating Static Function Calls](https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#id-eval-static-function-call) below."
* In section 4.4.1.1 Static Function Call Syntax, missing text (bad hyperlink?) in the sentence "If the EQName in a static function call is an unprefixed [lexical QName](https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-qname), it is expanded using the [default function namespace](https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-default-function-namespace) in the ."  In the next paragraph, missing whitespace in "in the [static context](https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-static-context)using the rules".  (Other instances of missing whitespace immediately following a hyperlink are also visible, but I won't try to list them all.)
* Section 4.4.1.2 Evaluating Static Function Calls says, inter alia: "Any parameters that have not been associated with any argument expression by applying the above rules (which are necessarily optional parameters) are then associated with the default expression for the relevant parameter in the declared function."  
  It is not clear to me that the rules given earlier in the document guarantee that parameters not yet associated with any argument expression will be optional parameters.  (I found this sentence while looking for the place where the text would say that it is an error if at this point any required parameter does not yet have an argument expression.)
  In particular, consider a function definition named *my:foo* with two required and two optional parameters named *A*, *B*, *C*, and *D*.  I think the description of function definitions in section 2.2.1 Static Context tells us that its arity range is two to four, inclusive.  Consider then a static function call `my:foo(23, C := 42, D := "skiddoo")`, with one position and two keyword parameters.  I think item 1 in section 4.4.1.2 tells us the required arity is three.  That falls into the range two to four inclusive, so the call to `my:foo` is bound to the function definition for *my:foo*.  I think item 2 in the same section then tells us that *A* is bound to the expression `23`, *C* to `42`, and *D* to `"skiddoo"`.  
  Item 2 then tells me that any parameter that doesn't have an expression (here *B*) is "necessarily" an optional parameter.  But *B* is not an optional parameter.  It is false to say that it is, and it is doubly false to say that it is *necessarily* an optional parameter.  
  Perhaps I need to say that I take "necessarily" in this sentence to mean that the claim follows as a logical consequence of the rules already given.  If "necessarily" is meant here to mean that the state of affairs described is a pre-condition of error-free evaluation, then I think we would do better to find another way of making the point.
* In item 3 of the same list, some unmarked legacy text says "it is not required that an argument be evaluated if the function can evaluate its body without evaluating that argument."  I think this would be clearer if rephrased as "it is not required that an argument be evaluated if the function body can be evaluated without evaluating that argument."  Or "it is not required that an argument be evaluated if the function body can be evaluated it."

specifications/xquery-40/src/expressions.xml
```diff
@@ -646,43 +655,91 @@ inferred by static type inference as discussed in  <specref
...
+                     <item><p>A return type (a <termref def="dt-sequence-type"/>)</p></item>
+                     <item><p>A (possibly empty) set of <term>function annotations</term></p>
+                        <p role="xquery">In XQuery, function annotations are described in <specref ref="id-annotations"/>.</p></item>
+                     <item><p>An implementation. The function implementation contains the logic that enables the function
+                     result to be computed from the supplied parameters and information in the dynamic context.</p></item>
+                  </ulist>
+                  
+                  <p>The names of the parameters must be distinct.</p>
+                  
+                  <p><termdef id="dt-arity-range" term="arity range">A <termref def="dt-function-definition"/> has an <term>arity range</term>
```
Arithmeticus, line 696: </term> -> </term>, (in this context "which" is non-restrictive)

specifications/xquery-40/src/expressions.xml
```diff
@@ -646,43 +655,91 @@ inferred by static type inference as discussed in  <specref
...
+                     language such as Java or Python.</p></item>
+                  </ulist>
+                  
+                  <p diff="add" at="B"><termdef id="dt-built-in-function" term="built-in function"><term>Built-in functions</term>
+                  are <termref def="dt-function-definition">function definitions</termref> that are always
+                  present in the static context by virtue of rules in the host language; they will typically include
+                  the functions specified in <bibref ref="xpath-functions-40"/>.</termdef></p>
+                  
+                  
+                  <p>Implementations must ensure that no two declared functions have the same <termref
```
Arithmeticus, line 724: functions -> function definitions

specifications/xquery-40/src/expressions.xml
```diff
@@ -6543,35 +6589,43 @@ At evaluation time, the value of a variable reference is the value to which the
...
-               or as user-defined functions declared in the query prolog.</p></item>
-            <item><p>Dynamic functions are XDM items: values that can be bound to variables, passed as
-            arguments, returned as function results, and generally manipulated in the same way as other
-            XDM values.</p></item>
+         <ulist diff="chg" at="B">
+            <item><p>A <termref def="dt-function-definition"/> contains information
+               about a family of functions with the same name and a defined arity range. These functions
+               are in most cases known statically (they appear in the <termref def="dt-statically-known-function-definitions"/>), 
+               but there may be further function definitions
+               that are only known dynamically (appearing in the <termref def="dt-dynamically-known-function-definitions"/>).</p>
```
Arithmeticus, line 6599: only known -> known only

specifications/xquery-40/src/expressions.xml
```diff
@@ -646,43 +655,91 @@ inferred by static type inference as discussed in  <specref
...
-        <xtermref
-                           spec="DM31" ref="dt-signature"
-                        />. 
-    </termdef>
-    The entries in this mapping define the set of 
-     functions that are available to be called from a 
+                    
+                        <termdef id="dt-statically-known-function-definitions"
+                           term="statically-known function definitions">
+                           <term>Statically-known function definitions.</term> This is a set of
```
Arithmeticus, line 663: Statically-known -> Statically known
The non-hyphenated form predominates 5:1, and is (in my editorial circles) to be preferred in constructions of [-ly adverb] [adj.] [noun].

Arithmeticus (approved):
The revision is much clearer, and the simpler approach is to be preferred. Comments address only stylistic issues.

michaelhkay:
Thanks for the comments.



On 25 Oct 2022, at 02:55, C. M. Sperberg-McQueen ***@***.******@***.***>> wrote:



Some incomplete and unsystematic notes:

  *   Is it intended that in section 2.2.1 Static Context of XPath, in the description of function definition, both required type and default value are glossed as "(a sequence type)"? It seems unexpected enough that I suspect a copy/paste error.

No, that's a typo.Thanks.

  *   I think the sentence "It is not acceptable to have a function definition with one required parameter and another (with the same name) having three optional parameters", would do its work better without the parenthesis marks.

I'll try to find a more elegant way of phrasing it.

  *   A propos the sentence "Implementations must ensure that no two declared functions have the same expanded QName<https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-expanded-qname> and overlapping arity ranges (even if the signatures are consistent)."
     *   The term "declared function" seems important and is used in several places, but is nowhere defined in the XPath spec. Does it mean a function definition in the statically known function definitions? ... in the dynamically known function definitions?

This should have been replaced by "function definitions" in the latest draft. The occurrences that remain somehow escaped a search of the XML content but are revealed by searching the HTML.

  *
     *   The requirement that no two declared functions have certain properties entails that we need a clean story about how to tell whether a declared function f1 is the same declared function as a declared function f2, or they are different functions. (I suspect this need is particularly acute when there is some possibility that the same module definition [for some definition of "same"] might by accident or design be referenced twice under different URIs, but the need is probably more general. I don't think we want complicated rules, just a clear one.)

The rules here go back to how XQuery and XSLT populate the static context, and are different between XQuery and XSLT.

In XQuery the rules are in 5.18: a function declaration in the query prolog, or a public function declaration in an imported module, cause a function definition to be added to the static context. I wish the rules on "import module" were clearer about module identity, but I don't think this is the place to fix them.

In XSLT the rules are more complex because of import precedence and packages, but I think the proposal spells them out in sufficient detail.

  *
  *   In section 2.3.4 Consistency constraints, the last bullet item appears to be redundant.

I suspect it's not the only thing that's said twice. I'll try to improve it.

  *   In 4.4.1 Function definitions, the sentence "The function is evaluated by executing its implementation, ..." makes my blood run cold.

Me too. I've always hated this. It's straight out of the 3.1 spec. I'll see if I can find something better.

  *   In section 4.4.1.1 Static Function Call Syntax, missing text (bad hyperlink?) in the sentence "If the EQName in a static function call is an unprefixed lexical QName<https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-qname>, it is expanded using the default function namespace<https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-default-function-namespace> in the ." In the next paragraph, missing whitespace in "in the static context<https://qt4cg.org/pr/197/xquery-40/xpath-40-autodiff.html#dt-static-context>using the rules". (Other instances of missing whitespace immediately following a hyperlink are also visible, but I won't try to list them all.)

Noted. Some of the whitespace problems seem to be associated with a switch from roman to italic text, and may be a rendition issue.

  *   Section 4.4.1.2 Evaluating Static Function Calls says, inter alia: "Any parameters that have not been associated with any argument expression by applying the above rules (which are necessarily optional parameters) are then associated with the default expression for the relevant parameter in the declared function."
It is not clear to me that the rules given earlier in the document guarantee that parameters not yet associated with any argument expression will be optional parameters.

Yes, I had problems convincing myself of this too. I thought I had proved it to myself, but your counter-example seems to show I was wrong.

  *   In item 3 of the same list, some unmarked legacy text says "it is not required that an argument be evaluated if the function can evaluate its body without evaluating that argument." I think this would be clearer if rephrased as "it is not required that an argument be evaluated if the function body can be evaluated without evaluating that argument."

Yes, that's definitely better.

  *   —

specifications/xquery-40/src/expressions.xml
michaelhkay: Yes, again I've gradually learned that hyphenating adverb-adjective phrases seems to be much more the norm in BrE than in AmE. (The Oxford Manual of Style says use no hyphen if the adverb ends in "-ly", but use one otherwise (so "a well-known and happily married couple". Bizarre.)

specifications/xquery-40/src/expressions.xml
michaelhkay: yes.

specifications/xquery-40/src/expressions.xml
michaelhkay: Yes, in my years of doing this job I've learned that Americans are much fussier about which/that distinctions than Brits.

Arithmeticus: Here it's just the comma, not the which/that distinction, which the Cambridge Grammar of English Language has convinced me of as being unsound, _pace_ Garner.

rhdunn (commented):
Overall, this looks good. I've left some comments, and have comments from my previous review.

specifications/xslt-40/src/xslt.xml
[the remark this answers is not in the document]
rhdunn: Q: Should we also call this out in the XPath and XQuery specs? -- Parameters declared before the current parameter's default expression are not included in the static context of that expression.

specifications/xquery-40/src/expressions.xml
```diff
@@ -7004,88 +7064,53 @@ At evaluation time, the value of a variable reference is the value to which the
...
+                        in the declared function are concatenated into a single list, in order.</p></item>
+                        <item><p>Any keyword arguments in the function call are then matched against
+                           the names of the parameters (whether required or optional) in the declared function. Each keyword must match
+                        the name of a declared parameter, and this must be one that has not already
+                        been associated with a positional argument. [TODO: error condition]. 
+                        The expression supplied in the keyword
+                        argument is then associated with the corresponding parameter.</p></item>
+                        <item><p>Any parameters that have not been associated with any argument expression
+                        by applying the above rules (which are necessarily optional parameters) are then associated 
+                        with the default expression for the relevant parameter in the declared function.</p></item>
```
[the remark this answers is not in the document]
rhdunn: This is now resolved.

ndw:
Approved at [meeting 011, 15 November 2022](https://qt4cg.org/meeting/minutes/2022/11-15.html)

PR #198, opened 2022-10-10
NW for MK: fn:op (173)

ndw:
This PR is intended to be technically equivalent to #173.


ChristianGruen (approved):
Looks good

ndw:
Accepted at the CG meeting of [11 October 2022](https://qt4cg.org/meeting/minutes/2022/10-11.html).

PR #199, opened 2022-10-10
NW for MK: Items before, etc. (177)

ndw:
This PR is intended to be technically equivalent to #177.

### DIFF
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -2453,11 +2453,7 @@ ErrorVal ::= "$" VarName
   <!-- [ start Types + Tests -->
 
   <g:production name="SingleType" >
-    <g:ref name="SimpleTypeName" if=" xslt40-patterns"/>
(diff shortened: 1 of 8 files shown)


cmsmcq:
I don't see the new functions (items-before etc.) here, in the F and O spec.

ChristianGruen:
@cmsmcq

You find them via https://qt4cg.org/
The direct link is https://qt4cg.org/pr/199/xpath-functions-40/autodiff.html

ndw:
Formatted versions of the specifications with this PR applied [are available](https://qt4cg.org/#pr-199)

cmsmcq:
Looks good for the most part, but:

* In 16.2.9 fn:items-after, the last example has no call to items-after.  I would suspect the intention was to modify it so that it did have such a call, but I don't see analogues to this example in the other sections.

ndw:
This PR was accepted [on 25 October at meeting 008](https://qt4cg.org/meeting/minutes/2022/10-25.html#pr-items-before)

PR #200, opened 2022-10-10
NW for MK: Drop xsl:match instruction (185)

### LINKED ISSUE #119: Allow a map's key value to be any sequence
Since being introduced in XSLT 3.0 and later in XPath 3.1 the **_[map](https://www.w3.org/TR/xpath-functions-31/#maps-and-arrays)_** datatype has become a powerful and expressive tool for programming in XPath.

At present the value of a key of a map can be **_["an arbitrary atomic value"](https://www.w3.org/TR/xpath-functions-31/#maps-and-arrays)_**, thus a sequence of zero or more than one atomic items cannot be used in a map-key specification.

Besides giving us an almost `1 : 1` correspondence to a `JSON` object (when used with arrays, which themselves can be thought of as maps) maps are useful for expressing the tabular representation of a function that has one argument of type  `xs:anyAtomicType`.

It is not possible using a map to naturally express the tabular representation of a function having two or more (or 0) arguments. While something like this can be done using nested maps as in the example below, this technique is cumbersome and error-prone even when having two arguments, and almost prohibitively difficult when applied to expressing functions with more than 2 arguments.

Here is how we could express one possible tabular form of the function `M**N` (`M` to the power of `N`), where the two arguments are of type `xs:positiveInteger`:

```xq
let $m1 := map {1 : 1, 2 : 2, 3 : 3, 4 : 4, 5 : 5, 6 : 6, 7 : 7, 8 :  8, 9 : 9, 10 :10},
    $m2 := map {1 : 1, 2 : 4, 3 : 9, 4 : 16, 5 : 25, 6 : 36, 7 : 49, 8 : 64, 9 : 81, 10 :100},
    $m3 := map {1 : 1, 2 : 8, 3 : 27, 4 : 64, 5 : 125, 6 : 216, 7 : 343, 8 : 512, 9 : 729, 10 :1000},
    (:  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  . :)  
  
    $M := map{ 1: $m1,  2: $m2,  3: $m3 (: ... :)  }
  return $M(3)(4)
```

Here evaluating `$M(3)(4)` produces the value `4**3` (`4` to the power of `3`), that is: 
`64`

**_This proposal is to expand the allowed value-space for a key of a map to any sequence._**

Thus one will be able to write:

```xq
let $M := map{ (2, 2) : 4, (3, 2) : 9, (: . . . :)
               (2, 3) : 8, (3, 3) : 27 (: . . . :)
}
 return
    $M((2, 3))
```

And evaluating this returns the correct result `8`.

**Possible implementation**:
The implementation is very straightforward: just use in **_[op:same-key](https://www.w3.org/TR/xpath-functions-31/#func-same-key)_** as comparison function an improvement of **[`fn:deep-equal()`](https://www.w3.org/TR/xpath-functions-31/#func-deep-equal)**, which has similar behavior, but never throws errors, is context-free, error-free and transitive. Due to lack of fantasy I called this function **`deep-equal-safe()`** and its description is **[here](https://github.com/dnovatchev/FXSL-XSLT2/blob/master/fn-deep-equal-safe.pdf)**

**Note**:
No new datatypes and no changes to the XDM are necessary in order to implement this proposal.

ndw:
This PR is intended to be technically equivalent to #185.

Close #119

### DIFF
--- a/specifications/xslt-40/src/element-catalog.xml
+++ b/specifications/xslt-40/src/element-catalog.xml
@@ -737,22 +737,6 @@
          <e:parent name="try"/>
       </e:allowed-parents>
    </e:element-syntax>
-   
-   <e:element-syntax name="match">
-      <e:in-category name="instruction"/>
-      <e:attribute name="select" required="no">
-         <e:data-type name="expression"/>
-      </e:attribute>
-      <e:attribute name="pattern" required="yes">
-         <e:data-type name="pattern"/>
-      </e:attribute>
-      <e:sequence>
-         <e:element repeat="zero-or-more" name="fallback"/>
-      </e:sequence>
-      <e:allowed-parents>
-         <e:parent-category name="sequence-constructor"/>
-      </e:allowed-parents>
-   </e:element-syntax>
 
    <!-- Variables -->
 
--- a/specifications/xslt-40/src/xslt.xml
+++ b/specifications/xslt-40/src/xslt.xml
@@ -10740,31 +10740,7 @@ and <code>version="1.0"</code> otherwise.</p>
 
 
             </div3>
-            <div3 id="testing-against-pattern" diff="add" at="A">
-               <head>Testing a Value Against a Pattern</head>
-               <p>The <elcode>xsl:match</elcode> instruction can be used to test a value against a pattern.</p>
-               <?element xsl:match?>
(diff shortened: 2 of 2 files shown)


ndw:
Accepted at [meeting 007](https://qt4cg.org/meeting/minutes/2022/10-18.html).

Issue #187, opened 2022-10-07
Add a 'while' clause to FLWOR expressions


michaelhkay:
This proposal adds a new clause, the **while** clause, to FLWOR expressions. I'll start with the proposal, and then give some rationale.

**Proposal**

We add a new WhileClause which can appear anywhere a WhereClause can appear. The semantics are deliberately almost identical to the WhereClause.

3.12.x While Clause

[60]   	WhileClause	   ::=   	"while" ExprSingle	
A while clause serves as a filter for the tuples in its input tuple stream. The expression in the while clause, called the while-expression, is evaluated once for each of these tuples. If the effective boolean value of the while-expression is true, the tuple is retained in the output tuple stream; otherwise the tuple **and all subsequent tuples in the stream** are discarded.

Examples:

This example illustrates the effect of a while clause on a tuple stream:
Input tuple stream:
```
($a = 13, $b = 11)
($a = 91, $b = 42)
($a = 17, $b = 30)
($a = 85, $b = 63)
```
while clause:
`while $a > $b`
Output tuple stream:
```
($a = 13, $b = 11)
($a = 91, $b = 42)
```
The following query illustrates how a while clause might be used to extract all items in an input sequence before the first one that fails to satisfy some condition. In this case it selects the leading `para` elements in the input sequence, stopping before the first element that is not a `para` element.
```
for $x in $section/*
while $x[self::para]
return $x
```
Note:
Although the semantics are described in terms of discarding all the tuples following the first one that fails to match the condition, a practical implementation is likely to avoid evaluating those tuples, thus giving an "early exit" from the iteration performed by the FLWOR expression.

**Justification**

FLWOR expressions remain the primary control construct in XQuery, despite the introduction of higher-order functions. The inability to process input selectively until some condition is true remains one of their biggest limitations. The window clause provides a workaround for some use cases, but it is very complex, and although it can partition an input sequence based on conditions found in the data, it has no direct way of stopping processing when a condition is found. Also, it only operates on input sequences, not on the tuple stream, so it isn't able to handle a composite condition involving multiple variables, for example `for $i at $p in //emp while ($i/location = 'x' and $p < 50) return $i`.

That particular example could be written with a `where` clause. But if a `where` clause is used, both the human reader and the query optimiser need to do some thinking to work out that early exit is possible (once $p reaches 50, no further tuples are going to have values less than 50).

dnovatchev:
[...]

PR #210, opened 2022-10-13
Issue 80: fn:iterate-while (before: fn:while)

ChristianGruen:


ChristianGruen:
It appears no diffs were generated for my pull request. Maybe I missed something (or I am just too impatient)?

image

ChristianGruen:
I was my impatience that needs to be blamed. All fine now.

michaelhkay:
I think it would be useful to try to write a summary that explains to people why they might want to use this function. Something like "Processes a supplied value repeatedly, continuing while some condition remains true, and returning the first value that does not satisfy the condition."

I think this is one of those functions where it would help the reader if the rules were structured as "informally, X. More formally, the function is equivalent to..."

Perhaps the informal explanation (if I have understood it correctly) is something like this.

Informally, the function behaves as follows:

1. The supplied $input value is tested against the supplied $predicate.
2. If the result of the predicate is true, a new $input value is evaluated as the result of $action($input), and the process repeats from step 1
3. If the result of the predicate is false, the function returns the value of $input.

It would be nice to give a simple XML example, for example a function that returns the value of @xml:space on the nearest ancestor-or-self that has such an attribute.

I'm not quite sure what is meant by "a single argument" in the Errors section.

The Notes section might be a good place to put some discussion of why the function has been introduced and when it's useful: the examples give the reader some ideas, but I think more narrative would be useful.

michaelhkay:
We might also like to consider that having a function named "while" prevents us ever using "while" as a language keyword.

(Perhaps `fn:iterate()` would suggest that the function is in the same general area as `xsl:iterate`?)

michaelhkay:
And another idea for an example/use-case - does it make sense to use the function in conjuction with `random-number-generator()`?

michaelhkay:
Now implemented - 10 lines of code, so it must be a pretty clean spec. All tests passing except while-014 which depends on the `->` operator, and while-016 which fails with a numeric precision issue. Rounding the result before testing would be useful.

michaelhkay:
Readers should be made aware that it's very easy to construct an infinite loop using this function if not used correctly.

ChristianGruen:
Thanks for the suggestions from a near(est)-native speaker, very appreciated. I’ll incorporate your suggestions as good as possible.

The structure of my text was mostly inspired by the fold functions. I’m not sure: Maybe we should give them an informal section as well?

> I'm not quite sure what is meant by "a single argument" in the Errors section.

Maybe “one argument” sounds better? The error section is basically a mixture from the corresponding sections for `fn:fold-left`/`fn:fold-right` (which look at the arguments) and `fn:filter`/`array:filter` (which look at the return values). I’m uncertain if the result is really consistent, though, and I noticed that other functions with higher-order function parameters – `fn:index-where`, `fn:for-each`, etc. – have no error section at all.

Maybe we should simply skip this section, as we more or less seem to explain here how function arguments are to be evaluated in general?

> We might also like to consider that having a function named "while" prevents us ever using "while" as a language keyword.

I wonder if that wouldn’t apply to `fn:some` as well, when thinking of the `some` keyword in quantifier expressions? And I assume that we could still use `while` inside a FLWOR expression, even if we had an equally named function?

> (Perhaps `fn:iterate()` would suggest that the function is in the same general area as `xsl:iterate`?)

Personally, I would probably be confused (without reading the spec and without knowing about XSLT) why a language has `for-each` and `iterate`.

> And another idea for an example/use-case - does it make sense to use the function in conjuction with `random-number-generator()`?

I haven’t used the function a lot. Do you have an example in your mind that I can add?

> All tests passing except while-014 which depends on the `->` operator

Would you recommend removing `->` from the upcoming text cases, or was it just an observation?

> and while-016 which fails with a numeric precision issue. Rounding the result before testing would be useful.

I see. I’ll update the test case.

michaelhkay:
As regards the keyword, It could certainly lead to parsing difficulties, e.g. if we had a function called "where" then we couldn't allow "where" as the first clause in a FLWOR expression (something that would make sense to allow).

In practice it's probably not the technical grammar ambiguity that's the issue, as the user confusion that could result if we use the same word to mean different things in different contexts.

michaelhkay:
I've been fairly guilty myself in using syntax that isn't yet accepted in test cases, so I can't really complain. But it's probably best avoided if unnecessary - there's a general principle when writing tests to focus on the feature you're testing.

michaelhkay:
Trying to use random-number-generator to generate a random sequence (or anything else to generate a sequence), I find myself wanting to output values when the predicate is true, rather than accumulating them until it is false. `xsl:iterate` has the advantage that you can add values to the output at any stage of the iteration, and also that you can pass any number of values (parameters) from one iteration to the next. The xsl:iterate/on-completion clause (like a FLWOR return) also enables the final result to be computed from the running state. With fn:while, this seems to require maintaining a map to maintain state - and that also means the initial value and final result also have to use the same map structure, which exposes the internal state on completion. Perhaps I shouldn't have been thinking of this function as a sequence generator.

ChristianGruen:
> As regards the keyword, It could certainly lead to parsing difficulties, e.g. if we had a function called "where" then we couldn't allow "where" as the first clause in a FLWOR expression (something that would make sense to allow).

Wouldn’t we have to rename `fn:some` as well, then? [`ExprSingle`](https://qt4cg.org/specifications/xquery-40/xquery-40-diff.html#prod-xquery40-ExprSingle) can be both a `FLWORExpr` and `QuantifiedExpr`.

> Trying to use random-number-generator to generate a random sequence […]

I’ve concocted a little example which I’ll add in my updated version. I won’t be offended if you plan to edit my pull request (I don’t know if the notes will be explanatory enough).

ndw:
Formatted versions of the specifications with this PR applied [are available](https://qt4cg.org/#pr-210)

ChristianGruen:
@cmsmcq Thanks, very appreciated. I have incorporated your suggestions.

Arithmeticus:
The function seems quite useful. In fact, as I think of using it, my first impulse would be to use it as a keyword, on analogy to `if` or `for`. Could there be discussion, `while` this proposal is open, on exposing the functionality via keyword? (One possibility to hack into `while`: introduce the expression `break` to exit a `for`)

ChristianGruen:
@Arithmeticus Thanks.

I think that both a function and keyword could be beneficial. Michael Kay has added a proposal for a `while` clause in FLWOR expressions in #187.

cmsmcq:
Looks good, I think.  For the second rule I wonder if a different formulation would be clearer. For:

> If the result of the predicate is true, a new $input value is evaluated as the result of $action($input), and the process repeats from step 1.

perhaps read:

> If the result of the predicate is true, the $action function item is called with the $input as its argument (i.e. the expression $action($input) is evaluated), the resulting value is used as a new $input, and the process repeats from step 1.

(This is longer than I hoped when I started, and if we can shorten it that would be good.)  The key change for me is to say more clearly that the input value is produced by evaluating the action.

ChristianGruen:
@cmsmcq Thanks. I’ve changed it to:

> If the result of the predicate is <code>true</code>, <code>$action($input)</code> is evaluated, the resulting value is used as a new <code>$input</code>, and the process repeats from step 1.

ndw:
Approved at [meeting 011, 15 November 2022](https://qt4cg.org/meeting/minutes/2022/11-15.html)

michaelhkay:
Superseded by #962

PR #215, opened 2022-10-17
First attempt at parse-uri and build-uri functions

### LINKED ISSUE #72: [FO] Provide better support for URI processing within an expression
> Use Case 1: Decode an encoded URI string.

This is difficult to implement correctly, and is a commonly asked question/request on sites like stackoverflow. Vendors have even implemented their own functions, like `xdmp:uri-decode`.

> Use Case 2: Extracting the hash/parameters from a URI string.

This is common when manipulating URI strings and not using something like RESTXQ to bind the query parameters to function parameters. The API should:
1. extract the hash as a string, and the parameters as a name/value map;
2. combine parameters with the same name into the same map entry;
3. decode the values where necessary.

> Use Case 3: Extract the other parts of a URI string.

This can be useful if writing a RESTXQ or similar implementation in XSLT/XQuery. It can also be useful for generating response headers such as Origin, or doing HTTP to HTTPS redirects.

It is easy to make mistakes and the wrong assumptions when writing a URI parser by hand. Additionally, it is not easy to implement in XSLT/XQuery as functions like `analyse-string` and `tokenize` are not powerful enough to implement a lexer, and manipulating codepoints is difficult without stateful logic.

ndw:
Close #72 

This is a first attempt to resolve issue 72 by describing `parse-uri` and `build-uri` functions.

I decided to try an approach that leverages the ABNF defined in RFC 3986. Attempting to redefine the rules just opens up the possibility that the rules we write will differ from the rules in RFC 3986 which would be wrong.

I added an options map and specified a "details" option that will further parse the path and query components. This gives the caller easy access to the percent-decoded forms of those elements without making the parse function irreversable.

Notes:

1. More examples are needed
2. I tried to use `eg` in the results so that they'd be formatted nicely, but it didn't work. Is there any markup for verbatim text in examples?
3. There's an open question about whether non-hierarchical URIs can have fragment identifiers or queries. I think not, but I need to review the RFC a little more carefully.

### DIFF
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -1259,8 +1259,6 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
       <g:string>$</g:string>
       <g:ref name="VarName"/>
     </g:optional>
-    <g:string>$</g:string>
-    <g:ref name="VarName"/>
     <g:string>return</g:string>
     <g:ref name="ExprSingle"/>
   </g:production>
@@ -1272,9 +1270,6 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
       <g:ref name="VarName"/>
       <g:string>as</g:string>
     </g:optional>
-    <g:string>$</g:string>
-    <g:ref name="VarName"/>
-    <g:string>as</g:string>
     <g:ref name="SequenceTypeUnion"/>
     <g:string>return</g:string>
     <g:ref name="ExprSingle"/>
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -23701,10 +23701,16 @@ declare function fn:some(
          is present and it’s value is not a string of length one.</p>
 
          <p>A <emph>query-segments</emph> is constructed as follows: tokenize
-         the <emph>query</emph> on the <emph>query separator</emph>,
-         apply <emph>uri decoding</emph> on each token, and convert
-         the result to an array.</p>
-         
+         the <emph>query</emph> on the <emph>query separator</emph>. For each
+         token, construct a map. If the token contains an equal sign (“=”),
+         the map contains a key named <code>key</code> with a value equal to the
+         string preceding the first equal sign and a key named <code>value</code>
+         with a value equal to the string following the first equal sign. If the
+         token does not contain an equal sign, the map contains a single key named
+         <code>value</code> with a value equal to the token. In every case,
+         <emph>uri decoding</emph> is applied to each value add to the map.
+         The resulting sequence of maps is converted into an array.</p>
+
          <p>The following map is returned:</p>
 
          <eg>{
@@ -23791,7 +23797,10 @@ map {
   "port": "080",
   "path": "/path",
   "query": "s=%22hello world%22&amp;sort=relevance",
-  "query-segments": array { "s=""hello world""", "sort=relevance" },
+  "query-segments": array {
+    map { "key": "s", "value": "&quot;&quot;hello world&quot;&quot;" },
+    map { "key": "sort", "value": "relevance" }
+  },
   "path-segments": array { "", "path" }
 }</eg></fos:result>
     </fos:test>
@@ -23929,7 +23938,9 @@ map {
   "uri": "?q=1",
   "path": "",
   "query": "q=1",
-  "query-segments": array { "q=1" }
+  "query-segments": array {
+    map { "key": "q", "value": "1" }
+  }
 }</eg></fos:result>
     </fos:test>
   </fos:example>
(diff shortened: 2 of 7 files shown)


ndw:
Formatted versions of the specifications with this PR applied [are available](https://qt4cg.org/#pr-215)

ndw:
I started with a vision of parsing RFC 3986 (absolute) URIs. I think Mike makes a very good point that one reason to parse a URI is to find out things like, is it absolute or relative? The trouble is, once you start accepting those strings, you quickly find yourself up against special cases.

Experience has shown, for example, that `file:` URIs on Windows are especially problematic. One often encounters  `file:\\c:\path` which isn't even really a URI, though it's "obvious" what was intended. Constraining ourselves to actual URIs, we still get awkward things like `file://c:/path` and `file:////unc/path`. (There is no specification for the `file:` URI scheme, so we can't appeal to anyone for help.)

How far down the road to [p:urify()](https://spec.xproc.org/3.0/xproc/#f.urify) do we want to go?

I'm tempted to add things like:

> If a string begins with a single letter followed by a colon, assume it's a file: URI on Windows.

and 

> Convert all backslashes in the string into forward slashes, before analyzing the string

I think we can get something useful and interoperable this way, but it'll be a little ad hoc of necessity.

michaelhkay:
Just been reading p:urify(). At least it reads like a spec. Everything else I've seen reads like arm-waving.

ndw:
1. I'd be happy to have `fn:urify()`, it doesn't do exactly what parse and build do.
2. The implementation of `fn:urify()` is specifically different on Windows versus other environments. I don't know if we want to do that or not.

gimsieke:
The `p:un` would be lost on `fn:urify` though.

ndw:
I’ve done a pretty substantial reworking of *parse-uri()* based on my attempts to address the feedback here.

The apparently spurious commits caused by rebasing on master were necessary in order to get some stylesheet updates into the build that will make the formatted version much easier to read!

Mike wrote:

> the function is described as context-independent, but then it says it expands a relative URI using the static base URI.
> 
> I'm not sure expanding a relative URI is a good thing. One thing you want to know about the URI is whether it's relative or absolute.

Right. Okay. Let’s see what happens if we just make it operate on the string passed in. The new prose is a bit procedural, but that’s (explicitly) about exposition and not implementation.

> There are no error conditions defined, which seems a little surprising.

I decided that it shouldn’t raise errors. This is especially true if it’s just going to attempt to parse random strings.

> The ABNF that I'm looking at uses different names for the URI parts: iuserinfo rather then userinfo, etc. So we need to define the mapping.

`iuserinfo`? What spec are you looking at? I’m reading [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986).

> It's not clear to me whether absent components should appear in the result map using a value of "", or (), or should be omitted.

I’ve currently opted for making them appear in the map as `()`. I could be persuaded that absent values should be left out of the map. In fact, in the examples, I’ve left them out simplicity, which is maybe kind of a clue.

> I would be inclined to do the details=true() option unconditionally.

Yes.

> Should allow ";" as an alternative to "&" between query parts? I know it's not widely used but it's there and is often more convenient than "&" in an XML context.

It’s not standard. But I added it as an option. I added a path separator option as well.

> The expansion of the query part as a sequence of maps preserves all the information, but isn't as easy to use as, for example, returning a single map in which multiple entries for the same keyword are combined into a sequence (So ?x=3&x=4 becomes map{'x', (3,4)} rather than (map{'x':3}, map{'x':4}).

I’ve replaced that with an array. You have to parse on the equal sign yourself.

> "Escaped" should probably be "percent-encoded".

That text is gone, I think.

Arithmeticus:
Thanks for all this work, Norm. The variety of examples are very useful. Could you also include a tag URI example?
https://tools.ietf.org/html/rfc4151

e.g., 
tag:textalign.net,2015:ns
tag:jan@example.com,1999-01-31:my-uri

ndw:
Sure, @Arithmeticus. FYI:
```
map {
    "uri": "tag:textalign.net,2015:ns",
    "scheme": "tag",
    "path": "textalign.net,2015:ns",
    "path-segments": [ "textalign.net,2015:ns" ]
  }
```
and
```
map {
    "uri": "tag:jan@example.com,1999-01-31:my-uri"
    "scheme": "tag",
    "path": "jan@example.com,1999-01-31:my-uri",
    "path-segments": [ "jan@example.com,1999-01-31:my-uri" ],
}
```

ndw:
Updated per comments from [meeting 009](https://qt4cg.org/meeting/minutes/2022/11-01.html#pr-parse-uri):

1. Added prose describing how parse-uri and resolve-uri are related.
2. Reviewed RFC 3986 wrt relative URIs and reworded the introduction.
3. Made the return-type a record, mad the argument type in `build-uri` a record, and updated the stylesheets to support a more readable rendering if record types as arguments and return types
4. Added an error test for parse-uri; added a non-error test demonstrating the use of `query-separator`
5. Made the `query-segments` an array of maps
6. Redrafted the prose for `build-uri`.
7.

ndw:
Accepted at [meeting 011, 15 November 2022](https://qt4cg.org/meeting/minutes/2022/11-15.html)

PR #223, opened 2022-10-27
Allow eg in fos:expression and fos:result

ndw:
This PR formats `eg` in `fos:expression` and `fos:result` as a block. I don't think this will have any detrimental effects as it's done during the "merging" process.


michaelhkay:
Be aware of the stylesheet generate-qt3-test-set.xsl which generates a test set containing all the executable examples. We haven't regenerated this test set for a while, and we probably should. If I remember right, it may need a little manual tweaking. It generates test app/fo-spec-examples.xml.

ndw:
Right. This appeared to be in the stylesheet that's merging results back and generating HTML. But if I got it wrong, I'll figure it out when we regenerate that, I guess.

PR #226, opened 2022-10-31
Add the Data Model specification; rename Serialization

ndw:
**ATTENTION**

There are a *whole* bunch of markup errors in the F&O specification. The build system has not been validating it before publication 😞 😢 😭 . I have fixed them in this PR. I fixed the markup where it was clear how to do that. I hacked at the DTDs to allow some new markup patterns. It's kind of a mess and if you've got an open PR on the F&O spec, I'm afraid it's going to be painful.

Aside from that horror show (hey, it's Halloween, maybe it's appropriate), I:

1. Added the Data Model specification
2. Renamed the Serialization specification from 3.1 to 4.0
3. Reworked the way the cross reference index files (the `/etc/` directory) are handled. The 4.0 versions are now generated every time, and should not be committed to the repository.

I expect that the old ant build scripts will no longer work. I'm probably going to pull all that infrastructure out Real Soon Now™ as we've been working successfully with the new system.


ndw:
This fails because the PR runs with the "old" infrastructure and I've made changes that depend on the "new" infrastructure. I'm going to see about refactoring this without the F&O changes.

Issue #74, opened 2021-05-14
[FO] Support parsing HTML


rhdunn:
It is common for applications that use an XQuery database engine to want to parse HTML documents when adding content from HTML pages into a database, or in other applications like generating epub documents from HTML source files. Vendors like MarkLogic (`xdmp:tidy` via HTML Tidy for HTML4), BaseX (`html:parse` via TagSoup), Saxon (`saxon:parse-html` via TagSoup), and eXist-db (`util:parse-html` via Neko) have provided custom methods to support this.

Q: Should there also be functions to list the supported methods and character encodings?

# fn:parse-html
## Summary
Parses HTML-based input into an XML document.
## Signature
```
fn:parse-html($input as union(xs:string, xs:hexBinary, xs:base64Binary),
              $options as map(*) := map { "method": "html5" }) as document-node()
```
## Properties
This function is ·deterministic·, ·context-independent·, and ·focus-independent·.
## Rules
The $options map conforms to `record(method as union(enum("html5"), xs:string), encoding as xs:string?, *)`. A vendor may provide ·implementation-dependent· options that may vary between the different `method` values.

The `method` property of $options defines the approach used to convert the HTML document to XML. This specification supports `html5` for using the HTML5 parsing rules for HTML content. The exact version of HTML5 used is ·implementation-dependent·.

The `encoding` property of $options defines the character encoding used to decode binary data. By default, this is an empty sequence. Implementations must support at least `utf-8`, `utf8`, `ascii`, and `latin1`. Other encoding values are ·implementation-dependent·, but it is recommended that the encodings documented in the WHATWG Encoding specification [3] are supported.

If $input is an `xs:string`, no character decoding is performed as the input is already decoded.

If $input is an `xs:hexBinary` or `xs:base64Binary`, the character encoding used to decode the binary data is determined as follows:
1. if the binary data has a valid Unicode Byte Order Mark (BOM), the character encoding specified by that BOM is used.
1. if `encoding` is specified in $options, that value is used;
1. if prescanning the first 1024 bytes of data contains a character encoding (using the rules from https://html.spec.whatwg.org/multipage/parsing.html#prescan-a-byte-stream-to-determine-its-encoding), the detected encoding is used;
1. if ·implementation-dependent· heuristics (in line with the HTML5 rules) detect a character encoding, that encoding is used;
1. otherwise, the encoding is "utf-8".

[...]

PR #259, opened 2022-11-18
Issue #74 - add the fn:parse-html function

rhdunn:
This PR makes the following changes:
1. Add an `.editorconfig` file to allow editors and IDEs to indent the XML source documents consistently.
1. Group the XML and JSON parsing and serialization functions into a common top-level section.
1. Update the html5 bibref to the WHATWG specification -- the old HTML5 link redirects there.
1. Add a new fn:parse-html function specification.
1. Define a mapping between the HTML DOM and the XDM nodes.

The function specification itself is complete. The HTML DOM to XDM node mapping currently only contains the overview. I'm going to write that for this PR, this is just to start the ball rolling in specifying this function.


liamquin:
Possibly there should be an html-version option as per Serialization spec and XSLT. And possibly allow automatic detection of HTML 5 vs quirks mode vs XHTML with validation?? I like simple, but i am not confident hard-wiring HTML 5 (as opposed to XHTML 5 or XHTML 1.1) is necessarily the way to go. Of course, the HTML 5 spec says what to do with these too, but the proposed prose here seems explicit that only HTML 5 is accepted. Nonetheless i'm very pleased to see this.

rhdunn:
@liamquin I've incorporated your suggestions into the spec. Thanks!

michaelhkay:
Could I ask what are the implications of

> Add an .editorconfig file to allow editors and IDEs to indent the XML source documents consistently.

We've always had a problem with multiple (human) editors using multiple (software) editors causing whitespace changes that disrupt change marking and conflict resolution. I don't know what an .editorconfig file does: is it going to help with this problem or make it worse?

rhdunn:
> Could I ask what are the implications of
> 
> > Add an .editorconfig file to allow editors and IDEs to indent the XML source documents consistently.
> 
> We've always had a problem with multiple (human) editors using multiple (software) editors causing whitespace changes that disrupt change marking and conflict resolution. I don't know what an .editorconfig file does: is it going to help with this problem or make it worse?

An editor config file allows you to specify the encoding, line ending, indentation style, etc. for files in the project. This allows you to have different rules for different language extensions, or even different files. The settings are documented in https://editorconfig.org/.

I added it because my IDE (IntelliJ IDEA) was defaulting to 4 characters indentation of the elements instead of the 3 that the files use. As such, I've added a rule in the editor config to make that consistent. I've not extended it to other settings such as encoding, line endings, or removing trailing spaces.

rhdunn:
The XDM node mapping for the HTML DOM is now complete. I've incorporated @michaelhkay's notes from the issue, except for the point about if case normalization is needed.

ndw:
A higher-level comment is that, while I don't object to the options provided, I'd have been equally happy if the function said that the input string is parsed with the HTML5 parsing algorithm, full stop.

ChristianGruen:
> I think I'd have a marginal preference for putting method-specific additional parameters in their own key:

Sounds reasonable to me, not just because TagSoup has a `method` option, too: https://www.mankier.com/1/tagsoup#--method

michaelhkay:
Drop 2.6.14 fn:parts. Superseded by parse-uri.

Drop 2.6.15 EQNames - done.

Drop 2.6.16 parse-html - duplicate of 2.3

Drop 2.6.18 fn:until - done (as fn:iterate-while)

Drop 2.6.19  - (fn:CR etc) duplicate of 2.4

Mike

ndw:
I think the comment about dropping various 2.6.x things is related to the agendas. I've made those changes for the next agenda.

michaelhkay:
I think we need greater clarity as regards error handling. The proposed spec says:

> A dynamic error is raised [[err:FODC0011] if the content of $html is not a well-formed HTML document.

But I'm not sure the concept of a "well-formed HTML document" is well defined.

Specifically, there seem to be differences between implementations of the HTML5 parsing algorithm (both in browsers, and in free-standing libraries) as to how much recovery is attempted. For example the proposed test case test-003 include the ill-formed SVG fragment:

```
<svg style="overflow: clip; width: 100px; height: 100px">
  <rect width="200" height="200" fill="green">
</svg>
```

(note the missing end tag for `rect`). Browsers and parsing libraries appear to differ on how they handle this (validator.nu reports an error, JSoup takes silent recovery action).

We could take several paths:

- be prescriptive one way or the other
- make it a user option
- make it implementation-defined

I think this is probably a case for pragmatism; implementations are going to want to use existing parsing libraries, and if these differ in their behaviour, it's probably best to leave things implementation-defined.

There are no easy answers here; the HTML5 spec has a lot of fuzziness. For example, most of the requirements it places on "user agents" are not applicable because they only affect rendering and interaction, but the construction of a DOM from lexical HTML is also described as a function of a "user agent".

PR #268, opened 2022-11-24
New type-hierarchy images / descriptions

### LINKED ISSUE #265: Type hierarchy tables/diagrams
On 22 November, we mentioned in passing that the type hierarchy tables in the Data Model spec are hard to read. They're a garish yellow for no obvious reason and the font size is very small. Would replacing them with diagrams like this be an improvement?

out

(This is the largest and most difficult to represent, I think. I've grouped some of the atomic types together to make the drawing more aesthetically pleasing. I don't think that interferes with comprehension, though might need to be explained.)

There's no particular rhyme or reason to the order of the items in the "second column" except to leave room for the longer hierarchies further to the right.

ndw:
Close #265 

This is my attempt to improve the type-hierarchy images.

1. There *were* graphics already, but they were accidentally not being copied into the right place.
2. But, they're for 3.1 and they're in some format I don't recognize, so
3. I created new SVG ones anyway, that folks may or may not like
4. I replaced the big, ugly yellow tables with prose. I concluded that accessibility was the only reason they were present.
5. The colors are a little different, and a little "off", but they're explicitly chosen from a pallet that offers unambiguously differenty colors for the three most common forms of color-blindness.
6. One of the SVG images is too wide, I'm not sure what to do about that
7. The labels are text, so you can search for them, and xs:anyAtomicType is a link. We could make more links.


ndw:
Agreed with actions at the meeting on 13 December 2022.

PR #287, opened 2022-12-14
PR #279 with merge conflicts resolved

### LINKED ISSUE #277: Overriding functions using xsl:import
In XSLT 3.0, you can have a module M that contains an xsl:function F#2, and in another module you can import M, and declare another xsl:function F#3. 

In the 4.0 spec we appear to disallow this with the paragraph (in §10.3.4):

> A stylesheet function may be overridden by another stylesheet function with the same name that has higher [import precedence]. This is only allowed, however, if the [arity range] of the overriding function includes the totality of the arity range of the overridden function.

In fact, the error conditions we go on to define (XTSE0769 and XTSE0770) are more carefully worded and do not make the above situation an error. XTSE0769 says that if F has higher import precedence than G, then either the arity ranges of F and G must be disjoint, or the arity range of F must include the totality of G. XTSE0770 says that if F and G have the same import precedence, then their arity ranges must be disjoint.

The paragraph cited should be replaced with:

> A stylesheet function may be overridden by another stylesheet function with the same name that has higher [import precedence]. This is only allowed, however, if either (a) the [arity range] of the overriding function includes the totality of the arity range of the overridden function, or (b) the two arity ranges are non-overlapping.

There is also scope for editorial improvement to §10.3.4. It describes three separate scenarios:

(a) overriding functions using xsl:import and import precedence

(b) overriding functions using xsl:use-package and xsl:override

(c) overriding extension/external functions using xsl:function (or vice versa)

and it would be much easier to read the section if these were clearly distinguished.

ndw:
Close #279
Close #277

### DIFF
--- a/specifications/xslt-40/src/xslt.xml
+++ b/specifications/xslt-40/src/xslt.xml
@@ -18359,54 +18359,116 @@ and <code>version="1.0"</code> otherwise.</p>
 
 
             </div3>
-            <div3 id="function-visibility-and-overriding" diff="chg" at="variadicity">
+            <div3 id="function-visibility-and-overriding" diff="chg" at="2022-11-29">
                <head>Visibility and Overriding of Functions</head>
+               
+               <p>It is possible to have multiple function declarations sharing the same function name and arity:</p>
+               <ulist>
+                  <item><p>Multiple <elcode>xsl:function</elcode> declarations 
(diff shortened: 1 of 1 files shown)


ndw:
We agreed to merge this at the 13 Dec 2022 meeting.

PR #303, opened 2023-01-09
Mike's proposed schema and stylesheet changes

ndw:
These are the schema and stylesheet changes from PR #292. They don't break the build and on casual inspection they seem fine, so I'm just going to accept them.


ndw:
The checks don't really tell us much because stylesheet changes don't apply to PRs, but hey, I let the run anyway.

PR #304, opened 2023-01-09
Mike's content changes from PR 292

### LINKED ISSUE #35: [FO]The `union ( | )`, `itersect`, `except`  and `combine (,)` operators are not mentioned in the F & O. Have not the best categorization in the XPath spec.
1. Searching the F & O document for the `union ` ( `|` ), `intersect ` and `except` operators finds 0 occurrences. Neither can `,` be found, and of course, it is unsearchable.

2. In the XPath spec these are described in section "4.7.3 Combining Node Sequences". The word node-set is not mentioned at all and this is misleading, because the `union ` ( `|` ), `intersect ` and `except` operators are in fact set-operators, or at least set-generating operators.

Thanks,
Dimitre

### LINKED ISSUE #70: [FO] Built-in function changes to support default values
This issue tracks the changes needed to the built-in functions to allow them to combine the declarations into a single definition with default parameter values.

The general approach to this is to make required arguments optional if they are for a function signature that is not the lowest argument count signature, and move any associated logic into the function.

## array:subarray
1. Change `array:subarray`/`$length` from `xs:integer` to `xs:integer?`.

__Rules__

Except in error cases, the result of the function is the value of the expression op:A2S($array) => fn:subsequence($start, $length) => op:S2A().

__Error Conditions__

A dynamic error is raised [err:FOAY0001] if $start is less than one or greater than array:size($array) + 1.

A dynamic error is raised [err:FOAY0002] if $length is not an empty sequence and is less than zero.

A dynamic error is raised [err:FOAY0001] if $length is not an empty sequence and $start + $length is greater than array:size($array) + 1.

## fn:concat
This should be modified to use a sequence-variadic signature, either as a 1 parameter function (taking an `xs:anyAtomicType*` value, allowing 0 and 1 arguments), or a 3 parameter function with the last parameter having the type `xs:anyAtomicType*`.

## fn:differences
The `$options` parameter should be moved to the end of the parameter list in order to make the function a map-variadic function when default values are applied. This then makes it possible to specify the collation argument using a keyword argument in addition to specifying options as keyword arguments.

## fn:resolve-uri
1. Change `fn:resolve-uri`/`$base` from `node()` to `node()?`.

```
- If the $base argument is not supplied,
+ If the $base argument is the empty sequence,
```

## fn:subsequence
1. Change `fn:subsequence`/`$length` from `xs:double` to `xs:double?`.

When $length is the empty sequence, this function returns:

    $input[fn:round($start) le position()]

When $length is not the empty sequence, this function returns:

    $input[fn:round($start) le position() 
             and position() lt fn:round($start) + fn:round($length)]

## fn:substring
1. Change `fn:subsequence`/`$length` from `xs:double` to `xs:double?`.

More specifically, when $length is not the empty sequence the function returns the characters in $value whose position $p satisfies:

    fn:round($start) <= $p and $p < fn:round($start) + fn:round($length)

When $length is the empty sequence the function assumes that $length is infinite and thus returns the ·characters· in $value whose position $p satisfies:

    fn:round($start) <= $p

## fn:tokenize
1. Change `fn:tokenize`/`$pattern` from `xs:string` to `xs:string?`.

If $pattern is the empty sequence, the $value argument is set to `fn:normalize-space($value)` and $pattern is set to `' '`.

## fn:unparsed-text / fn:unparsed-text-available
1. Change `fn:unparsed-text`/`$encoding` from `xs:string` to `xs:string?`.
1. Change `fn:unparsed-text-available`/`$encoding` from `xs:string` to `xs:string?`.

## fn:unparsed-text-lines
1. Change `fn:unparsed-text-lines`/`$encoding` from `xs:string` to `xs:string?`.

The result of the function is the same as the result of the expression `fn:tokenize(fn:unparsed-text($href, $encoding), '\r\n|\r|\n')[not(position()=last() and .='')]`.

## Collations
1. Change `fn:collation-key`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:compare`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:contains`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:contains-token`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:deep-equal`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:differences`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:distinct-values`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:ends-with`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:index-of`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:max`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:min`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:starts-with`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:substring-after`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:substring-before`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:uniform`/`$collation` from `xs:string` to `xs:string?`.
1. Change `fn:unique`/`$collation` from `xs:string` to `xs:string?`.

Passing the empty sequence to the $collation argument is equivalent to supplying the default collation to that argument.

### LINKED ISSUE #257: Improving the styling/presentation/prepresentation of the record types in the F&O spec
I'm not keen on the † symbol being used to indicate and link record types. This is not used elsewhere, and when a record type is used as a type, there is a link to the record type definition. I think that that link is sufficient.

The presentation of the record type definition has the following format:

    †type-name:
    record(
        ...
    )

The record part is fine, but I find the `†type-name:` part clunky. For functions, omiting the `declare function` part from the XQuery syntax makes sense as the function signature is readable without that.

The XQuery 4.0 draft spec has the following (as of yet unapproved) syntax for defining type aliases: https://qt4cg.org/specifications/xquery-40/xquery-40.html#id-item-type-declaration. I would suggest using something similar, e.g.:

    type-name as record(
        ...
    )

That is in line with the way function declarations are specified and looks more readable to me.

Finally, ommitting the id from the `record` element results in `†:` before the record. -- It would be nice if instead of:

            <div3 id="html-parser-options">
               <head>HTML parser options</head>
               <example role="record">
                  <record id="parse-html-options">
                     <arg name="method" type="union(enum(&quot;html5&quot;), xs:string)"/>
                     <arg name="*"/>
                  </record>
               </example>

you could write something like:

            <div3 id="parse-html-options">
               <head>HTML parser options</head>
               <example role="record">
                  <record type-name="parse-html-options">
                     <arg name="method" type="union(enum(&quot;html5&quot;), xs:string)"/>
                     <arg name="*"/>
                  </record>
               </example>

### LINKED ISSUE #288: Error in fn:path specification
On the XML.com slack, Phil Fearon observes:

> The XPath 3.1 specification has an error in the [definition of fn:path](https://www.w3.org/TR/xpath-functions-31/#func-path) 
> 
> The properties section states:
> 
> The one-argument form of this function is [·deterministic·](https://www.w3.org/TR/xpath-functions-31/#dt-deterministic), [·context-dependent·](https://www.w3.org/TR/xpath-functions-31/#dt-context-dependent), and [·focus-dependent·] (https://www.w3.org/TR/xpath-functions-31/#dt-focus-dependent).
> The two-argument form of this function is [·deterministic·](https://www.w3.org/TR/xpath-functions-31/#dt-deterministic), [·context-independent·](https://www.w3.org/TR/xpath-functions-31/#dt-context-independent), and [·focus-independent·](https://www.w3.org/TR/xpath-functions-31/#dt-focus-independent).
> 
> The term one-argument form should be zero-argument form and consequently, two-argument form should be one-argument form

### LINKED ISSUE #291: DTD validity of F&O spec
Despite PR #228, I'm still seeing a lot of validation errors when (using Oxygen) I apply DTD validation to build/expanded/xpath-functions-40/xpath-functions-40.xml

Many of these relate to style attributes not being allowed on table, td, and code elements.

The merge-function-specs.xsl explicitly creates elements with these attributes:

See 

* `<xsl:template match="fos:options">` line 382
* `<xsl:template match="fos:option">` line 408
* `<xsl:template match="processing-instruction('local-function-index')">` line 347

It all works because xmlspec-2016.xsl handles these attributes even though the DTD doesn't allow them: see `<xsl:template name="style-attributes">` at line 2096.

There seem to be three possible options:

* Extend the DTD to allow these attributes
* Change merge-function-specs.xsl to not generate these attributes, instead perhaps generating a role attribute which gets picked up in the final styling
* Sweep the issue under the carpet; avoid validating the intermediate file.

The main problem with the third approach is that it allows other errors to go unnoticed, for example putting text directly within an `<item>` without a containing `<p>`.

ndw:
I teased apart some of the omnibus PR #292. I've commited the schema and stylesheet changes. This PR covers the remaining prose changes.

Mike writes:

> I regret that this has turned into a bit of an omnibus PR. The main changes are:
>
> * Fix validity issues with the function catalog and its schema (Issue 291)
> * Convert all functions to use a single signature with optional parameters (Issue 70)
> * Extend the function catalog to handle record definitions (Issue 257)
> * Fix the (trivial) bug with properties of fn:path (Issue 288)
> * Add introductory text concerning the handling of operators (Issue 35)

Fix #291 
Fix #70
Fix #257 
Fix #288 
Fix #35

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
 <fos:functions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://www.w3.org/xpath-functions/spec/namespace fos.xsd"
    xmlns:fos="http://www.w3.org/xpath-functions/spec/namespace">
@@ -17,9 +16,33 @@
       <fos:variable id="v-item2" name="item2" select="$po/line-item[2]"/>
       <fos:variable id="v-item3" name="item3" select="$po/line-item[3]"/>
    </fos:global-variables>
+
+   <fos:type id="uri-structure-record">
+      <fos:record extensible="true">
+         <fos:field name="uri" type="xs:string" required="false"/>
+         <fos:field name="scheme" type="xs:string" required="false"/>
+         <fos:field name="authority" type="xs:string" required="false"/>
+         <fos:field name="userinfo" type="xs:string" required="false"/>
+         <fos:field name="host" type="xs:string" required="false"/>
+         <fos:field name="port" type="xs:string" required="false"/>
+         <fos:field name="path" type="xs:string" required="false"/>
+         <fos:field name="query" type="xs:string" required="false"/>
+         <fos:field name="fragment" type="xs:string" required="false"/>
+         <fos:field name="path-segments" type="array(xs:string)" required="false"/>
+         <fos:field name="query-segments" type="array(record(key? as xs:string, value? as xs:string, *))" required="false"/>
+      </fos:record>
+   </fos:type>
+   
+   <fos:type id="random-number-generator-record">
+      <fos:record extensible="true">
+         <fos:field name="number" type="xs:double" required="true"/>
+         <fos:field name="next" type="function() as #random-number-generator-record" required="true"/>
+         <fos:field name="permute" type="function(item()*) as item()*" required="true"/>
+      </fos:record>
+   </fos:type>
+   
    <fos:function name="node-name" prefix="fn">
       <fos:signatures>
-         <fos:proto at="F" name="node-name" return-type="xs:QName?"/>
          <fos:proto name="node-name" return-type="xs:QName?">
             <fos:arg name="node" type="node()?" default="." usage="inspection"/>
          </fos:proto>
@@ -38,9 +61,9 @@
          <p>Returns the name of a node, as an <code>xs:QName</code>.</p>
       </fos:summary>
       <fos:rules>
-         <p>If the argument is omitted, it defaults to the context item (<code>.</code>). The
+         <p>If the argument is omitted, it defaults to the context item (<code>.</code>).<phrase diff="del" at="2022-11-29"> The
             behavior of the function if the argument is omitted is exactly the same as if the
-            context item had been passed as the argument.</p>
+            context item had been passed as the argument.</phrase></p>
          <p>If <code>$node</code> is the empty sequence, the empty sequence is returned.</p>
          <p>Otherwise, the function returns the result of the <code>dm:node-name</code> accessor as
             defined in <bibref
@@ -79,7 +102,6 @@
    </fos:function>
    <fos:function name="nilled" prefix="fn">
       <fos:signatures>
-         <fos:proto name="nilled" return-type="xs:boolean?"/>
          <fos:proto name="nilled" return-type="xs:boolean?">
             <fos:arg name="node" type="node()?" default="." usage="inspection"/>
          </fos:proto>
@@ -99,9 +121,9 @@
          <p>Returns true for an element that is <term>nilled</term>.</p>
       </fos:summary>
       <fos:rules>
-         <p>If the argument is omitted, it defaults to the context item (<code>.</code>). The
+         <p>If the argument is omitted, it defaults to the context item (<code>.</code>). <phrase diff="del" at="2022-11-29">The
             behavior of the function if the argument is omitted is exactly the same as if the
-            context item had been passed as the argument.</p>
+            context item had been passed as the argument.</phrase></p>
          <p>If <code>$node</code> is the empty sequence, the function returns the empty sequence.</p>
          <p>Otherwise the function returns the result of the <code>dm:nilled</code> accessor as
             defined in <bibref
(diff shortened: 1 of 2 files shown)


michaelhkay:
I noticed a couple of problems.

1. fn:trace() - no default is defined for $label, so we have lost the arity-1 version of the function
2. fn:tokenize() - the semantics of tokenize(x) is not the same as tokenize(x, ' '), because the former case effectively does a `normalize-space()` call before tokenizing.
3. In uri-structure-record, the fields in the record seem to have become mandatory.
4. The prose for fn:sum() refers to "the first signature" and "the second signature". Ditto fn:compare and fn:round-to-half-even.
5. On fn:all-different, the type of a parameter is given as "xs:anyAtomicType**"

ndw:
Accepted at [meeting 018](https://qt4cg.org/meeting/minutes/2023/01-17.html)

PR #347, opened 2023-02-15
Attempt to clarify fn:parse-uri and fn:build-uri

### LINKED ISSUE #307: Parsing and building URIs comments and queries
1. fn:build-uri states:

> If the scheme key is present in the map, the URI begins with the value of that key concatenated with //, otherwise it begins //.

a. Shouldn't the concatenation be `://` so e.g. `http` becomes `http://`?
b. How are non-heirarchical schemes handled like `urn`, and `mailto`?

2. RFC 3986 allows IPv6 and IPvFuture addresses that contain `:` characters, e.g. `http://[::1]:80`.

My understanding of fn:parse-uri is that this will fail to parse.

3. RFC 3986 states that for userinfo, the `user:password` form is deprecated.

Browsers will reject this due to the security risk, and the RFC suggests that applications should not render the password (the part after the `:`) in clear text. -- Should fn:build-uri follow suite, or (along with fn:parse-uri) have an option to control the behaviour (keep, remove, invalid), where if the option is invalid, it will throw an fn:error?

ndw:
Fix #307

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -24412,8 +24412,7 @@ declare function fn:some(
       <fos:errors>
          <p>A dynamic error is raised <errorref class="UR" code="0001"/> if
          the URI contains an open square bracket in the authority component that
-         is not followed by a close square bracket.
-         </p>
+         is not followed by a close square bracket.</p>
       </fos:errors>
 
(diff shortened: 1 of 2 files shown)


ndw:
The CG accepted this PR at [meeting 025](https://qt4cg.org/meeting/minutes/2023/03-07.html)

PR #348, opened 2023-02-15
Attempt to build PR with merge-base version of master

ndw:
This PR changes the CI `build-pr.yml` script so that it checks out the version of master that the branch started from, rather than the current version of master, for building the specifications.

* Pro: we won't get build failures when the current master can't build the old version (for example, when images have been removed)
* Con: we won't get any features from the current master, such as stylesheet updates

Since failing builds are more troublesome than formatting issues, I'm going to say the pros outweigh the cons.


ndw:
There's no point waiting for this. It doesn't do anything.

PR #394, opened 2023-03-13
Minor correction to fn:parse-uri

ndw:
The `fn:parse-uri()` function recognizes "URIs" of the form `c:/path/to/thing` as implicitly being `file:` URIs. This small change adds a leading "/" to make the fact that it is a path explicit.


ndw:
Accepted at [meeting 027](https://qt4cg.org/meeting/minutes/2023/03-21.html#pr-394)

PR #395, opened 2023-03-13
Make the (non-)hierarchical nature of URIs explicit

ndw:
The `fn:parse-uri` function will parse hierarchical or non-hierarchical URIs, however, the parse cannot be reversed if the `fn:build-uri` function doesn't know whether the scheme is hierarchical. Consider `fn:parse-uri("querty:abc")`:

```
map {
  "path":"abc",
  "scheme":"querty",
  "path-segments":["abc"],
  "uri":"querty:abc"
}
```

When `fn:build-uri` parses that map, it produces: `querty://abc` because the scheme is not known to be non-hierarchical.

This PR changes `fn:parse-uri` so that it records whether or not the URI was hierarchical and `fn:build-uri` to use that information.

It's possible that we could finesse this by setting the `authority` to the empty string for hierarchical URIs, but it seems clearer to be explicit.

This PR also fixes a bug. Previously, if the `scheme` was not present when building a URI, the URI began with `//`. That's an error. If the scheme isn't present, there should be no *scheme separator*.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -24304,6 +24304,15 @@ declare function fn:some(
          <emph>string</emph> does not match either expression, the <emph>scheme</emph>
          is the empty sequence and the <emph>string</emph> is unchanged.</p>
 
+         <p>If the <emph>scheme</emph> is known to be hierarchical, or known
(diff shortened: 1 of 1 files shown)


ndw:
Accepted at [meeting 027](https://qt4cg.org/meeting/minutes/2023/03-21.html#pr-391)

PR #415, opened 2023-04-03
Revise parse/build URI functions for UNC names

### LINKED ISSUE #390: Should parsing and building URIs attempt to special case Windows URIs for UNC names?
Depending on the platform and language APIs involved, we see `file:` URIs encoded in a variety of different ways. It doesn't help that there's no official RFC for `file:` URIs.

* `file:/path/part` is a `file:` URI with no host and a path of `/path/part`.
* `file:///path.part` is a `file:` URI with an explicitly empty host and a path of `/path/part`.
* `file://path/part` is a `file:` URI with an authority of `path` and a path of `/part`. I think one common way to interpret this is as if it was `file:/part`. That is, in `file:` URIs, although a different host is possible, it's often just ignored.
* `c:\path\part` is most usefully interpreted as `file:/c:/path/part`, a `file:` URI with no host and a path of `/c:/path/part`. These are only going to be useful on a Windows system, so it isn't a problem to treat them the same way on all platforms. (Aside: I don't actually know if the path part should be `c:/path/part` instead, but it's currently got the leading slash in `fn:parse-uri()`.)

And then there's this: `file:////name/path/part`.

One interpretation is, "look, we accept `file:/` and `file:///` so let's just accept `file://` and `file://///////`, etc. as the same." And I think that's generally right, with the single special exception of `file:////`. The problem is that on Windows, this is a very common way to encode the URI for a UNC path, that is: `\\name\path\part` which is a Windows UNC path for `\path\part` on a host named `name` (via whatever networking protocol backs UNC).
 
You'd think that this should be `file://name/path/part`, but I think because browsers and maybe other tools just discard the authority part of a `file:` URI (or maybe because these are *paths* in some Windows sense?), that's not how they're encoded.

> Aside: Yes, I'm sure you also see `file:\\\\name\path\part` and `file:c:\path\part` and other forms as well. Those are out of scope, they're simply, flatly, completely wrong. You can't use `\` as a delimiter in a URI. RFC 3986 is authoritative on this point. Step one of dealing with random strings we think should be URIs is replacing all `\` with `/` because RFC 3986.

It's problematic to deal with `file:////` as a special case, but it's also problematic to leave out support for a common pattern on a widely deployed operating system.

Recognizing four slashes after `file:` and treating that specially isn't hard. The hard part is how do we encode this in the map that `fn:parse-uri` produces bearing in mind that the result should round-trip if you push it back through `fn:build-uri`.

Consider `file:////uncname/path/part`

Today, that is parsed as:

```
map {
  "uri": "file:////uncname/path/part",
  "scheme": "file",
  "authority": "uncname",
  "host": "uncname",
  "path": "/path/to/file",
  "path-segments": array { "", "path", "to", "file" }
}
```

and that doesn’t round trip. If you feed that to `fn:build-uri`, you get `file://uncname/path/part` and that absolutely doesn’t mean the same thing on a Windows machine.

We could encode the slashes in the `authority` in which case we also have to encode them in the `host` because in the presence of `host`, the `authority` isn’t used to by `fn:build-uri()`:

```
map {
  "uri": "file:////uncname/path/part",
  "scheme": "file",
  "authority": "////uncname",
  "host": "////uncname",
  "path": "/path/to/file",
  "path-segments": array { "", "path", "to", "file" }
}
```

It kind of works, but it’s really ugly and it means we have a host value that is a complete kludge. It doesn’t match the RFC rules for hostnames at all.

The other option that occurs to me is to add a “unc-path” property to the map:

```
map {
  "unc-path": true(),
  "uri": "file:////uncname/path/part",
  "scheme": "file",
  "authority": "uncname",
  "host": "uncname",
  "path": "/path/to/file",
  "path-segments": array { "", "path", "to", "file" }
}
```

That works but it introduces all sorts of possibilities for incoherent data, such as an `https:` URI with a `unc-path` flag set to `true()`.

What’s the right answer?

1. Ignore the UNC path special case, it’s the users problem to deal with them.
2. Recognize them, encode the details in the `authority` and `host`.
3. Recognize them, use a special property like `unc-path`.
4. Recognize them, and do this other much better idea I have: ________________

### LINKED ISSUE #398: User-defined functions clashing with constructor functions
There is no explicit rule in either XQuery or XSLT that a user-defined function must not clash (in name and arity) with a constructor function for an imported atomic type. It's implicit in the rule that you can't have two conflicting functions in the static context, but it would be helpful to say so explicitly and define an error code.

I have added tests to the XSLT3 and XQuery3 test suites.

See also Saxon bug 5921 - https://saxonica.plan.io/issues/5921

ndw:
Fix #398 
Fix #390 

This PR attempts to address the questions raised in issues 389 and 390:

1. It adds a `unc-path` option that is used to guide the parsing and construction of URIs that represent Windows UNC paths
2. It adds a `filepath` property to the result of `parse-uri`. This property represents the local path part of the URI. For file: URIs, this is the local path.
3. It addresses the use of "`|`" in URIs to represent the ":" in Windows filenames
4. It clarifies that percent-decoding a path also involves intepreting the result as a UTF-8 sequence
5. It clarifies that percend-decoding and encoding apply to the query parts of a URI as well.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -24596,6 +24596,14 @@ declare function fn:some(
                <fos:type>xs:boolean</fos:type>
                <fos:default>false()</fos:default>
             </fos:option>
+            <fos:option key="unc-path">
+               <fos:meaning>Indicates that an input URI that begins
+               with two or more leading slashes should be interprted
+               as a Windows Universal Naming Convention
+               Path. (Specifically: that it has the <code>file:</code> scheme.)</fos:meaning>
+               <fos:type>xs:boolean</fos:type>
+               <fos:default>false()</fos:default>
+            </fos:option>
          </fos:options>
 
          <p>Processing begins with a <emph>string</emph> that is equal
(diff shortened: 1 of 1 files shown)


ChristianGruen (commented):
Looks good! Maybe 1, 2 examples could be attached with the `filepath` key?

I noticed that some example results use `array { ... }` and others use `[ ... ]`. This could be unified (if the serialization method `adaptive` is used, you would get `[ ... ]`).

Next, there is »The map SHOULD only be populated with keys that have a non-empty value.« Maybe we… should… change it to MUST? The fewer MAYs and SHOULds, the better.

ndw:
Yes, there's a `FIXME:` comment in the history section about the fact that the examples need to be updated. I want to fix my implementation and then re-generate them.

I'd be happy saying that the map MUST only be populated with keys that have non-empty values.

ndw:
Just leaving this here because I'm not sure where else to put it. https://github.com/curl/trurl

ndw:
Accepted at [meeting 030](https://qt4cg.org/meeting/minutes/2023/04-11.html) on 11 April 2023.

PR #426, opened 2023-04-04
Resolve #280 by allowing a fragid

### LINKED ISSUE #280: Why is resolve-uri forbidden from resolving against a URI that contains a fragment identifier?
The 3.1 F&O spec says, of `fn:resolve-uri()`:

> A dynamic error is raised [[err:FORG0002](https://www.w3.org/XML/Group/qtspecs/specifications/xpath-functions-31/html/Overview.html#ERRFORG0002)] if $base is not a valid IRI according to the rules of RFC3987, extended with an implementation-defined subset of the extensions permitted in LEIRI, or if it is not a suitable IRI to use as input to the chosen resolution algorithm (for example, if it is a relative IRI reference, if it is a non-hierarchic URI, or **if it contains a fragment identifier**).

(emphasis added by me)

What in the name of all things is that about? I've never noticed that before, and I haven't seen any other API (the URL API in Node and the browser for example) that cares.

`resolve-uri('test.xml', 'http://example.com/path/file.xml#foo')` === `http://example.com/path/test.xml`

I don't see why the presence of a fragment identifier should matter in the least.

e.g.

```
>> let url = new URL("http://example.com/path/file.xml#foo")
>> console.log(url.href)
http://example.com/path/file.xml#foo
>> let resolved = new URL("test.xml", url);
>> console.log(resolved.href)
http://example.com/path/test.xml
```

ndw:
Fix #280

(Based off the right branch this time. I hope.)

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -5620,8 +5620,14 @@ Tak, tak, tak! - da kommen sie.
             not a valid IRI according to the rules of RFC3987, extended with an
             implementation-defined subset of the extensions permitted in LEIRI, or if it is not a
             suitable IRI to use as input to the chosen resolution algorithm (for example, if it is a
-            relative IRI reference, if it is a non-hierarchic URI, or if it contains a fragment
-            identifier). </p>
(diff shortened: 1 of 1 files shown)


ChristianGruen (approved):
Looks good.

PR #427, opened 2023-04-05
Change fn:transform to use the stylesheet's default mode

### LINKED ISSUE #315: fn:transform inconsistency: initial-mode
The `fn:transform` specification in F+O says that if no initial-mode is supplied, the unnamed mode is used.

The XSLT 3.0 specification says that if no initial mode is supplied, then the default mode is used if one has been specified, or the unnamed mode is used if not.

I think the XSLT 3.0 spec should win here: it makes more sense if a default has been declared that it should actually be used.

(Thanks to Amanda Galtman for pointing this out.)

ndw:
Close #315


michaelhkay:
Looks good. I've added a QT3 test transform-88, which Saxon already passes (it looks like we interpreted the spec this way).

PR #430, opened 2023-04-06
fn:doc et al, error handling: inconsistencies. Closes #293

### LINKED ISSUE #293: Error in fn:doc-available specification
Observed on the XML.com slack (Oct 6 2022):

A rule in the XQFO 3.1 specification seems to be inconsistent. In the error summary, it’s stated that:

> `err:FODC0005`, Invalid argument to fn:doc or fn:doc-available. Raised (optionally) by `fn:doc` and `fn:doc-available` if the argument is not a valid URI reference.

The rules for `fn:doc-available` say: “[…] In all other cases this function returns false. This includes the case where an invalid URI is supplied, and also the case where a valid relative URI reference is supplied, and cannot be resolved, for example because the static base URI is absent.”

ChristianGruen:
Resolves action item QT4CG-029-04.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -15073,7 +15073,7 @@ else
                spec="XP31" ref="dt-available-text-resources"
                >available text
                resources</xtermref> component of the dynamic context.</p>
-         <p>If the value of the <code>$href</code> argument is an empty sequence, the function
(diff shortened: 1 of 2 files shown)


michaelhkay (approved):

ndw:
@ChristianGruen I think this is ready for merging. Can you fix the merge conflicts for me?

ChristianGruen:
+1, I’ll do so.

ChristianGruen:
@ndw done

PR #444, opened 2023-04-18
Resolve merge conflict for PR 420

### LINKED ISSUE #357: Representing key-value pairs
A map can be decomposed into, or composed from, a sequence of key-value pairs (KVPs).

There are two natural representations of a key-value pair (K, V): it can be represented as a singleton map (`map{ K: V }`) or as a "doubleton" map (`map{ 'key': K, 'value': V}`).

This issue examines how well either of these representations is currently supported, which of them is preferable, and how this support should be improved.

I'll consider the following basic operations: constructing a KVP from a key and a value, assembling a map from a set of KVPs, decomposing a map into a sequence of KVPs, extracting the key from a KVP, and extracting the value from a KVP.

**Singleton Representation**

Constructing a KVP from a key and a value:

```
map{ $key : $value }
map:entry($key, $value)
<xsl:map:entry key="$key" select="$value"/>
```
Assembling a map from a set of KVPs

```
map:merge($kvps)
<xsl:map>
```

Decomposing a map into a sequence of KVPs:

`map:for-each($map, map:entry#2)`

Extracting the key from a KVP:

`map:keys($kvp)`

Extracting the value from a KVP:

`$kvp?*`

**Doubleton Representation**

Constructing a KVP from a key and a value:

```
map{ 'key': $key, 'value': $value }
```
Assembling a map from a set of KVPs

`map:build($kvps, ->{?key}, ->{?value})`

Decomposing a map into a sequence of KVPs:

`map:for-each($map, ->($K, $V){map{ 'key': $key, 'value': $value })`

Extracting the key from a KVP:

`$kvp?key`

Extracting the value from a KVP:

`$kvp?value`

**Analysis**

The singleton representation is better supported at present, and it makes sense therefore to fill in the gaps that currently make it awkward. The main attraction of the doubleton representation is the ease of extracting the key and the value using `$kvp?key` and `$kvp?value`. The equivalents for the singleton representation (`map:keys($kvp)` and `$kvp?*`) feel clumsy and unintuitive; however, it's not at all obvious what would be better, short of introducing new custom syntax, which seems over-the-top. The best idea I can come up with is to have two functions `map:key($kvp)` and `map:value($kvp)` which require $kvp to be a singleton map. But I hate the namespace prefixes...

The other thing needed to "fill the gaps" is a function `map:entries($map)` equivalent to `map:for-each($map, map:entry#2)`.

What if we chose to go the other way, and improve support for the doubleton representation?

We could add `map:key-value-pair($key, $value)` to create  KVP, and `map:of($kvps)` to build a map from a set of KVPs, and `map:key-value-pairs($map)` to decompose a map. The trickiest problem is what to do about XSLT, where the 3.0 instructions `<xsl:map>` and `<xsl:map-entry>` use the singleton representation.

ndw:
Close #420 
Close #357

### DIFF
--- a/specifications/css/xslt-40.css
+++ b/specifications/css/xslt-40.css
@@ -41,3 +41,16 @@ div.proto        {
 		background: #ffe6ff;
 		overflow: auto;
                  }
+
+table.proto tr td,
+table.record tr td, {
+  font-family: monospace;
+  padding-right: 0.5em;
+}
+table.proto tr.arg td:first-child,
+table.record tr.arg td:first-child  {
+  padding-left: 2em;
+}
+table.proto tr.name span.name {
+  font-weight: bold;
+}
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -813,7 +813,7 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
       <g:ref name="AssignmentExpr" if="scripting" lookahead="3"/>
       <g:ref name="ExitExpr" if="scripting" lookahead="2"/>
       <g:ref name="WhileExpr" if="scripting" lookahead="2"/>
-      <g:ref name="OrExpr" not-if="xpath40 xquery40"/>
+      <g:ref name="OrExpr" if="xpath40 xquery40"/>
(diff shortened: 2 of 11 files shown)


ndw:
No. I'm going to do this differently.

PR #455, opened 2023-04-25
410: Converting doubles to decimals, fractional digits

ChristianGruen:
@michaelhkay In this PR, I tried to undo the changes that were introduced to make comparisons transitive. I haven’t made any changes to `distinct-values` and `group by`, because I am uncertain if I have spotted all the relevant parts of the specification. Maybe/hopefully we can address them in a next step.

Any feedback is welcome.


michaelhkay:
I think we need to add the following change to distinct-values:

> Two items $J and $K in the input sequence (after atomization, as required by the function signature) are compared as follows:
> 
> (1) If both $J and $K are instances of xs:numeric, they are considered equal if fn:atomic-equal($J, $K) is true
> (2) Otherwise, they are considered equal if fn:deep-equal($J, $K, $coll) is true, where $coll is the collation selected according to the rules in [5.3.5 Choosing a collation]. This collation is used when string comparison is required.
> 

No further change is needed to `xsl:for-each-group`, because this already relies on `distinct-values`.

The rules for "group by" in FLOWR expressions should be changed to invoke `distinct-values`.

ndw:
> @ndw I assume that’s nothing I can fix?
> https://github.com/qt4cg/qtspecs/actions/runs/4798182033/jobs/8536145986#step:11:123

No, that’s weird. I’ve just started the workflow running again.
Hopefully it was transient.

                                        Be seeing you,
                                          norm

--
Norm Tovey-Walsh ***@***.***>
https://norm.tovey-walsh.com/

> Everyone has talent. What is rare is the courage to follow the talent
> to the dark place where it leads.--Erica Jong

ChristianGruen:
> I'm having trouble working out how your proposed rule for comparing numerics differs from my proposal to use atomic-equal(), or why you don't think that using atomic-equal() is suitable. I'm reluctant to have yet another rule if we can avoid it.

I still need to get used to the `atomic-equal` function. You’re of course right, it provides exactly the semantics we need. Thanks.

So my suggestion is basically the same, but it can be expressed shorter: We can keep `distinct-values` untouched if we modify `deep-equal` to use `atomic-equal`:

> d. If both the following conditions are true:
>   * $i1 is an instance of `xs:string` or `xs:untypedAtomic`.
>   * $i2 is an instance of `xs:string` or `xs:untypedAtomic`.
> 
> then the function `equal-strings($i1, $i2, $collation, $options)` returns true.
> Otherwise (at least one of the items is of some other type): `atomic-equal($i1, $i2)` returns true.

This way, we could keep the group by clause and the switch expression untouched (which both seem to rely on the collation in the static context), and the special `NaN` rule in the deep-equality rules can be dropped as well.

Next, it seems consistent to me to let `deep-equal(1.1, 1.1e0)` return `false` – even more so if `atomic-equal` now behaves that way.

ChristianGruen:
I’ve incorporated my latest proposal in the PR.

michaelhkay:
At present the text I'm seeing for distinct-values() hasn't restored transitivity. Am I reading the wrong version?

ChristianGruen:
@michaelhkay I have fixed the conflicts in this PR. Do you my approach to change the rules of `fn:deep-equal` (which is utilized by `fn:distinct-values` and other expressions) is feasible, or am I missing something?

ChristianGruen:
@ndw I believe the PR and the action item `QT4CG-034-01` can now be closed.

PR #462, opened 2023-04-27
434: Added examples for parse-integer()

Arithmeticus:
Supplemental examples related to pr #434 .

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -2185,6 +2185,19 @@
                <fos:expression>fn:parse-integer("vv", 32)</fos:expression>
                <fos:result>1023</fos:result>
             </fos:test>
+            <p>Alphabetic base-26 numbering systems (hexavigesimal) can be parsed via translation.
+               Note, enumerating systems that do not assign a symbol to zero (e.g., spreadsheet
+               columns) must be preprocessed in a different fashion.</p>
+            <fos:test>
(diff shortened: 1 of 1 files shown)


michaelhkay:
I'm not sure of the rationale for some of these changes, which go well beyond adding examples.

- The ability to include underscores in the string has been dropped
- There is no longer a clear statement that there must be at least one digit
- The indication that the upper-case equivalents of a-z are A-Z (which I added in response to a comment from MSpMcQ stating that this couldn't be taken for granted) has disappeared.

I'm actually wondering if the text (other than examples) has accidentally reverted to a previous draft?

ndw:
That diff markup version does look weird. Consequence of other accepted PRs, I think. The diffs are supposed to be against the version that was the source of the fork, but that certainly doesn't look like what's going on here. The actual PR diff is just a single hexavigesimal example.

Arithmeticus:
PR 462 was intended to be some simple supplements to what was, at the time, the main/master trunk. So any loss of text is due to version conflict.

Please advise on how to avoid this in the future.

ChristianGruen:
> Please advise on how to avoid this in the future.

It might help to invoke `git rebase master` (and to hope that there won't be any conflicts).

gimsieke:
Ah, the magic incantations of the early 21st century.

ndw:
You could try rebasing on the current master and pushing it again. But I don't think this is your fault. I'm a little perplexed about how this happened, but it's in the build tools, not your PR.

PR #465, opened 2023-05-02
80: fn:iterate-while: Examples revised

ChristianGruen:
This PR is editorial. I’ve reformatted the examples for `fn:iterate-while`…

image

…to make them better readable.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -17270,38 +17270,32 @@ declare function fn:iterate-while(
             </fos:test>
          </fos:example>
          <fos:example>
-            <fos:test>
-               <fos:expression><![CDATA[let $input := (0 to 4, 6 to 10)
-return fn:iterate-while(0, function($n) { $n = $input }, function($n) { $n + 1 })]]></fos:expression>
-               <fos:result>5</fos:result>
-               <fos:postamble>This returns the first positive number missing in a sequence.</fos:postamble>
-            </fos:test>
+            <p>Return the first positive number that is missing in a sequence (the result is 5):</p>
+            <eg><![CDATA[
+let $input := (0 to 4, 6 to 10)
+return fn:iterate-while(0, function($n) { $n = $input }, function($n) { $n + 1 })]]></eg>
          </fos:example>
          <fos:example>
-            <fos:test>
-               <fos:expression><![CDATA[fn:iterate-while(
+            <p>Return the first number in a sequence greater than 5 and all remaining numbers:</p>
+            <eg><![CDATA[
+fn:iterate-while(
   1 to 9,
-  function($seq) { head($seq) < 5 },
+  function($seq) { head($seq) <= 5 },
   function($seq) { tail($seq) }
-)]]></fos:expression>
-               <fos:result>(5, 6, 7, 8, 9)</fos:result>
-            </fos:test>
+)]]></eg>
          </fos:example>
          <fos:example>
-            <fos:test>
-               <fos:expression><![CDATA[let $input := 3936256
+            <p>Approximate the square root of a number:</p>
+            <eg><![CDATA[
+let $input := 3936256
 return fn:iterate-while(
   $input,
   function($result) { abs($result * $result - $input) >= 0.0000000001 },
   function($guess) { ($guess + $input div $guess) div 2 }
-)]]></fos:expression>
-               <fos:result>1984</fos:result>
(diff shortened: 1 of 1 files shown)


michaelhkay:
While the examples are now more readable, they seem to be no longer executable.

ChristianGruen:
> While the examples are now more readable, they seem to be no longer executable.

They look ok to me. Which of the queries fails?

michaelhkay:
As far as I can see the examples no longer use `fos:test` markup, which means they won't be tested.

ChristianGruen:
I understand. So there is some automated testing of the queries in the specification?

@ndw There seem to be a bunch of query strings in the spec that stretch across multiple lines, and that are difficult to read when presented as plain text. Would it be possible to change the rendering of those example queries that contain newlines?

ndw:
Difficult to read generally, or difficult to read in the diff versions?

I know the diff versions have trouble with correct rendering of newlines in `pre` elements. It's fixable ("it's just a simple matter of programming"), but it's a little bit complicated. I can take a look. Pointers to specific examples that are hard to read most appreciated.

ChristianGruen:
> Difficult to read generally

Generally; see e.g. the screenshot in the initial comment of this issue or some more complex examples in the spec (e.g. https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-function-lookup) – mostly those ones that are not limited to a simple function call and include let/for clauses.

It could certainly be worse, so… Just ignore it if it turns out to be a bigger issue. Thanks!

michaelhkay:
We should improve the markup design and/or the stylesheets so that we can achieve readability and testability at the same time.

We haven't run the stylesheet that generates test cases from the examples for some time. It's difficult to do such testing when the spec is typically ahead of implementations that can be used for the testing; this certainly makes it difficult to do continuous testing. But we should probably take a snapshot soon.

ndw:
Well, `code` is inline markup:
```xml
         <fos:example>
            <p>The expression <code>let $f := fn:function-lookup(xs:QName('zip:binary-entry'), 2)
                  return if (exists($f)) then $f($href, $entry) else ()</code> returns the result of
               calling <code>zip:binary-entry($href, $entry)</code> if the function is available, or
               an empty sequence otherwise.</p>
         </fos:example>
```
I don't think we'd like the results if we just made line breaks in `code` significant. If you want to have a code block, use `eg`:

```xml
         <fos:example>
            <p>The expression
         <eg>let $f := fn:function-lookup(xs:QName('zip:binary-entry'), 2)
return if (exists($f)) then $f($href, $entry) else ()</eg>
returns the result of
               calling <code>zip:binary-entry($href, $entry)</code> if the function is available, or
               an empty sequence otherwise.</p>
         </fos:example>
```
to get:

eg

I think the CSS could use a little work, but is that what you were after, or am I misunderstanding?

michaelhkay:
The issue is that it's no longer marked up with `fos:test`, which means that when we regenerate the tests to test the examples, it won't be tested.

ChristianGruen:
@ndw Right (sort of). Here’s an example from the current spec:

```xml
<fos:example>
  <fos:test>
     <fos:expression><![CDATA[let $input := 3936256
return fn:iterate-while(
$input,
function($result) { abs($result * $result - $input) >= 0.0000000001 },
function($guess) { ($guess + $input div $guess) div 2 }
)]]></fos:expression>
     <fos:result>1984</fos:result>
     <fos:postamble>This computes the square root of a number.</fos:postamble>
  </fos:test>
</fos:example>
```

…which is rendered as follows:

image

I found this hard to read, so in this PR I changed the XML to…

```xml
<fos:example>
  <p>Approximate the square root of a number:</p>
  <eg><![CDATA[
let $input := 3936256
return fn:iterate-while(
$input,
function($result) { abs($result * $result - $input) >= 0.0000000001 },
function($guess) { ($guess + $input div $guess) div 2 }
)]]></eg>
</fos:example>
```

…to get:

image

Michael indicated that the new version cannot be tested automatically, so I wondered if we can stick with the original XML solution, and if `fos:example` can be output differently if the text contains newlines, e.g. by using `<pre>` instead of `<code>`, to get something like:

---------

The expression

```
let $input := 3936256
return fn:iterate-while(
  $input,
  function($result) { abs($result * $result - $input) >= 0.0000000001 },
  function($guess) { ($guess + $input div $guess) div 2 }
)
```

returns `1984`. (*This computes the square root of a number.*)

ndw:
See #474

ChristianGruen:
Superseded (see #474).

Issue #471, opened 2023-05-04
Unify formatting (function calls, code blocks, quotes) in the specification


ChristianGruen:
**Todos (2023-05-18)**:

* [x] Initial cleanup of `fn:` prefixes → [#475](https://github.com/qt4cg/qtspecs/pull/475)
* [x] Remove more `fn:` prefixes → [#509](https://github.com/qt4cg/qtspecs/pull/509)
  * `fn:` prefixes in examples that raise an error.
  * `fn:` prefixes in `eg` code blocks
  * other documents: `expressions.xml`, `query-examples.xml`, …
* [x] Render `false`, `true`, `NaN`, `INF`, `+INF` as code → [#510](https://github.com/qt4cg/qtspecs/pull/510)
* [x] Render string values as code and use quotes: `"yes"`, `"true"`, `"0"` → [#511](https://github.com/qt4cg/qtspecs/pull/511)
* [x] Omit quotes for single characters: `\b`, `\f`, … → [#511](https://github.com/qt4cg/qtspecs/pull/511)
* [x] Rewrite simple to typographic quotes → [#511](https://github.com/qt4cg/qtspecs/pull/511)

-----------

Inspired by https://github.com/qt4cg/qtspecs/pull/454#issuecomment-1534633089 ff.

The syntax of the examples in the XQFO specification is inconsistent. Sometimes, functions in the standard function namespace have an `fn` prefix…

* `fn:fold-right(1 to 5, "", fn:concat(?, ".", ?))`
* `fn:substring("motor car", 6)`

…sometimes they don’t…

* `data(123)`
* `concat("http://www.example.com/", encode-for-uri("~bébé"))`

…sometimes it’s both:

* `fn:fold-right(1 to 5, "$zero", concat("$f(", ?, ", ", ?, ")"))`
* `fn:concat(01, 02, 03, 04, true())`
* `fn:tokenize(fn:unparsed-text($href), '\r\n|\r|\n')[not(position()=last() and .='')]`

Should we drop or keep the prefix – or doesn’t it really matter? If there’s interest, I can create a PR (I’d tend to drop the prefixes).

In addition, there doesn’t seem to be a consistent rule for representing booleans. We have:

Syntax | Comment
--- | ---
…returns false | mostly used in the rules (seems appropriate to me)
…returns `false` | used in the rules; maybe we should use replace it with the first syntax?
…the result is `fn:false()` | used in the rules; maybe we should use replace it with the first syntax?
[...]

PR #474, opened 2023-05-05
Per comments on #465, improve presentation of multi-line expressions

ndw:
This PR addresses points raised in the comments on #465.

If an `fos:expression` element is a code block, nest an `eg` inside it:

```xml
               <fos:expression><eg><![CDATA[let $input := 3936256
return fn:iterate-while(
  $input,
  function($result) { abs($result * $result - $input) >= 0.0000000001 },
  function($guess) { ($guess + $input div $guess) div 2 }
)]]></eg></fos:expression>
```

I found four examples where an `fos:expression` contained more than one newline and I added `eg` wrappers in those cases. 

There are many more `fos:expression` elements that contain a single newline, but automatically formatting them as code blocks was often unsuccessful. Many of those cases seem to be just newlines entered for authoring convenience.

I also fixed the CSS for code blocks and attempte to remove trailing newlines from code blocks.

### DIFF
--- a/specifications/css/qtspecs.css
+++ b/specifications/css/qtspecs.css
@@ -70,7 +70,7 @@ div.exampleInner { background-color: #d5dee3;
                    border-bottom-style: double;
                    border-bottom-color: #d3d3d3;
                    padding: 4px;
-		   margin-bottom: 4px;
+		   margin-bottom: 1em;
                  }
 
 div.issueBody    { margin-left: 0.25in;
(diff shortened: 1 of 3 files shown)


ChristianGruen:
Excellent! With the new formatting, it will be easier to create new multiline examples and reformat existing ones (possibly as part of #471).

ChristianGruen (approved):

ndw:
I'm going to call this non-controversial and merge it.

PR #475, opened 2023-05-05
471: fn: prefix removed from function calls in the examples

ChristianGruen:
#471: I’ve removed the `fn:` prefixes from the function calls in the examples.

I have left pretty much all `true`/`false` strings untouched, since I’m not sure what would be the most consistent approach to clean them up. It will be better anyway to create a separate PR for that.


ChristianGruen:
It no one objects, maybe we can merge this PR earlier than others to avoid conflicts with upcoming commits.

PR #478, opened 2023-05-07
467: map:keys-where

ChristianGruen:


michaelhkay:
The proposed spec is technically sound, but I do wonder whether we need both `map:keys-where()` and `map:filter()`. It seems to me that `map:filter()` is more general because the predicate has access to both the key and the value. The main argument for `map:keys-where()` seems to be analogy with `fn:index-where` and `array:index-where`; but those two functions meet a requirement that `fn:filter` and `array:filter` don't satisfy because the filter functions don't retain the original index of the selected entries. In the case of `map:filter()` its selection capability is more powerful than `map:keys-where()` and it returns all the information that `map:keys-where()` returns.

ChristianGruen:
> The proposed spec is technically sound, but I do wonder whether we need both `map:keys-where()` and `map:filter()`. It seems to me that `map:filter()` is more general because the predicate has access to both the key and the value. The main argument for `map:keys-where()` seems to be analogy with `fn:index-where` and `array:index-where`; but those two functions meet a requirement that `fn:filter` and `array:filter` don't satisfy because the filter functions don't retain the original index of the selected entries. In the case of `map:filter()` its selection capability is more powerful than `map:keys-where()` and it returns all the information that `map:keys-where()` returns.

I wonder if users think of *filtering a map* if they want to look up keys. For that case, I think that the existing `map:for-each` function is even a better match:

```xquery
map:for-each  ($map, function($k, $v) { $k[$v = 5] })
map:filter    ($map, function($k, $v) { $v = 5     }) => map:keys()
map:keys-where($map, function(    $v) { $v = 5     })
```

But I agree it may be critical to flood users with too many new functions. Next, thinking of performance, it may be better anyway to motivate users to create inverted maps instead of supporting value lookups.

ndw:
Accepted at [meeting 034, 16 May 2023](https://qt4cg.org/meeting/minutes/2023/05-16.html)

PR #483, opened 2023-05-08
452: window: make 'start' and 'when' optional

ChristianGruen:


michaelhkay:
A minor quibble: in the sentence "The end clause can only be omitted in [4.14.4.1 Tumbling Windows]"

(a) "can only be omitted" should be "can be omitted only"

(b) we shouldn't be using a specref like this: it's a reference to the concept not to the section in which it is described. Perhaps we should use an `<nt>` ref here, "can be omitted only within a `TumblingWindowClause`.

michaelhkay:
Could you add an entry to Appendix K please?

ChristianGruen:
…done

ndw:
Can you resolve this merge conflict, please?

Accepted at [meeting 034, 16 May 2023](https://qt4cg.org/meeting/minutes/2023/05-16.html)

PR #492, opened 2023-05-11
Fix examples, change filepath definition slightly

ndw:
This PR fixes the examples in the `parse-uri()` function. It also makes a small change to the `filepath` property, eliding it when the scheme is known not to be file.


ChristianGruen (commented):
@ndw It seems that simple (0022) instead of typographic (201C/201D) quotation marks are used in most texts in the spec. Would you mind changing that in your sections for `fn:parse-uri`?

Related: https://github.com/qt4cg/qtspecs/issues/471#issuecomment-1538709138

ndw:
It makes me feel dirty, but okay.

ChristianGruen:
> It makes me feel dirty, but okay.

;) Would be nice to replace all quotes with typographical ones, but it’s probably too ambitious. If all code snippets are consistently marked as `code` in the future, it could be automatized.

PR #493, opened 2023-05-11
32: try/catch: New variable for all error information

ChristianGruen:


ChristianGruen:
Probably not me causing this: https://www.githubstatus.com/

michaelhkay:
I wonder if rather than having a map entry $err:map?additional` containing the same information as `$err:additional`, we should encourage implementations to put additional information in separate map entries with meaningful keys?

I wonder also if we should provide a better way of re-throwing the error, for example `rethrow($err as map(*))`.

ndw:
It seems to me that `$err:map?moon-phase` would be better than `($err:map?additional)?moon-phase`.¹

And yes, I like the rethrow suggestion.

---

¹ Are those parentheses necessary in that expression? Perhaps not. Probably not. Am I going to try to internalize that? No I'm just going to do the easy, safe thing. Dont @ me.

ChristianGruen:
> I wonder if rather than having a map entry $err:map?additional`containing the same information as`$err:additional`, we should encourage implementations to put additional information in separate map entries with meaningful keys?

As this is implementation-defined, we could try to find out if it used at all:

* If yes, and if it seems helpful to everyone else, we could introduce official bindings for it.
* If no, I would tend not to give it too much consideration.

But it would be easy to revise the text.

> I wonder also if we should provide a better way of re-throwing the error, for example `rethrow($err as map(*))`.

Sounds good ෴ #501

ndw:
Accepted at meeting 034, 15 May 2023.

PR #494, opened 2023-05-11
Remove legacy materials from the working master branch

ndw:
This is *intended to be* an entirely uninteresting change. This PR removes a whole bunch of historical artifacts from the `master` branch, things like the requirements and use-cases documents that we aren't maintaining for QT4, the errata which don't apply to QT4, etc.

I will push a separate branch, `legacy-documentation`, to the repository that contains all of the the files removed by this PR so that the aren't lost and can easily be recovered. (I won't do that as a PR, I'll just push it to the repository.)

In the meantime, I think this trimmed down master branch works just fine and it's a lot simpler and easier to explain.

This PR *does* remove support for the legacy ant builds, but I doubt they've worked for a while now.


ndw:
Bah humbug. The github action falls over because it expects some directories to exist that have been removed. I'll have to clean them up in a separate PR that's merged first.

In the meantime, I think this will work fine for ordinary users.

PR #498, opened 2023-05-12
Fix typo, replace sequence-of with sequence

### LINKED ISSUE #495: separator example in https://qt4cg.org/specifications/xslt-40/Overview-diff.html#for-each-separator has xsl:sequence-of instead of xsl:sequence element
While looking through the XSLT 4 draft spec, I have found the following example in https://qt4cg.org/specifications/xslt-40/Overview-diff.html#for-each-separator:

```
<xsl:for-each select="6, 3, 9" separator="|">
   <xsl:sort select="."/>
   <xsl:sequence-of select="., .+1"/>
</xsl:for-each>
```

`xsl:sequence-of` should be `xsl:sequence`.

ndw:
Fix #495

### DIFF
--- a/specifications/xslt-40/src/xslt.xml
+++ b/specifications/xslt-40/src/xslt.xml
@@ -14305,7 +14305,7 @@ and <code>version="1.0"</code> otherwise.</p>
             <eg><![CDATA[
 <xsl:for-each select="6, 3, 9" separator="|">
    <xsl:sort select="."/>
-   <xsl:sequence-of select="., .+1"/>
+   <xsl:sequence select="., .+1"/>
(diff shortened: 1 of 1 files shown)


innovimax:
@ndw is there any reason this has to be fixed multiple times ?
https://github.com/qt4cg/qtspecs/pull/422

ndw:
No good reason.

Looking at the history, apparently I reverted the fix by mistake while correcting merge conflicts in #420 for Mike. We've been trying to streamline things, so that PRs don't stay open so long, so that merge conflicts are less likely, but mistakes happen.

innovimax:
LGTM

PR #502, opened 2023-05-16
Fix typographic quotes in XPath Data Model

### LINKED ISSUE #500: Fix errant typographic quotes in XPath Data Model
In a comment on #471, @ChristianGruen observes that there are some errant typographic quotes in code examples in the XPath Data Model specification. I assume these are errors, mostly likely on my part, and should be corrected.

ndw:
Close #500 

On closer inspection, there were only a few places where typographic quotes *were not* used in prose. I've fixed those. I think the DM spec could probably use an editorial pass to add `code` around some literals, but I'm not doing that in this PR.

I've left typographic quotes around code and literals because I don't think straight quotes would be an improvement: the literal `“<code>3</code>”` doesn't need straight quotes because the quotes are not part of the literal.

### DIFF
--- a/specifications/xpath-datamodel-40/src/attribute.xml
+++ b/specifications/xpath-datamodel-40/src/attribute.xml
@@ -48,7 +48,7 @@ present then a prefix <rfc2119>must</rfc2119> also be present.
 </olist>
 
 <p>For convenience, the &elementNode; that owns this attribute is called
-its "parent" even though an &attributeNode; is not a "child" of its
+its “parent” even though an &attributeNode; is not a “child” of its
 parent element. </p>
 
 </div3>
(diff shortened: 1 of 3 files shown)


ChristianGruen (commented):
Thanks, Norm. Looks good.

Regarding quoted strings, I remember that Michael suggested rendering `xs:string` values with both the content and the quotes in code font (https://github.com/qt4cg/qtspecs/issues/471#issuecomment-1537534617); but that would certainly affect many other sections as well.

ndw:
I tried that in one place in this commit, actually (the last change in the DM spec). I think Michael's suggestion is a good one:
```
The string value <code>"test"</code>
```
But I still think it's ok to use typographic quotes when the quotes aren't intrinsic to the value, like the "3" example above.

ChristianGruen (approved):

ChristianGruen (approved):

PR #511, opened 2023-05-19
471: <code> elements, simple/typographic quotes

### LINKED ISSUE #471: Unify formatting (function calls, code blocks, quotes) in the specification

ChristianGruen:
That was a work-intensive one, as expected, but I’m optimistic that the PR improves the overall situation.
I’ll be happy to see subsequent PRs if I missed something (e.g., I didn’t touch `ebnf.xml`).

Closes #471

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -641,9 +641,10 @@
          <fos:example>
             <p>Consider a situation in which a user wants to investigate the actual value passed to
                a function. Assume that in a particular execution, <code>$v</code> is an
-                  <code>xs:decimal</code> with value <code>124.84</code>. Writing <code>fn:trace($v,
-                  'the value of $v is:')</code> will put the strings <code>"124.84"</code> and
-                  <code>"the value of $v is:"</code> in the trace data set in implementation
+                  <code>xs:decimal</code> with value <code>124.84</code>.
+                  Writing <code>fn:trace($v, 'the value of $v is:')</code> will put the strings
+                  <code>"124.84"</code> and <code>"the value of $v is:"</code>
+                  in the trace data set in implementation
                dependent order.</p>
          </fos:example>
       </fos:examples>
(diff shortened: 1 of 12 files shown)


ndw:
There are *a lot* of changes in here. (Thank you, @ChristianGruen ). 

I believe that this PR is entirely editorial and it's work we've agreed to do. I'm tempted to merge it as quickly as possible because it's a nest of merge conflicts waiting to happen.

Anyone disagree?

PR #538, opened 2023-06-06
480: Attempt to allow xs:string to be 'promoted to' xs:anyURI

ndw:
I think it might be a little cheeky to call it "promotion" in both directions, but this really has more to do with a kind of conversion, so I'm willing to let it slide.

If accepted, this PR resolves ACTION QT4CG-035-03.

### DIFF
--- a/specifications/xquery-40/src/back-matter.xml
+++ b/specifications/xquery-40/src/back-matter.xml
@@ -23,8 +23,31 @@ of the types <code>xs:float</code> or <code>xs:double</code>.  The
 result of this promotion is created by casting the original value to
 the required type. This kind of promotion may cause loss of
 precision.</p></item></olist></item>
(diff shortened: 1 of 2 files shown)


michaelhkay:
I'm not sure about the sentence "a value of type <code>xs:string</code> can be promoted to any
 <code>xs:anyURI</code> type". What exactly do we mean by "any xs:anyURI type"? In other cases we only allow promotion to xs:double, xs:float, xs:string etc, not to derived types.

The available type promotions are referenced in §4.4.3 Coercion Rules, which needs to be revised to add another permitted promotion (or to be generalized to cover all permitted promotions without enumerating them).

In fact I think the coercion rules as written leave it a little vague as to whether promotion and downcasting can be combined. If the required type is a subtype of xs:double allowing only positive values (say my:positiveDouble), and the supplied value is +1, can you promote the supplied integer to xs:double and then "relabel" it as my:positiveDouble?

ndw:
Yeah. I said it seemed a little cheeky. I'll see what I can do.

michaelhkay (approved):

michaelhkay:
I believe I have absorbed this material into PR #815 and this one is now obsolete.

ndw:
Supersceded by #815

PR #546, opened 2023-06-12
414: Attempt to implement expanding the allowed character repertoire

### LINKED ISSUE #414: Lift character set restriction of xs:string
Adopted from https://github.com/qt4cg/qtspecs/issues/413#issuecomment-1491469514

> I guess that raises the question of whether it is still appropriate to restrict the character set of xs:string to that of XML 1.0. Are there any benefits in doing so?

I believe that would simplify things a lot, in particular when working with input/output functions.

ndw:
Fix #414 

This PR addresses ACTION QT4CG-036-01 on me.

### DIFF
--- a/specifications/xpath-datamodel-40/src/xpath-datamodel.xml
+++ b/specifications/xpath-datamodel-40/src/xpath-datamodel.xml
@@ -955,10 +955,10 @@ is a sequence of zero or more
 <termref def="dt-character">characters</termref>.</termdef></p>
 
 <p diff="chg" at="2023-06-12"><termdef term="character" id="dt-character">A
-<term>character</term> is any Unicode character.
-</termdef> Implementations <rfc2119>may</rfc2119> restrict characters
+<term>character</term> is any Unicode character.</termdef>
+Implementations <rfc2119>may</rfc2119> restrict characters
 to those Unicode characters allowed by the <code>Char</code>
-production in <bibref ref="xml"/>.</p>
+production in <bibref ref="xml"/>. Unpaired surrogates are always forbidden.</p>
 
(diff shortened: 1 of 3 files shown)


ndw:
Yes, good suggestions.

Arithmeticus:
For consistency, the type of edits applied in this commit also need to be applied to the following (my comments based upon the qt4cg diff html files for this PR):

- [1.8.1](https://qt4cg.org/pr/546/xpath-functions-40/autodiff.html#character-terminology) -- the data model edits need to be applied to their summary counterparts. 
- 1.8.1 first note: these substantive restrictions would be better placed in the data model document, particularly material referencing the surrogate blocks.
- 1.8.1 second note: I think the revisions mean that the set of codepoints and the set of characters are coterminous, no? (With the proviso that an implementer may choose to not make them so.) Or maybe I misunderstand.
- first note for `fn:unparsed-text-available()`: "no characters that are invalid in XML"

ndw:
I managed to lose all my commits and GitHub thoughtfully closed the issue for me. :-/

I've reapplied the changes and reopened the PR.

@michaelhkay I've added a note about unpaired surrogates. I don't believe any new errors are required in serialization, there's already an error SER0006 for "character not allowed in this version of XML" which I think applies.

ChristianGruen (approved):

Arithmeticus:
@ndw You may also want to adjust the language in xpath-datamodel.xml at lines 932-934:
> This means, for example, that the xs:string data type should (at the time of writing) support the set of characters defined by the Char production in XML 1.1 Second Edition.

The relative parenthetical clause gives me slight chills, as well.

ndw:

Issue #464, opened 2023-04-29
Serialization sequence normalization step 3 needs clarification


Arithmeticus:
The specifications currently read:

> If the item-separator serialization parameter is absent, then for each subsequence of adjacent strings in S2, copy a single string to the new sequence equal to the values of the strings in the subsequence concatenated in order, each separated by a single space. Copy all other items to the new sequence. Otherwise, copy each item in S2 to the new sequence, inserting between each pair of items a string whose value is equal to the value of the item-separator parameter. The new sequence is S3.

As written ("If...then.... Otherwise...."), this implies that the process whereby a block of adjacent strings are joined into a single string is performed only when the item separator parameter is absent. I.e., if the item-separator parameter is not absent, it will not be used to string-join adjacent groups of strings. Perhaps that is as intended, but I want to make sure.

Also, as written, this implies that when the parameter is absent, the sequence begins by finding all subsequences of adjacent strings, performing concatenation. Then all non-strings are added to the sequence. That seems wrong, because it appears to advise the processor to rearrange the sequence of input items.

I propose a revision along these lines:
> Copy each item in S2 to a new sequence. If a given pair of adjacent items are both strings, then separate them with a string whose value is equal to the value of the item-separator parameter or is a single space if the item-separator parameter is absent. If a given pair of adjacent items are not both strings, insert between each pair of items a string whose value is equal to the value of the item-separator parameter. Once this is finished, take each adjacent group of strings and concatenate them into a single string. The new sequence is S3.

My revision is based upon what I imagine happens, but implementers will know better than I.

I am working on some editorial touch-ups of the Serialization specifications, and can incorporate comments/suggestions made in this thread in that larger enterprise.

michaelhkay:
I agree this text can easily be misread, and I don't think your reading is quite correct.

It's really easiest to explain this by example, and the spec could probably do with more examples.

Consider the query

<a/>,<b/>,3 to 5,<c/>,6 to 8

With an item separator = "|" it serializes as

<a/>|<b/>|3|4|5|<c/>|6|7|8

Here the separator is inserted between all items, including nodes as well as atomic values. The sequence S3 contains 17 items, 9 from the original sequence, and 8 separators.

without an item-separator, this serializes (with the XML method) as

<a/><b/>3 4 5<c/>6 7 8

that is, a single space is inserted between adjacent atomic values (which by now have been converted to strings).

[...]

PR #555, opened 2023-06-16
464: Revised narrative of normalization steps for serialization

Arithmeticus:
This PR acts on #464 by revising the description of steps involved in normalizing a sequence that is input to serialization. I normally would wield a light editorial hand, but the issues raised in #464 as well as closer reading of the prose convinced me that a wholesale revision would be beneficial. For example, new sequences were described as if the reader already knew about them, but they are really only introduced as the last sentence in many steps.

I have capitalized on the original version of step 1's appeal to `array:flatten` to abbreviate the description for two of the steps.


ndw:
At [meeting 041](https://qt4cg.org/meeting/minutes/2023/07-11.html#open-pull-requests), the CG
agreed to accept this PR without further discussion.

Issue #361, opened 2023-02-20
Named arguments: $input vs. $value


ChristianGruen:
Great effort has been made in unifying the parameter names of the XQFO standard; thanks for that!

I believe to remember that:

* `$value`, `$values`, `$value1`, etc. is used for atomic/atomized arguments, whereas
* `$input`, `$input1`, etc. is used for input, mostly of type `item()`, that is processed unchanged.
* `$uri` is used for arguments that could have been defined as items of type `xs:anyURI`.

I believe the following argument names need to be double-checked (if not, It may be that I haven’t fully grasped how the naming rules are supposed to work):

Function | Currently | Presumably | Justification
--- | --- | --- | ---
`array:slice` | `$input` | `$array` | Alignment with `array:size` et al.
`trace` | `$value` | `$input` | Argument is not atomized
`json` | `$input` | `$value` | Argument is atomized
`string` | `$item` | `$value`  | `$item` is used nowhere else
`expanded-QName` | `$qname` | `$value` | Alignment with `prefix-from-QName` et al.
`resolve-QName` | `$qname` | `$value` | Alignment with `prefix-from-QName` et al.
`parse-QName` | `$eqname` | `$value` | Alignment with `parse-xml` et al.
`parse-json` | `$json` | `$value` | Alignment with `parse-xml` et al.
`json-to-xml` | `$json` | `$value` | Alignment with `parse-xml` et al.
`char` | `$name` | `$value` | Input may also be codepoint values, etc.
`namespace-uri-for-prefix` | `$prefix` | `$value` | `$prefix` is used nowhere else
`resolve-uri` | `$relative` | `$uri` | Absolute URIs are legal as well
`array:append` | `$add` | `$member` | Alignment with `array:put`

And we should probably pay particular attention to the naming conventions when adding new functions.

ChristianGruen:
@michaelhkay Do you believe the suggested amendment would be reasonable and consistent?

michaelhkay:
It's moving in a slightly different direction from what I originally envisaged, which was to use a specific name describing the role of the argument unless I couldn't think of anything suitable; but I have no objection to these changes.

ChristianGruen:
Thanks. I’ve created a PR.

> It's moving in a slightly different direction from what I originally envisaged, which was to use a specific name describing the role of the argument unless I couldn't think of anything suitable; but I have no objection to these changes.

[...]

PR #562, opened 2023-06-21
361: Named arguments: $input vs. $value

ChristianGruen:
#361: Just an editorial one to get one more issue closed.


ChristianGruen:
> In resolve-uri(), I'm not sure $uri is a good name for an argument that expects a relative URI reference. Technically, despite common usage, a relative reference is not a URI. Perhaps $href?

True, `$href` could be a viable alternative (I chose `$uri` because absolute URIs are legal inputs). I’ll change that.

ChristianGruen:
(updated, can be merged)

ndw:
The CG agreed to merge this PR at [meeting 040](https://qt4cg.org/meeting/minutes/2023/06-27.html).

michaelhkay:
Unfortunately the stylesheet for generating the tests in misc-BuiltInKeywords used the parameter names as a hint for what value to generate, eg. if the parameter was named "json" then the value in the test case would be valid JSON. We're going to have to make it smarter.

Issue #566, opened 2023-06-22
fn:parse-uri, fn:build-uri: Feedback


ChristianGruen:
Feedback on `fn:parse-uri` (thanks, @ndw, for the comprehensible rules in the spec):

1. With the current rules, the port is not detected in `http://x:80`. Maybe `x:` is misinterpreted as the beginning of a Windows path?
2. The *uri decoding* should be revised. It’s currently possible to create strings with invalid Unicode characters: `parse-uri('%FF')`. With the current rules, the following expression returns `df83` and `dc00`:

```xquery
parse-uri('%FF00')
=!> map:get('filepath')
=!> string-to-codepoints()
=!> format-integer('16^XX')
```

More to come (or not).

ChristianGruen:
3. Most regex strings ends with a dollar sign, but only some of them start with a circumflex character. Maybe it’s better to use `^` everywhere?

4. I believe that either the test cases or the specification needs to be aligned. The `path` in the test result often starts with a slash `/`, e.g. in [fn-parse-uri-016](https://github.com/qt4cg/qt4tests/blob/70b0839940503210421fbe43fd4e0b6363249e18/fn/parse-uri.xml#L305-L319). According to the rules, leading slashes are removed:
    * If the *string* matches `^[a-zA-Z][:|].*$`: […] a leading slash `/` is added to the *string*.
    * …
    * If the *string* matches `^//*([a-zA-Z]:.*)$`, the *authority* is empty and the *string* is the first match group.
    * …
    * If the *string* is the empty string, then *path* is the empty sequence, otherwise path is the whole *string*.

ChristianGruen:
I noticed that the rules for `fn:build-uri` include invocations of `fn:encode-for-uri`.

The initial suggestion in #72 was to provide a function for decoding a URI. Maybe we should still think about adding a `fn:decode-uri` or `fn:decode-from-uri` function, in which we could tackle the open issues that need to be solved for 2. The URI decoding rules for `fn:parse-uri` could then be replaced with a reference to the new function.

ChristianGruen:
5. I would prefer to have redundant data removed from the result of `fn:parse-uri`. Otherwise it will be difficult for the user to understand which information is needed to construct URIs via `fn:build-uri` (without reading the spec in full detail). In particular, I find `uri` confusing because it’s a simple copy of the input string.

[...]

PR #597, opened 2023-07-07
Editorial fixes from #566

ndw:
This PR addresses some minor comments from issue #566. There are more substantive comments but I think they warrant discussion first.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -24702,6 +24702,16 @@ declare function fn:some(
          otherwise <emph>hierarchical</emph> is
          <code>true</code> if <emph>string</emph> begins with <code>/</code> and <code>false</code> otherwise.</p>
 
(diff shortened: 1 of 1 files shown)


specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -24809,7 +24819,9 @@ declare function fn:some(
          (that is, <code>0xfffd</code>, the Unicode replacement character, in UTF-8).
          After replacing all of the percent-escaped characters, the character sequence is
          interpreted as UTF-8 to get the string. In other words <code>"A%XYC%Z%FO%9F%92%A9"</code> becomes
```
ChristianGruen, line 24821:
```suggestion
         interpreted as UTF-8 to get the string. In other words <code>"A%XYC%Z%F0%9F%92%A9"</code> becomes
```

…and there’s a superfluous quote some lines above: `<code>"ABC"</code>"`.

See #600 for a comment on the last line of the paragraph.

ndw: Nice catches!

PR #611, opened 2023-07-16
329: Keyword parameters: Error codes

ChristianGruen:


michaelhkay (changes requested):
I'm OK with some merging of error codes, but I would be happier if the description of the combined error code were a bit more expansive. Something like:

[ERR XPST0017] It is a [static error] if the [expanded QName] and number of arguments in a static function call do not match the name and [arity range] of a [function definition] in the [static context], or if an argument keyword in the function call does not match a parameter name in that function definition, or if two arguments in the function call bind to the same parameter in the function definition.

ChristianGruen:
Revised as suggested

ndw:

PR #614, opened 2023-07-17
123: fn:duplicate-values

ChristianGruen:
I decided to create a PR for the initial proposal of this function, as I came across at least two other use cases for it since the issue was created.

I believe that the `group by` clause is the best choice for more complex operations, such as advanced comparisons or creating histograms.


michaelhkay:
It's unfortunate that the function cannot, in fact, be implemented using a FLWOR group-by clause, as that does not support dynamic collations. Perhaps it would be better to provide an XSLT implementation:

```
<xsl:for-each-group select="$values" group-by="." collation="{$collation}">
  <xsl:sequence select="current-group()[2]"/>
</xsl:for-each>
```

(although this differs in that the result is more strictly defined; to be precisely equivalent, you would want to randomise the order of the input).

ChristianGruen:
> Perhaps it would be better to provide an XSLT implementation:

Added, thanks.

ndw:

PR #615, opened 2023-07-17
Xdm minor edits, chh. 3-5

Arithmeticus:


michaelhkay (commented):
You have made a number of changes to the hyphenation of the terms "typed value", "string value", "implementation defined". In all these cases the document currently hyphenates inconsistently, but I'm not sure your changes improve this much. Since these are all defined terms in the spec my own preference would be to hyphenate them consistently, regardless of their grammatical context, and link them more consistently to their definitions. I think there's a danger that de-hyphenating "typed value" increases the risk that the reader won't recognize it as a defined term.

Arithmeticus:
Most compounds that are naturally hyphenated when used in an adjectival construction are unhyphenated in nominal or predicate form. E.g., "zero-length string" vs. "a string of zero length", "implementation-defined solution" vs. "the property is implementation defined" (XDM 6.1.3), "mass-produced artefact" but "the artefact was mass produced", "kilowatt-hour rating" vs. "measured in kilowatt hours", "English(-)speaking", "high(-)profile", "book(-)length" etc. Better to think of the problem in the other direction. Many compounds that we naturally don't hyphenate when used absolutely must be hyphenated in an adjectival construction, lest there be any confusion whether in the phrase "a b x", "a" modifies "b" or "x" (otherwise we would naturally keep it open). My favorite, from a real-world example (a book on Chinese art I edited), was when my author used "baked potato peddler".

It its current form, the spec consistently renders the adjectival form hyphenated, "string-value X" and "typed-value X," but it is inconsistent when using it nominally or in predicate form, using both hyphenated and open constructions. Because most hyphenated adjectival compounds are converted to open form in a nominal or predicate grammatical construction (see examples previous paragraph), I pushed my edits in that direction.

If we go in the direction of converting all open compounds in nominal/predicate position to hyphenated, in imitation of the adjectival construction, numerous other edits need to be made, e.g., "content-type" vs "content type" and "media-type" vs "media type" (search other specs).

NB, in one paragraph in my XDM edits, I kept the repeated use of the hyphenated form but supplied the noun "property"; when I first read this paragraph, I paused and wondered what the difference was between a "string-value property" and a "string-value".

ndw:

Arithmeticus:
Although this PR/issue closed, I'll note that in the XPath spec (which I have turned to now), I see that "typed value" and "string value" are rendered 84 of 85 times without the hyphen. The one time it is hyphenated is in an adjectival construction; all 84 others are nominal constructions. The XPath spec defines these terms. In a future PR I'll provide an xref from XDM spec to XPath spec.

PR #619, opened 2023-07-18
XDM ch. 6 minor edits

Arithmeticus:


ChristianGruen (approved):

michaelhkay:
My personal preference it to write "types of item" rather than "types of items", but I have no objection: no doubt both can be justified by appeal to authority.

ChristianGruen (approved):

Issue #624, opened 2023-07-19
XPath function definition clarification


Arithmeticus:
In the XPath specs, 2.2, [function definitions](https://qt4cg.org/specifications/xquery-40/xpath-40-diff.html#dt-function-definition), the reader is informed that every (statically known) function definition takes one of three mutually exclusive categories: application, system, or external.

> [Definition: Application functions are function definitions written in a host language such as XQuery or XSLT whose semantics are defined in this family of specifications. Their behavior (including the rules determining the static and dynamic context) follows the rules for such functions in the relevant host language specification.]

The first sentence appears to point to user-written functions in the host languages, and the second sentence appears to point to functions defined by the host language specifications. I assume that this category is meant to include both, and I propose the language be tightened up to make that clear. If my assumption is incorrect, then some other type of revision is needed.

Later on, when the term “built-in function” is introduced, it is not as clearly stated as it should be how that term maps onto the three-way division. It seems that “built-in function” encompasses all system functions and only those application functions that are defined by the specifications, and not user-written functions. Whatever the case, and whatever the best path of revision, this paragraph would be most effective if moved up with the tripartite category discussion.

The term "external function" is introduced here in the static context, with language that is highly suggestive of a definition. But the definition proper is reserved for the dynamic context, and in shorter, different prose that I think loses some of the considerations provided in the static context. I propose that the two passages be consolidated and located in the static context, where the two other types of functions are defined, with an xref in the dynamic context to the meaning of external function.

Down in the dynamic context: 

> The dynamically known function definitions may include external functions.

This sentence is a bit puzzling, because of course they are allowed to include all three types of functions -- none are forbidden, since, after all, the dynamically known functions are a superset of statically known ones. Perhaps the point is to drawn the reader's attention to those functions that are known dynamically but not statically? Perhaps this revision?: "Many of the function definitions known dynamically but not statically will be external functions, but they may include user-written application functions written in a host language, known only dynamically, e.g., through `fn:transform`."

Cumulatively the above points could go beyond mere minor touch-up, so comments are welcome before I attempt any edits.

michaelhkay:
I think I made good progress in tidying this up from its state in 3.1 (got rid of "functions with an implementation-defined implementation"...) but I agree there is still room for further improvement.

For example, it should probably be stated that so-called "constructor functions" for atomic types (both built-in types and user-defined types) are classified as "system functions".

[...]

PR #625, opened 2023-07-20
XPath minor edits, chh. 1-2

Arithmeticus:
All minor edits. Some edits are motivated by an attempt to address broken parallelism or some other form of localized inconsistency. Does not include questions about function descriptions, raised at #624.


Arithmeticus:
From my notes:
Abstract: removed “news” relevant to 3.0 not 4.0
Introduction: “host language” has just been defined, so it can be used instead of its definition; “closely” overly used; broken grammatical parallelism

ndw (approved):

Issue #556, opened 2023-06-16
Serialization phase 5 note unclear


Arithmeticus:
[Section 4](https://qt4cg.org/specifications/xslt-xquery-serialization-40/Overview-diff.html#serphases) of the Serialization spects ends with this note:

> Serialization is only defined in terms of encoding the result as a stream of octets. However, a [serializer](https://qt4cg.org/specifications/xslt-xquery-serialization-40/Overview-diff.html#serializer) MAY provide an option that allows the encoding phase to be skipped, so that the result of serialization is a stream of Unicode characters.

What is a stream of Unicode characters? AFAIK, Unicode characters cannot be streamed _an sich_ but require some encoding. And how does would such a stream not consist of bytes/octets of one sort or another? A stream of what exactly?

michaelhkay:
I think "stream" here is synonymous with "sequence", so a "stream of Unicode characters" is actually synonymous with "string". But perhaps that's not quite right here, because we think of a string as something in memory, whearas this might be for example a UNIX pipe. The key thing here is that it's very common to send the serialized result to a destination that accepts characters rather than octets.

Serializing to a string is in fact very common.

Arithmeticus:
Another challenge is that the term "stream" is introduced only with step 5, and presumes of the reader awareness that steps 1 through 4 are streams (but is not otherwise stipulated). Further, "character" is ambiguous, meaning one thing as a PL data type and another as a Unicode concept. 

These are two separate issues. Here's an attempt to tackle the second, easier one:

> ...so that the Unicode resulting from serialization is not a stream of octets.

But it seems that streaming is still not clear. Is a processor forced to always provide the result as a stream? Or to handle each of the four steps as a stream?

ChristianGruen:
I think there’s no necessity to provide streaming in an implementation, and I think it would make sense to replace all occurrences of the term “stream” with “sequence”.

ChristianGruen:
#626 was merged; this issue can be closed

PR #626, opened 2023-07-20
Adjusted serialization step 5 note

Arithmeticus:
This brief edit addresses issue #556, which I've chosen to handle apophatically.


ndw (approved):

michaelhkay (changes requested):
I don't feel the phrase "at the result of serialization is a Unicode character sequence that is not a stream of octets" adds any clarity. I think the real point is that the result of serialization may be sent to a destination that requires Unicode characters to be represented in a particular way (for example, a Java StringBuffer), and when that is the case, the serializer will represent the characters in whatever way is required, which involves skipping the encoding phase,

ChristianGruen:
If we want to get rid of “stream”, maybe we can simply say “sequence of Unicode characters”.

Arithmeticus:
Thanks for the comments. See new commit.

PR #629, opened 2023-07-20
574: fn:log: Trace and discard results

ChristianGruen:
* New function `fn:log`
* Rules of `fn:trace` revised

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -628,15 +628,19 @@
          <p>In addition, the values of <code>$input</code>, typically serialized and converted
             to an <code>xs:string</code>, and <code>$label</code> (if supplied
             <phrase diff="add" at="2022-12-19">and non-empty</phrase>) <rfc2119>may</rfc2119>
-            be output to an <termref def="implementation-defined"/> destination.
-            The format of the trace output and its order are
-            <termref def="implementation-dependent"/>.</p>
+            be output to an <termref def="implementation-defined"/> destination.</p>
+         <p>The serialization of the trace output <rfc2119>must not</rfc2119> raise an error.
+            This can e.g. be achieved by using a serialization method that can handle arbitrary
+            input, such as the <xspecref spec="SER31" ref="adaptive-output"/>.</p>
+         <p>The format of the trace output and its order are
+            <termref def="implementation-dependent"/>. Therefore, the order in which the output
+            appears is not predictable. This also means that if dynamic errors occur
+            (whether or not they are caught using try/catch), it may be unpredictable whether
+            any output is logged before the error occurs.</p>
       </fos:rules>
       <fos:notes>
-         <p>Sometimes there is a need to output trace information unrelated to a specific value.
-            In such cases it can be useful to set <code>$input</code> to an empty string or an empty
-            sequence, and to compute the value of the <code>$label</code> argument: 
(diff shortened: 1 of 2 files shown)


michaelhkay (commented):
I suggest adding some notes (taken from xsl:message, may need adapting). 

(1) Any [dynamic error] that occurs while evaluating the argument expression , and any [serialization error] that occurs while processing the result, should not cause the transformation to fail; at worst, it means that no message is output, or that the only message that is output is one that relates to the error that occurred. [Is that achievable? I've put it as a "should"]

(2) Because the order of execution of expressions is implementation-defined, the order in which log messages appear is not predictable. This also means that if dynamic errors occur (whether or not they are caught using try/catch) it may be unpredictable whether any output is logged before the error occurs.

ChristianGruen:
> (1) Any [dynamic error] that occurs while evaluating the argument expression , and any [serialization error] that occurs while processing the result, should not cause the transformation to fail; at worst, it means that no message is output, or that the only message that is output is one that relates to the error that occurred. [Is that achievable? I've put it as a "should"]

I think it’s too invasive to enforce a general try/catch behavior for the function. Catching errors can be convenient for debugging, but it could even be disruptive in productive use cases (such as for writing logs). I can imagine many cases in which the rewriting/optimization of queries can lead to unexpected behavior, both for the user and across different implementations. For example, one might expect the following expressions to fail…

```xquery
(1 + "X") ! log(.),
let $a := 1 + "X" return log($a)
```

…but a processor would be allowed to simplify both of them to `log(1 + "X")`.

Next, I believe that developers who have worked with other languages may be used to the fact that the input of log functions must be valid  (e.g., I think of Java – `System.out.println` – JavaScript – `console.log` – or Python – `print(undefined-name)`.

When it comes to serialization, I fully agree. I’ll add a note for both `fn:trace` and `fn:log` to use a serialization method that can handle arbitrary output without raising errors, e.g. the `adaptive` serialization method.

> (2) Because the order of execution of expressions is implementation-defined, the order in which log messages appear is not predictable. This also means that if dynamic errors occur (whether or not they are caught using try/catch) it may be unpredictable whether any output is logged before the error occurs.

+1

ChristianGruen:
GitHub fails to adopt my latest commit (https://github.com/ChristianGruen/qtspecs/commit/a45a6f11f912bc32116ddeb45e61902f259b844c).

Maybe I’m just impatient. If the problem persists, I’ll create a new PR.

ChristianGruen:
The pipeline was successfully triggered by yet another minor commit.

michaelhkay:
The technical substance is fine.

Editorial nits:

>The serialization of the log output must not raise an error. This can e.g. be achieved by using a serialization method

I suggest "This can be achieved (for example) by using...". Abbreviations such as `e.g.` are best avoided.

>that can handle arbitrary input, such as the [Section 10 Adaptive Output Method]

`specref` and `xspecref` markup should be used only when referring to the section, not to the things described by the section. If there's no suitable xtermref, write "such as the adaptive output method (see `<specref...`)."

ChristianGruen:
> I suggest "This can be achieved (for example) by using...". Abbreviations such as `e.g.` are best avoided.

Thanks: I wondered about this. We should definitely fix this in the remaining spec, as it’s used a lot.

> `specref` and `xspecref` markup should be used only when referring to the section, not to the things described by the section. If there's no suitable xtermref, write "such as the adaptive output method (see `<specref...`)."

✓ resolved

ndw:

michaelhkay:
Subsequently renamed fn:message

Issue #600, opened 2023-07-10
fn:decode-from-uri: counterpart of fn-encode-to-uri


ChristianGruen:
Adopted from https://github.com/qt4cg/qtspecs/issues/566#issuecomment-1607397586:

The initial suggestion in #72 was to provide a function for decoding a URI. Maybe we should still think about adding a `fn:decode-uri` or `fn:decode-from-uri` function, in which we could tackle the open issues that need to be solved for `fn:parse-uri`. The current URI decoding rules could then be replaced with a reference to this new function.

ndw:
It would make the function library more symmetric. I assume we're thinking that `decode-from-uri` would do the decoding of URIs to IRIs.

1. If the input does not contain any '%' characters, return it unchanged.
2. Replace all %nn with the corresponding bytes and then UTF-8 decode the resulting string to get a sequence of Unicode characters. Return that.

I suggest that if the string contains invalidly encoded characters (A `%` followed by less than two characters or followed by characters not in the range 0-9, a-f, A-F; or if the resulting sequence of bytes isn't a valid UTF-8 encoding) the function inserts U+FFFD characters in the relevant locations but does not throw an error.

ChristianGruen:
> I assume we're thinking that `decode-from-uri` would do the decoding of URIs to IRIs.

I must admit I’ve never purposely worked with IRIs. Maybe we could have an additional `fn:uri-to-iri` function for that, but maybe that wouldn’t make any sense?

> If the input does not contain any '%' characters, return it unchanged.

Similar as for, `fn:parse-uri`, I would expect `+` to be converted to a space.

> Replace all %nn with the corresponding bytes […]

I’ve added another comment on scanning UTF-8 at about the same time; I think variant 2 would be the most flexible one:

* For `%F0%9F%92%41`, we get `�A`, as only the 4th byte does not match the UTF-8 rules.
* For `%F0%F0%9F%92%A1`, we get `�💡`, as the 2nd byte does not match the rules, but can successfully be scanned as first byte of the upcoming UTF-8 sequence.

> […] but does not throw an error.

+1

PR #631, opened 2023-07-21
600: fn:decode-from-uri

ChristianGruen:
I did my best to define rules for a counterpart of the `fn:encode-for-uri` function, including various edge cases.

I’m convinced that the function has been requested often enough to justify its inclusion in the spec. I’m also aware that users may have different expectations regarding the details of the conversion rules. On the other hand, this discussion can be observed for other languages as well, and that’s mostly due to the… heterogeneous history of URIs, not the actual implementations. For example, `URLDecode.decode` in Java converts the plus character to a space, and JavaScript’s `decodeURI` adopts it unchanged. I decided to convert it as well, as `fn:encode-for-uri` encodes the plus sign to `%2B`.

@ndw My rules have largely been inspired by your decoding rules for `fn:parse-uri`. I hope these rules can be dropped and replaced with a reference to this new function (analogous to `fn:build-uri`, which references `fn:encode-for-uri`).

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -4353,6 +4353,7 @@ return normalize-unicode(concat($v1, $v2))</eg>
          </fos:example>
       </fos:examples>
    </fos:function>
+
    <fos:function name="encode-for-uri" prefix="fn">
       <fos:signatures>
          <fos:proto name="encode-for-uri" return-type="xs:string">
@@ -4399,9 +4400,9 @@ return normalize-unicode(concat($v1, $v2))</eg>
          <fos:example>
             <fos:test>
                <fos:expression><eg>encode-for-uri(
-  "http://www.example.com/00/Weather/CA/Los%20Angeles#ocean"
+  "http://example.com/00/Weather/CA/Los%20Angeles#ocean"
 )</eg></fos:expression>
-               <fos:result>"http%3A%2F%2Fwww.example.com%2F00%2FWeather%2FCA%2FLos%2520Angeles%23ocean"</fos:result>
+               <fos:result>"http%3A%2F%2Fexample.com%2F00%2FWeather%2FCA%2FLos%2520Angeles%23ocean"</fos:result>
                <fos:postamble>This is probably not what the user intended because all of the
                   delimiters have been encoded.</fos:postamble>
             </fos:test>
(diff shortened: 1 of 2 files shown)


ChristianGruen:
I have clarified the UTF-8 conversion rules; the revised version should be available in a few minutes.

It may be better to align the rules for parsing the percent-encoded substrings to the subsequent parsing of the UTF-8 octet sequence. The first invalid character in a substring would then be parsed as a new character (see also https://github.com/qt4cg/qtspecs/issues/600#issuecomment-1628526326):

* `%X` would become `�X` (instead of `�`)
* `%0X` would become `�X` (instead of `�`)
* `%XX` would become `�XX` (instead of `�`)
* `%%41` would become `�A` (instead of `�1`)

Any opinions?

We could also raise an error for invalid input (I have chosen the fallback variant, as it was suggested by Norm for `fn:parse-uri`).

Next, we may need to revise the rules if we lift the character set restriction for strings (see #414).

michaelhkay:
A few comments, mainly on presentation.

>This function returns the original representation of a URI-encoded string.

"original"? "URI-encoded"? (encode-for-uri uses the terms "URI escaping" and "percent-encoded". I wonder which is correct?)

>occurrences of %[a-fA-F0-9][a-fA-F0-9] are replaced with octets for the two-digit hexadecimal numbers that follow the percent sign. 

Perhaps "any substring that matches the regular expression ... is replaced with an octet corresponding to the two-digit hexadecimal number that follows the percent sign. "

>All other characters are added as UTF-8 octet sequences. 

Perhaps "Characters that are not part of such a substring are replaced with the octets of their UTF-8 encoding."

Presumably there is one 0xFFFD character for the substring, not one for each character in the substring?

>For example, "A%42+C" becomes x41x42x20x43, and "💡" becomes xF0x9Fx92xA1.

Is this the best way of representing an octet sequence? It's confusingly similar to the way we represent a Unicode codepoint.

>If % is followed by up to two characters that are no hexadecimal digits, 

"No" => "not"? Does "up to two" include zero? What happens to a "%" at or near the end of the string? What if "%" is followed by three characters that are not hexadecimal digits? This all seems very unclear. I suspect the condition is "If a "%" character is present that is not followed by two hexadecimal digits", but I'm still unclear as to which characters are replaced by the 0xFFFD.

>these characters are replaced by octets representing the Unicode replacement character (0xFFFD): xEF, xBF, and xBD.

Might as well use the same notation for an octet sequence: "replaced by the octet sequence xEFxBFxBD, that is, the UTF-8 encoding of the Unicode replacement character 0xFFFD. And we need to be clear whether each character is individually replaced or whether the substring is replaced.

> The incomplete or invalid percent-encoded strings "%", "%X", and "%AX" are all replaced with these octets. For the string "%1X!", the octets xEFxBFxBDx21 are returned.

Is this supposed to be an example of the previous rule, or an additional rule?

>Next, the resulting octets are interpreted as UTF-8: x41x42x20x43 becomes "AB C", and xF0x9Fx92xA1 becomes "💡".

Signal more clearly that this is an example (not, say, an exception to the rule).

>If an invalid UTF-8 character subsequence is encountered, the octets that have successfully been parsed are replaced with a Unicode replacement character:

What exactly is an "invalid UTF-8 character subsequence", and how can it have found its way into the octet sequence we have just created?

>The octet xF0 is converted to "�".

In my browser this looks like a typo, rather than an intentional rendition of the SUB character. Need to find some other way of depicting this. Also I can't reconcile this with the suggestion above that F0 is the first octet in the UTF-8 representation of a light-bulb character.

>The octets xF0x9Fx92x41 are converted to "�A": 

We seem to be back in specification-by-example mode. Is this a rule, or an example of a rule that has already been stated? Or is it an example from which we are supposed to infer a rule?

>The bit pattern of the first octet indicates that the UTF-8 character comprises four octets. As the fourth octet is invalid, a Unicode replacement character is added for the first three octets, and the fourth (invalid) octet is parsed as a new character.

I'm afraid (a) I'm still confused as to how we managed to construct invalid UTF-8 in the first place, and (b) why we need to define our own UTF-8 decoding rules.

>Similarly, a character that is no valid XML character

Perhaps "a UTF-8 octet sequence that represents a codepoint that is not a valid XML character".

> is replaced with a Unicode replacement character: x00 becomes "�".

Again, signal the example.

ChristianGruen:
> A few comments, mainly on presentation.

Thanks.

> "original"? "URI-encoded"? (encode-for-uri uses the terms "URI escaping" and "percent-encoded". I wonder which is correct?)

I switched to “URI-escaped” (percent-encoded might suggest we won’t decode plus signs).

> Presumably there is one 0xFFFD character for the substring, not one for each character in the substring?

I’m not sure which rule/text this comment applies to?

> Is this the best way of representing an octet sequence? It's confusingly similar to the way we represent a Unicode codepoint.

I’ve replaced the representation with lists of octets: `<code>xF0<code>, <code>x9F</code>`, …

> "No" => "not"? Does "up to two" include zero? What happens to a "%" at or near the end of the string? What if "%" is followed by three characters that are not hexadecimal digits? This all seems very unclear. I suspect the condition is "If a "%" character is present that is not followed by two hexadecimal digits", but I'm still unclear as to which characters are replaced by the 0xFFFD.

This paragraph was basically pasted from the existing rules for [fn:parse-uri](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-parse-uri). My assumption was that it was already accepted and can thus be adopted (and it seems clear enough to me not to change it, but I may have got it wrong). I’ll be happy to adopt other sugestions.

Regarding the actual rules: Maybe you’ve seen my second comment in this PR, I think it relates to your question.

> > The incomplete or invalid percent-encoded strings "%", "%X", and "%AX" are all replaced with these octets. For the string "%1X!", the octets xEFxBFxBDx21 are returned.
> 
> Is this supposed to be an example of the previous rule, or an additional rule?

It’s an example; I’ve made this explicit now. I’ve found various other rules in the spec in which there’no explicit hint (for example:  `So [Y9,999,*] will output the year as 2,008.`), but it’s good to learn more about the expectations.

> What exactly is an "invalid UTF-8 character subsequence"

Changed to “invalid UTF-8 octet sequence ”.

> > The octet xF0 is converted to "�".
> In my browser this looks like a typo, rather than an intentional rendition of the SUB character. Need to find some other way of depicting this.

I hoped it would be obvious because the text repeatedly elaborates on the replacement character; but I’ll be happy to change the representation.

> Also I can't reconcile this with the suggestion above that F0 is the first octet in the UTF-8 representation of a light-bulb character.

The presented UTF-8 representation of the light bulb should be correct (see e.g. https://www.compart.com/de/unicode/U+1F4A1), but maybe I got you wrong?

> I'm afraid (a) I'm still confused as to how we managed to construct invalid UTF-8 in the first place, and (b) why we need to define our own UTF-8 decoding rules.

I don’t think we have to, but I’m not aware where we can find documentation on the proper decoding of UTF-8 sequences, and the rules of `fn:parse-uri` didn’t cover any details. I’m not even sure if different programming languages follow the same conventions. If there’s a unified approach, maybe someone can offer a good reference?

ndw:
The CG agreed to accept this PR at [meeting 045](https://qt4cg.org/meeting/minutes/2023/09-12.html)

PR #632, opened 2023-07-21
SENR0001: Error description updated

ChristianGruen:
Observed by @line-o:
https://xmlcom.slack.com/archives/C01GVC3JLHE/p1689946671105499

### DIFF
--- a/specifications/xslt-xquery-serialization-40/src/errors.xml
+++ b/specifications/xslt-xquery-serialization-40/src/errors.xml
@@ -12,8 +12,8 @@ namespace prefix binding in this document is not normative.</p>
 <error code="0001" class="NR" type="serialization" spec="SE">
 <p>It is an error if an item in <emph>S<sub>6</sub></emph> in <termref
 def="sequence-normalization">sequence normalization</termref> is an
-attribute node or a namespace node.</p></error>
-
+attribute node, a namespace node or a
+<termref def="dt-function-item">function</termref>.</p></error>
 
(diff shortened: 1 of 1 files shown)


line-o:
I would propose to reflect the change in https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-serialize under error conditions as well. It currently states:

> The serialization process will raise an error if $input is an attribute or namespace node.

And could be modified to 

> The serialization process will raise an error `SENR0001` if $input is an attribute, namespace node, function or map.

ChristianGruen:
I’ve dropped this rule as it is only valid for a subset of the serialization methods.
In addition, I’ve updated the type of the `$options` parameter.

line-o:
I agree on the options but the second paragraph you dropped is about $input

ChristianGruen:
The rule for `$input` was dropped as it is only valid for a subset of the serialization methods.

What do others think: Should we keep this specific serialization error in the XQFO spec and give more details on it?

ChristianGruen:
To be revised (https://github.com/qt4cg/qtspecs/commit/9a785062ac172f4fef41989f17f915339170e337#commitcomment-122390463)

line-o:
As an implementor I would like to see which _common_ errors are expected to be raised on typical conditions. I would therefore like to see the mention of serialization input error be generalised as such:

> The serialization process will raise an error SENR0001 if `$input` is of a type that cannot be serialised with the given method option.

ndw:

PR #633, opened 2023-07-22
Edits ch. 4.1 through 4.15

Arithmeticus:
Note, this batch of edits includes a shift from "built-in" to "system" when describing functions, per edits made in XDM.


Arithmeticus:
By the way, is there a good reason why xquery-40\src\expressions.xml has three root elements (and so is not well-formed)?

ChristianGruen (approved):

ChristianGruen (approved):

PR #667, opened 2023-08-18
XPath minor edits, 4.16 through end

Arithmeticus:
Various small edits. I added two simple examples to help readers quickly understand the difference between `=>` and `=!>` before going to more complex examples.


Arithmeticus:
@ndw yes, edits as intended, because it is repeated by the better-placed 3rd line down, 19105; thanks for checking. cc @ChristianGruen

PR #673, opened 2023-09-05
HTML namespace changes

rhdunn:
This PR applies my action items for updating the HTML XDM mapping around namespaces and local names.

Note: this currently makes `dm:namespace-nodes` return an empty sequence. I'm not currently sure what the best approach is here.

### DIFF
--- a/specifications/xpath-functions-40/src/xpath-functions.xml
+++ b/specifications/xpath-functions-40/src/xpath-functions.xml
@@ -6366,6 +6366,21 @@ correctly in all browsers, depending on the system configuration.</emph></p>-->
                         <item>
                            <p>The <emph>local name</emph> is the value of the
                               <code>Element.localName</code> property;</p>
+                           <note>
+                              <p>If the local name contains a character that is not a valid XML
+                                 <code>NameStartChar</code> or <code>NameChar</code>, then an
+                                 <termref def="implementation-defined">implementation-defined</termref>
+                                 replacement character is used. The result must be a valid
+                                 <code>NCName</code>.</p>
+                              <p>For example, in <bibref ref="dom-ls"/> section 13.2.9
+                                 <emph>Coercing an HTML DOM into an infoset</emph>, it uses a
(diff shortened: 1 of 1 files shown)


michaelhkay:
>The HTML DOM <code>CDATASection</code> interface is an instance of HTML DOM
                         <code>Text</code>, so CDATA sections also map to <xspecref spec="DM40" ref="TextNode"/>

Should also mention that XDM does not allow adjacent text nodes, therefore after converting CDATA nodes to text nodes, adjacent text nodes need to be combined.

michaelhkay:
>HTML DOM <code>Attr</code> instances in an HTML DOM <code>HTMLDocument</code> that are
                         namespace declarations are not supported; they are filtered out by this mapping. 

A couple of comments here. Firstly the phrase "are not supported" is highly ambiguous. Better to say simply that the nodes are filtered out. Secondly, does this really belong in a Note? It feels normative to me.

michaelhkay:
>If the local name contains a character that is not a valid XML
                                  <code>NameStartChar</code> or <code>NameChar</code>, then an
                                  <termref def="implementation-defined">implementation-defined</termref>
                                  replacement character is used.

Again, this feels normative: it doesn't feel like a note. Also, the example doesn't us a "replacement character" it uses a replacement string. Should we prescribe the escaping mechanism? Should it be reversible (which means escaping some names that are valid, e.g. those containing underscores)?

michaelhkay:
>Note: this currently makes dm:namespace-nodes return an empty sequence. I'm not currently sure what the best approach is here.

I think the best approach is to apply the namespace fixup rules: that is, we add the minimum number of namespace nodes to satisfy the consistency rules, ensuring that all prefixes actually used in  element and attribute names are defined in an in-scope namespace binding.

Mike

ndw:
We agreed to merge this PR at [meeting 046](https://qt4cg.org/meeting/minutes/2023/09-12.html#h-4CFD14DD-17A1-4E18-9D2B-98D8EFDA9813)

PR #683, opened 2023-09-06
XQFO context/focus in/dependent functions clarification note

Arithmeticus:
Added note clarifying the relationship between context and focus, for the purpose of illustrating the relationships between focus/context in/dependent functions. Wrapped with companion `<note>`s in a parent `<notes>`.


ChristianGruen (approved):
👍

PR #694, opened 2023-09-12
XQFO minor edits, with new examples and notes, 2 through 4.6

Arithmeticus:
Minor edits here are motivated by clarity or localized consistency. In one case, the change of `0` to `0` to `0` to `9` appears to address an important typo.

I have introduced select examples, to illustrate points in the corresponding rules.

The notes I have introduced need some context. Option 9 for the primary format token is somewhat vague, and the attached note of clarification stokes the imagination. I have trimmed that note for clarity (without substantive changes) but introduced a set of notes later, to help caution developers on unexpected behaviors with option 9, and secondarily to caution processor implementers on the challenges inherent in supporting this option. If  I had my druthers, I would advocate deprecating option 9. It is--to use a technical term--squishy.

Spec editors: I am happy to pull back on any of this.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -589,7 +589,7 @@
                   chooses.</p>
                <p>If no value is supplied for the <code>$value</code>
                   argument <phrase diff="add" at="2022-12-19">or if the value supplied
-                     is an empty sequence,</phrase>, then the
+                     is an empty sequence</phrase>, then the
                   effective value of the error object is <termref
                      def="implementation-dependent">implementation-dependent</termref>.</p>
             </item>
@@ -1187,6 +1187,12 @@
                <fos:result>10.5</fos:result>
             </fos:test>
          </fos:example>
+         <fos:example>
+            <fos:test>
+               <fos:expression>abs(-math:log(0))</fos:expression>
+               <fos:result>INF</fos:result>
+            </fos:test>
+         </fos:example>
       </fos:examples>
    </fos:function>
    <fos:function name="ceiling" prefix="fn">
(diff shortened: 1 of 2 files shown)


michaelhkay:
You added some examples with `<fos:result>INF</fos:result>` (or -INF) - the value of the `fos:result` element must be a valid XPath expression so that the example is executable. Needs to be `xs:double('INF')`

I feel that the phrase "If the argument is infinity, positive or negative, " is ambiguous. It's not clear whether the adjectives "positive" and "negative" apply to "the argument" or to "infinity".

>x3C3;). Because final capital sigma is
                  not attested, <code>format-integer(18, "Α;a")</code> may return #x3A2, a reserved
                  character currently assigned to no glyph. 

Firstly, I'm not sure what "attested" means in this context. Secondly, you seem to be speculating here about the quality of implementations, which doesn't seem appropriate in the spec. And it's not clear whether you are using "may" to mean "might" or "is permitted to".

Arithmeticus:
Agreed on all points, which I hope the commit I've just pushed addresses.

ChristianGruen (approved):

michaelhkay:
I don't understand your comments about "option 9".

The history here is that we had an inconsistency between different functions (and XSLT instructions): some used "001" as the picture for a three-digit number, others used "000". So we decided to allow any sequence of three digits, including for example "000", "001", "999", and "123". Historically, the pattern "999" is used to denote a three digit number in COBOL, and it's often used in informal data format specifications, for example the format of a UK vehicle registration is AA99AAA.

michaelhkay (approved):

Arithmeticus:
By "option 9" I was referring to the 9th enumerated way to make a primary format token in the specs at 4.6.1.

michaelhkay:
Ah, I see. Yes, that's always been very open-ended. It dates right back to xsl:number in XSLT 1.0.

ndw:
The CG agreed to accept this PR at [meeting 046](https://qt4cg.org/meeting/minutes/2023/09-19.html).

PR #696, opened 2023-09-13
566: Rework query parameters on build-uri/parse-uri

ndw:
Completes action QT4CG-042-02.

1. Rework the query segments so that they're a simple map of key/value pairs.
2. Rename `query-segments` to `query-parameters`.


ChristianGruen:
I’ve prefixed the PR title with the issue number #566 although it doesn’t exactly match (the issue shouldn’t be closed yet if this PR is merged).

PR #702, opened 2023-09-14
701: fn:concat: Support for 0 or more arguments

### LINKED ISSUE #701: fn:concat: Support for 0 or more arguments
With #161, we plan to introduce support for variadic functions.

The scope of this issue is much smaller and can be seen as a preparatory one; it’s about allowing the first two arguments of the function optional. I’ll create a little PR for it.

ChristianGruen:
Closes #701

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -3664,9 +3664,7 @@ return if (starts-with($preprocessed-value, "-")) then -$abs else +$abs]]></eg>
    <fos:function name="concat" prefix="fn">
       <fos:signatures>
          <fos:proto name="concat" return-type="xs:string">
-            <fos:arg name="value1" type="xs:anyAtomicType?"/>
-            <fos:arg name="value2" type="xs:anyAtomicType?"/>
-            <fos:arg name="..." type="xs:anyAtomicType?"/>
+            <fos:arg name="..." type="xs:anyAtomicType*"/>
          </fos:proto>
       </fos:signatures>
       <fos:opermap operator="||" types="xs:anyAtomicType xs:anyAtomicType"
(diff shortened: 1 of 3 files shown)


michaelhkay (commented):
For a bit more precision, I'd suggest defining it as:

This function accepts zero or more arguments, each declared with the required type <code>xs:anyAtomicType*</code>. The value supplied for each argument is reduced to a single string by applying the fn:string-join#1 function; the resulting sequence of strings (one for each supplied argument) is then reduced to a single string by a further application of fn:string-join#1.

michaelhkay:
Hmmm... I think that accepting sequences breaks 1.0 compatibility mode. In XPath 1.0, `concat(//h1, //h2)` concatenates the value of the first `h1` element and the first `h2` element. XPath 4.0 in 1.0 compatible mode would no longer reproduce that behaviour. Perhaps we should drop that part of the proposal.

ChristianGruen:
> the resulting sequence of strings (one for each supplied argument) is then reduced to a single string by a further application of fn:string-join#1.

I wonder if it would make a difference to invoke string-join just once for all strings, or repeatedly for the strings of each argument?

> We could make the actual semantics of the function dependent on whether compatibility mode is set in the static context.

Would be fine by me. I might be overwhelmed finding the right wording, though. Would you be ready to be revise the PR?

michaelhkay:
>I wonder if it would make a difference to invoke string-join just once for all strings

It wouldn't make a difference, it would just be a little bit more difficult to define (I think).

michaelhkay:
For 1.0 compatibility, try:

This function accepts zero or more arguments, each declared with the required type `xs:anyAtomicType*`. The value supplied for each argument $A is reduced to a single string as follows:

* if XPath 1.0 compatibility mode is set to true in the static context of the function call, then the result of `xs:string($A[1])`

* Otherwise, the result of fn:stirng-join($A)

The resulting sequence of strings (one for each supplied argument) is then reduced to a single string by applying  the function fn:string-join#1 to the sequence.

ChristianGruen:
Revised as suggested, thanks.

PR #710, opened 2023-09-17
36: fn:function-annotations

### LINKED ISSUE #36: fn:function-annotations (Allow support for user-defined annotations)
### Requirements/Use Cases
1. It should be possible for a library or application written in XQuery to define, access, and use custom annotations without relying on vendor extensions.
2. It should be possible for a processor or editor/IDE to check and verify user-defined annotation usage (can an annotation be used multiple times; what argument values/types are valid) to provide better validation for custom annotations.
3. It should be possible to access the details of the annotations on a function (name, argument values) so tools like xqDoc can read and format annotations without using vendor extensions.

### Annotation Declarations
1. An annotation declaration is a function declared using the `%annotation` annotation.
2. Annotation parameters must be a SequenceType that has an ItemType which is either a) a subtype of `union(xs:string, xs:double, xs:float, xs:decimal)*`, or b) `item()` (to denote any permitted literal value). __Note:__ This is because annotation values are restricted to literal values. __Note:__ It is therefore possible to use an EnumerationType to define a set of allowed string values (e.g. "yes" and "no").
3. If a function annotation name matches an annotation declaration in the statically-known functions using the function name resolver, it is verified using the same rules for static function calls; if there are no matching annotation declarations, then no error is issued, although an implementation may choose to issue a warning.

### Annotation Tests
1. A new AnnotationTest (`annotation(...)`) SequenceType is provided with the same structure and semantics as a FunctionTest.
2. AnnotationTest parameters must be a SequenceType that has an ItemType which is either a) a subtype of `union(xs:string, xs:double, xs:float, xs:decimal)*`, or b) `item()` (to denote any permitted literal value). __Note:__ This is because annotation values are restricted to literal values. __Note:__ It is therefore possible to use an EnumerationType to define a set of allowed string values (e.g. "yes" and "no").

Returns all the annotations on the function.

The `annotation` signature will match the signature of the annotation, so `%a(1, 2)` will have the `annotation(xs:integer, xs:integer)` annotation type.

If a function has multiple annotations with the same signature (e.g. `%values(1) %count(2)`) then two annotations will be returned with the same signature, referring to the different annotations.

Returns all the annotations on the function with the name `$name`.

The `annotation` signature will match the signature of the annotation, so `%a(1, 2)` will have the `annotation(xs:integer, xs:integer)` annotation type.

If a function has multiple annotations with the same name and signature (e.g. `%values(1) %values(2)`) then two annotations will be returned with the same signature, referring to the different annotations.

Returns the name of the annotation.

Returns the arguments passed to the associated annotation.

ChristianGruen:
@michaelhkay Again, some hints might need to be added for XPath (?).

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -17400,6 +17400,58 @@ return function-arity($initial)</eg></fos:expression>
          </fos:example>
       </fos:examples>
    </fos:function>
+
+   <fos:function name="function-annotations" prefix="fn">
+      <fos:signatures>
+         <fos:proto name="function-annotations" return-type="map(xs:QName, xs:anyAtomicType*)">
+            <fos:arg name="function" type="function(*)"/>
+         </fos:proto>
+      </fos:signatures>
+      <fos:properties>
+         <fos:property>deterministic</fos:property>
+         <fos:property>context-independent</fos:property>
+         <fos:property>focus-independent</fos:property>
+      </fos:properties>
+      <fos:summary>
+         <p>Returns the annotations of the function identified by a function item.</p>
+      </fos:summary>
+      <fos:rules>
+         <p>The <code>fn:function-annotations</code> function returns the annotations of
+            the function identified by <code>$function</code> as a map.
+            For each annotation, a map entry is returned: The key is the name of the annotation.
+            The value is a sequence comprising the annotation values, or an empty sequence
+            if the annotation has no values.
+            If a function (for example, a built-in function) has no annotations,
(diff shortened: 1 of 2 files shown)


michaelhkay (commented):
For avoidance of doubt, we should say in XQuery 4.6.2.4 (Named Function References) that the function created by a named function reference has its annotations taken from the function definition. There are other places where we are not explicit about the annotations of a function item, for example with partial function application. We should add a note that in XPath and XSLT, it is not possible to define function annotations, so this function will always return an empty result. (However, we should consider giving user-defined functions in XSLT annotations based on their attributes, e.g. visibility and streamability).

Issue #517, opened 2023-05-21
FO: fn:multi-compose : Evaluate a chain of functions in order


dnovatchev:
**FO: fn:multi-compose : Evaluate a chain of functions**
=============================================================
As per Wikipedia: 

" In [mathematics](https://en.wikipedia.org/wiki/Mathematics), **function composition** is an operation  ∘  that takes two [functions](https://en.wikipedia.org/wiki/Function_(mathematics)) *f* and *g*, and produces a function 

*h* = *g*  ∘  *f* such that *h*(*x*) = *g*(*f*(*x*)).

In this operation, the function *g* is [applied](https://en.wikipedia.org/wiki/Function_application) to the result of applying the function *f* to *x*. That is, the functions 

*f* : *X* → *Y* and *g* : *Y* → *Z* 

are **composed** to yield a function that maps *x* in [domain](https://en.wikipedia.org/wiki/Domain_of_a_function) *X* to *g*(*f*(*x*)) in [codomain](https://en.wikipedia.org/wiki/Codomain) *Z*.

Intuitively, if *z* is a function of *y*, and *y* is a function of *x*, then *z* is a function of *x*.

The resulting *composite* function is denoted *g* ∘ *f* : *X* → *Z*, defined by:

 (*g* ∘ *f* )(*x*) = *g*(*f*(*x*)) for all *x* in *X*
"
______________________________

In this Proposal we generalize function composition to the case when a sequence of functions are composed together, so that the last one is applied on an argument \$x, then  the last-but-one is applied on the result of this application, and so on … until finally the first function in the sequence is applied on the result produced so far.

This is an effective way of chaining a sequence of functions together, and we don’t need to invent or use any special operators or syntax, but we just pass this sequence of functions as argument to **_fn:multi-compose_**.  
   
**fn:multi-compose := function($funs as function(\*)\*, $x)**
 
**Here is an XPath 3.0 implementation of fn:multi-compose**:  

```xq
let $apply := function($f, $x) {fn:apply($f, [$x])},
    $multi-compose := function($funs as function(*)*, $x)
                  {
                    fold-right($funs, $x, $apply)
                  },
    (: The functions $incr and $times are needed just to show this example :) 
    $incr := function($x) {op("+")(?, $x)},
    $times := function($y) {op("*")(?, $y)}                 
                  
 return
   $multi-compose(($times(5), $incr(1)), 2)
```
  
As wanted, the result of evaluating this is

15: (2 +1) \* 5

[...]

PR #733, opened 2023-09-30
517: fn:chain

dnovatchev:
I added fn:chain, which has been discussed in https://github.com/qt4cg/qtspecs/issues/517


ChristianGruen:
@dnovatchev Thanks, and congrats for finalizing the PR. You may need to store the function-catalog.xml file with Unix-style newlines (\n), otherwise the diff cannot be properly displayed:

https://github.com/qt4cg/qtspecs/pull/733/files#diff-7625c07ae8131ff65c3caa677b188ed2b9b66237312d11c05a2fa2838c6f5c67

dnovatchev:
Hi Christian,

> You may need to store the function-catalog.xml file with Unix-style
newlines (\n), otherwise the diff cannot be properly displayed

How to do this? Never needed to do this before.

Thanks,
Dimitre

dnovatchev:
The line endings are now Unix style.

ChristianGruen:
> How to do this? Never needed to do this before.

It depends on the text editor you use (many allow you to choose the line ending). Feel free to give us more information on your setup.

dnovatchev:
This is now fixed

ChristianGruen:
> This is now fixed

I don't think so: https://github.com/qt4cg/qtspecs/pull/733/files#diff-7625c07ae8131ff65c3caa677b188ed2b9b66237312d11c05a2fa2838c6f5c67

Maybe you haven't committed your changes yet?

ChristianGruen:
You can also change your git configuration to always send `\n` (text eol=lf):

https://docs.github.com/de/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

dnovatchev:
Have patience: "build-pr / build-and-deploy-pr (pull_request_target)  n progress — This check has started..."   😄

dnovatchev:
The latest changes must be there now.

ChristianGruen:
Same as before, Dimitre:

gh

michaelhkay:
I suggest using `$functions` rather than `$funSequence` as the argument keyword. We don't normally use camel-case except where XSD uses it.

I'm reading the rendered version of the new spec linked from the dashboard page. On this the "more formally" XPath code appears to be incomplete - it ends in a comma.

I'm not comfortable with logic reading `if(not($x instance of array(*))) then array{$x} else $x`. This seems to break substitutability. We've generally taken the view that if something works on items in general then it shouldn't treat arrays specially. Also there doesn't seem to be any equivalent in the "informal" exposition. If we had a general approach of treating arrays differently then this might make sense, but the function library should try and follow common conventions throughout.

Generally I think it would be simpler if we confined this to functions of arity one. Or perhaps we could devise a general mechanism, a function `fn:singularize` (for want of a better name) that takes a function of arity N as input and converts it to a function of arity 1 that takes an array as its single argument?

In Note 4, I suspect this should be a type error rather than a dynamic error. Rather than saying "It is a dynamic|type error", which is a phrase we often use for a normative statement, it might be worth using different phrasing to indicate that here we are only telling the reader something they could have worked out from the normative rules, perhaps "A consequence of the rules given earler is that a type error occurs if...".

There is markup available for declaring variables that are used in examples. It's necessary to use this markup to ensure that the examples are testable; at the moment the examples will generate test cases that fail with static errors due to undeclared variables. I can help with this if necessary.

dnovatchev:
Christian,

I have done everything on my side:

1. Updated the 2 files.
2. Committed and pushed to (my fork's master branch)
3. Initiated a PR from my fork (master) to qtspecs (master)
4. The associated github actions/jobs completed successfully

image

So, what could be the issue?

ChristianGruen:
> So, what could be the issue?

https://github.com/qt4cg/qtspecs/pull/733#issuecomment-1741686976

dnovatchev:
Maybe this????


image

ChristianGruen:
> Not sure how to avoid this from happening . Line endings have never been a problem for me before.

https://github.com/qt4cg/qtspecs/pull/733#issuecomment-1741821436

michaelhkay:
Labelling as abandoned - superseded by #734

PR #763, opened 2023-10-19
686: XQFO diagnostic function documentation

Arithmeticus:
I think what prompted my original ticket was the use of "output". That term is used scores of times in the specs, and it always implies the output of a function within the XPath expression. Although the term is qualified in the diagnostic functions with the adjective "trace" or "log", it's tempting to think of the function's actual output. Coming back a few weeks later, I think some gentle massaging (instead of a convoluted preamble) can help readers not make the requisite adjustment, reflected in the present PR.

I have offered two examples that I hope show the practical side of these functions. I might have missed something though, so please chime in.


specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -694,10 +694,24 @@
             <p>Consider a situation in which a user wants to investigate the actual value passed to
                a function. Assume that in a particular execution, <code>$v</code> is an
                   <code>xs:decimal</code> with value <code>124.84</code>.
-                  Writing <code>fn:trace($v, 'the value of $v is:')</code> will output the strings
+                  Writing <code>fn:trace($v, 'the value of $v is:')</code> will return <code>$v</code>
```
ChristianGruen, line 697: I think the sentence should be shortened to “will return &#x3c;code&#x3e;$v&#x3c;/code&#x3e;”. The appendage “to the XPath expression” should be skipped as it’s the (XPath, XQuery) expression that returns the value, not the other way round – and the expression in this case is the `fn:trace` function call.

Next, I noticed that the example in the [XQFO 3.1 specification](https://www.w3.org/TR/xpath-functions-31/#func-trace)…

> Writing `fn:trace($v, 'the value of $v is:')` will put the strings `"124.84"` and `"the value of $v is:"` in the trace data set in implementation dependent order.

…has already been flawed, or at least it didn’t match the way we create trace output, which is to output a single string for the serialized value and the optional label. As my experience is that each implementation does things differently, and as it’s not even obligatory to convert the input to a string, I think that a more general formulation (or one that repeats the wording from the rule set) will be less innocuous, for example:

```
The processor <rfc2119>may</rfc2119> output <code>124.84</code> and
<code>"the value of $v is:"</code> to an ·implementation-defined· destination.
```

ChristianGruen (commented):
Hope this helps; the remainder of the revision is just fine.

specifications/xpath-functions-40/src/function-catalog.xml
Arithmeticus: thanks, reflected in the new commit

ChristianGruen (approved):

ndw:
The CG agreed to accept this PR at [meeting 051](https://qt4cg.org/meeting/minutes/2023/10-24.html)

PR #770, opened 2023-10-23
566: Use fn:decode-from-uri in fn:parse-uri

ndw:
I think this is sufficient to close issue 566. @ChristianGruen could you let me know if you agree?

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -27563,6 +27563,7 @@ declare function some(
          and the <emph>filepath</emph> is the <emph>string</emph>.
          </p>
 
+
          <p>If the <emph>scheme</emph> is known to be hierarchical, or known
          not to be hierarchical, then <emph>hierarchical</emph> is set accordingly.
          Exactly which schemes are known to be hierarchical or
@@ -27575,15 +27576,20 @@ declare function some(
          otherwise <emph>hierarchical</emph> is
          <code>true</code> if <emph>string</emph> begins with <code>/</code> and <code>false</code> otherwise.</p>
 
-         <p>If <phrase diff="add" at="2023-07-07">the scheme is not known or is known to be <code>file</code> and</phrase>
-         the <emph>string</emph> matches <code>^//*([a-zA-Z]:.*)$</code>,
+         <p>If the scheme is not known or is known to be <code>file</code>:</p>
+
+         <ulist>
+            <item><p>If the <emph>string</emph> matches <code>^//*([a-zA-Z]:.*)$</code>,
          the <emph>authority</emph> is empty and the <emph>string</emph> is
-         the first match group. Otherwise, if the <emph>string</emph>
-         matches <code>^///*([^/]+)(/.*)?$</code> then the <emph>authority</emph>
+         the first match group.</p></item>
+         <item><p>Otherwise, if the <emph>string</emph>
+         matches <code>^///*([^/]+)(/.*)?$</code>, the <emph>authority</emph>
          is the first match group and the <emph>string</emph> is the second
-         match group. If the <emph>string</emph> does not match either
+         match group.</p></item>
+         <item><p>If the <emph>string</emph> does not match either
          regular expression, the <emph>authority</emph> is the empty sequence
-         and the <emph>string</emph> is unchanged.</p>
+         and the <emph>string</emph> is unchanged.</p></item>
+         </ulist>
 
          <p>If the <emph>string</emph> matches <code>^//*([a-zA-Z]:.*)$</code>,
          the <emph>authority</emph> is empty and the <emph>string</emph> is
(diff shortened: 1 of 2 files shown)


ndw:
I believe I have fixed the remaining issues in #566.

I've removed the "merge without discussion" label, but I don't think it will require *much* discussion.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -27895,20 +27904,26 @@ path with an explicit <code>file:</code> scheme.</p>
...
+        <code>false()</code> or if the scheme is known to be
+        non-hierarchical. (In other words, schemes are hierarchical by
+        default.)</p>
+
+        <ulist>
+           <item><p>If the <code>scheme</code> is
+        known to be non-hierarchical, it is delimited by a trailing
+        <code>:</code>.</p>
+        </item>
+        <item><p>If the <code>scheme</code> is <code>file</code> and the <code>unc-path</code>
```
ChristianGruen, line 27920:
```suggestion
        <item><p>Otherwise, if the <code>scheme</code> is <code>file</code> and the <code>unc-path</code>
```

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -27194,15 +27195,20 @@ declare function some(
          otherwise <emph>hierarchical</emph> is
          <code>true</code> if <emph>string</emph> begins with <code>/</code> and <code>false</code> otherwise.</p>
 
-         <p>If <phrase diff="add" at="2023-07-07">the scheme is not known or is known to be <code>file</code> and</phrase>
-         the <emph>string</emph> matches <code>^//*([a-zA-Z]:.*)$</code>,
+         <p>If the scheme is not known or is known to be <code>file</code>:</p>
+
+         <ulist>
+            <item><p>If the <emph>string</emph> matches <code>^//*([a-zA-Z]:.*)$</code>,
```
ChristianGruen, line 27201: The rule on the pattern `^//*([a-zA-Z]:.*)$` (and the following rules) are now defined twice; see the section after the bullet list. Is it supposed to be like that?

ndw:
I've accepted your comment about /Otherwise/. The second issue was a failure on my part to delete a now redundant paragraph. I've fixed that as well.

ChristianGruen:
Thanks, Norm. I’m not sure if that’s enough, though. I tried to play through the current rules for `parse-uri('http://X')`:

> Processing begins with a string that is equal to the $uri. If the string contains any backlashes (\), replace them with forward slashes (/).

→ `uri`: `http://X`, `string`: `http://X`

> Otherwise, if the string matches `^([a-zA-Z][A-Za-z0-9\+\-\.]*):(.*)$`:

→ `scheme`: `http`, `string`: `//X`, `filepath`: ()

> If the string is the empty string, hierarchical is the empty sequence (i.e. not known), otherwise hierarchical is true if string begins with `/` and `false` otherwise.

→ `hierarchical`: `true()`

> If the authority matches `^(([^@]*)@)(.*)(:([^:]*))?$`, …

→ authority is not defined at this stage (anymore). Maybe (some of?) the bullet points should always apply, no matter what *scheme* is?

I’m sorry if I’m probably too lazy to think about what the best solution would be to get all rules straight. Right now, I can only observe that I don’t manage to the correct results with the current spec (provided, of course, that I implemented everything correctly).

ndw:
I'll review your example and the spec tomorrow. I thought I adjusted the spec to address the issue, but I've made several more changes since then, perhaps I broke it. I'm a bit unsatisfied with my own implementation right now with regards to escaping paths, so I'll revisit.

ndw:
With the version I just pushed, I believe that the rules for `parse-uri('http://X')` play out as follows: 

> Processing begins with a string that is equal to the $uri. If the string contains any backlashes (), replace them with forward slashes (/).

→ `uri`: `http://X`, string: `http://X`

> Otherwise, if the string matches `^([a-zA-Z][A-Za-z0-9\+\-\.]*):(.*)$`:

→ `scheme`: `http`, `string`: `//X`, `filepath`: `()`

> The scheme is known to be hierarchical

→ `hierarchical`: `true()`

> Otherwise, if the string matches `^///*([^/]+)?(/.*)?$`, the authority is the first match group and the string is the second match group.

→ `authority`: `X`, `string`: ` `

> otherwise userinfo is the empty sequence

→ `userinfo`: `()`

> 4. the host is the empty sequence

→ `host`: `()`

> Otherwise, the port is the empty sequence

→ `port`: `()`

> If the string is the empty string, the path is the empty sequence

→ `path`: `()`

> The path separator...

→ `path-segments`: `()`

> The query seperator…

→ `query-parameters`: `map{}`

ChristianGruen:
@ndw Thanks; things have improved.

I think it’s still too early to finalize this issue. Here’s my latest summary of the deviations between the expected test results and the results I get when according to the rules in the spec (please correct me if you think I’ve misinterpreted some rules):

* [fn-build-uri-2023-10-25.log](https://github.com/qt4cg/qtspecs/files/13168757/fn-build-uri-2023-10-25.log)
* [fn-parse-uri-2023-10-25.log](https://github.com/qt4cg/qtspecs/files/13168758/fn-parse-uri-2023-10-25.log)

ndw:
I think the backtick change will fix the "invalid entity" messages.

Looking at, for example:
```
fn-build-uri-from-parse-012
fn:build-uri(map { "uri": "file:///c:/path/to/file", "scheme": "file", "path": "/c:/path/to/file", "path-segments": ("", "c:", "path", "to", "file") })
Result: file:///c%3A/path/to/file
Expect: any of { "file:///c:/path/to/file" }
```
I get `file:///c:/path/to/file` instead of `%3A` because by the rules described in https://github.com/qt4cg/qtspecs/issues/774, `:` is not escaped. 

Similarly for `@` in `mailto:`, `+` in `tel:` and `:` in several other schemes.

The `news:` and `urn:` schemes are non-hierarchical.

ChristianGruen:
> I think the backtick change will fix the "invalid entity" messages.

…right.

> I get `file:///c:/path/to/file` instead of `%3A` because by the rules described in #774, `:` is not escaped.

Do you plan to…

1. revise the rules for `fn:encode-for-uri`,
2. don’t use `fn:encode-for-uri` for `fn:parse-uri`, or
3. adapt the tests?

> The `news:` and `urn:` schemes are non-hierarchical.

I think we should make the detection of some non-hierarchical (and hierarchical?) schemes obligatory in the spec. Otherwise, we won’t be able to test those cases.

ndw:
I don't think we can change what `fn:encode-for-uri` does, but I also don't think it's ever been useful for this purpose. It says:

> This function escapes URI delimiters and therefore cannot be used indiscriminately to encode “invalid” characters in a path segment.

If/when we get agreement on the rules that I outlined in #774, I plan to add those rules to `fn:build-uri`.

With respect to non-hierarchical schemes, I already [added a list](https://qt4cg.org/pr/770/xpath-functions-40/Overview.html#parse-build) to the spec.

ChristianGruen:
I cherry-picked 3 test cases and tried to find out why they fail (they should all be contained in the log files from above):

#### `fn-parse-uri-011c`
* Query: `fn:parse-uri("//uncname/path/to/file", map { "unc-path": true() })`
* Expected by the test suite: `("", "uncname", "path", "to", "file")`
* …but I get: `("", "", "uncname", "path", "to", "file")`
* …as the spec says: “A path-segments sequence is constructed by tokenizing the string on the path separator and applying uri decoding on each token.”

#### `fn-parse-uri-012`
* Query: `fn:parse-uri("file:///c:/path/to/file")`
* Expected by the test suite: `"path": "/c:/path/to/file"`
* …but I get: `"path": "c:/path/to/file"`
* …as the spec says: “…the string matches `^//*([a-zA-Z]:.*)$`, the authority is empty and the string is the first match group. … If the string is the empty string, then path is the empty sequence, otherwise path is the whole string.”

#### `fn-parse-uri-020`
* Query: `fn:parse-uri("ldap://[2001:db8::7]/c=GB?objectClass?one")`
* Expected by the test suite: `"path": "/c=GB"`
* …but I get: `"path": "/c=GB?objectClass"`
* …as the spec says: “If the string matches `^(.*)\?([^\?]*)$`”<br/>→ maybe it should be “If the string matches `^(.*?)\?(.*)$`” ?

Issue #767, opened 2023-10-23
parse-html(): case of SVG element names


michaelhkay:
Gunther Rademacher points out at https://github.com/qt4cg/qt4tests/issues/57 that the expected test results for parse-html() expect SVG elements to be output in lower case; and as far as I can see, this is consistent with the spec. But is it useful? It means we're producing XML that isn't valid according to the SVG schema, and that presumably might be rejected by tools that expect to handle valid SVG.

In passing, I note that it's very hard to work out from the parse-html() spec what the function actually does; it's somehow written as if it's obvious, but it never gets around to actually saying it explicitly. It also doesn't say what happens when the `method` option is omitted, as it is in these tests.

In addition, it seems that we don't have any tests that exercise the various options of the function, e.g. the ability to consue binary as well as character input.

michaelhkay:
I note that in §15.6.10 we say

> For example, in [[DOM: Living Standard]] section 13.2.9 Coercing an HTML DOM into an infoset, it uses a Unnnnnn escape sequence. That would map : to U00003A.

The  cited document does not appear to have a section 13.2.9, or to mention the concept of an infoset. This highlights the danger of citing a "living" (i.e. unstable) standard. I can't find any text in the current HTML DOM spec that talks about how characters that aren't allowed in NCNames should be treated.

rhdunn:
It would be useful to link to the sections in the external specs, to make this easier to follow.

michaelhkay:
Thanks, the specific problem appears to be that we link to §13.2.9 of the WHATWG DOM spec, not the WHATWG HTML spec.

I don't know if the section numbers/identifiers of the whatwg specs are stable enough for us to link to?

michaelhkay:
I believe that the only outstanding actions under this issue are (a) to fix the reference to the HTML5 spec, and (b) to add tests for the parse-html options parameter.

PR #772, opened 2023-10-24
Revise the fn:parse-html rules to make them clearer to follow.

rhdunn:
This actions the comment in #767 that the rules for the function are unclear.

The SVG element name handling is correct in the spec per the comments in https://github.com/qt4cg/qt4tests/issues/57, so don't need changing.


ChristianGruen (approved):

michaelhkay:
The spec still lacks a basic statement of what the function does: it's probably obvious that it parses HTML, but it doesn't actually say so. (It shouldn't rely on the summary, which should only summarise what is said in more detail in the Rules).

I suggest something like: "The function takes as input lexical HTML contained in the string or octet sequence supplied in the $html argument, and parses this to construct an XDM document node. Conceptually, this is done by parsing the HTML to construct a DOM representation following the rules in the relevant HTML specifications, and then converting the DOM to an XDM document node as described in section 15.6."

rhdunn:
I'll add something similar to that to the rules. Note, though, that providing an overview of the behaviour of the function (repeating/rephrasing the summary) is not consistent across the specification, e.g. https://www.w3.org/TR/xpath-functions-31/#func-json-doc -- that does not have a sentence in the rules that it parses the JSON document at the specified resource location.

ChristianGruen:
I agree with @rhdunn: It's often not easy to judge (by browsing through the existing function set) what's exactly expected to be contained in the summary and the rules, and what's supposed to be informal and formal.

On the other hand, I clearly see it would be time consuming to get the full XQFO specification unified. Maybe there are specific functions in the spec that we can regard as references for future additions?

michaelhkay:
Section §1.5 of F&O describes the notation conventions. In particular: _The first section in the proforma is a short summary of what the function does. This is intended to be informative rather than normative._

The summary should not say anything that is not already said elsewhere; equally, it should describe the "normal" behaviour of the function and can ignore edge cases and error cases.

rhdunn:
I've updated the element and attribute node-name rules to account for SVG and MathML case-sensitive local names as noted by @hsivonen.

ndw:

PR #782, opened 2023-10-26
469: array:of-members, map:of-pairs: Signatures, Examples

### LINKED ISSUE #469: array:of-members, map:of-pairs: Signatures, Examples
Just trivia:

a) The parameter name of `array:of-members` is [`$input`](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-array-of).
    `$members` may be a better choice (or we should change `map:of-pairs($pairs)` to `map:of-pairs($input)`).

b) The type of `$pairs` is [`record(key as xs:anyAtomicType, value as item()*, *)*`](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-map-of).
    Shouldn’t it be `record(key as xs:anyAtomicType, value as item()*)*` (without the trailing `, *`)?
    If the current syntax is correct, an explanatory comment could be helpful.

c) ~One `map:of` example needs to be fixed: `map:of((map:entry(0, "no"), map:entry(1, "yes")))`.~ See #607

See #508 for the proposal to rename `map:of` to `map:of-pairs`.

ChristianGruen:
I’ve eventually renamed `$pairs` to `$input` (I didn’t rename `$input` to $members` as initially suggested, as we have `$member` parameters in other functions that are of type `item()*`, not `record(value as item()*)`).

Closes #469

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -19564,8 +19564,8 @@ return fold-left($MAPS, map { },
    <fos:function name="of-pairs" prefix="map">
       <fos:signatures>
          <fos:proto name="of-pairs" return-type="map(*)">
-            <fos:arg name="pairs" 
-                     type="record(key as xs:anyAtomicType, value as item()*, *)*" 
+            <fos:arg name="input" 
+                     type="record(key as xs:anyAtomicType, value as item()*)*" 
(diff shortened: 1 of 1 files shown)


michaelhkay (approved):

ndw:
The CG agreed to merge this PR at [meeting 052](https://qt4cg.org/meeting/minutes/2023/10-31.html)

PR #793, opened 2023-10-30
704: Context Value Expression → Context Value Reference

### LINKED ISSUE #704: Context Value Expression → Context Value Reference
The specification defines Variable References for accessing values bound to a variable, and we should rename the equivalent operation for accessing the context from “Context Value Expression” to “Context Value Reference” (even more so if we should decide to introduce a Context Value Declaration later on, as discussed in #755).

Related: https://github.com/qt4cg/qtspecs/pull/703#issuecomment-1719345430

ChristianGruen:
Closes #704


ndw:

PR #794, opened 2023-10-30
216: fn:unparsed-text: End-of-line characters

### LINKED ISSUE #216: fn:unparsed-text: End-of-line characters
**UPDATED PROPOSAL (2023-10-31):**

Change the default behavior and normalize the input as known from XML (https://www.w3.org/TR/xml/#sec-line-ends). Even though backward-incompatible, the change should be noninvasive enough, and it allows us to dispense with adding options to this and other functions.

----

**ORIGINAL PROPOSAL:**

### Motivation

`fn:unparsed-text` can only be used if the input is valid. Next, the result may contain OS-specific newlines that may need to be normalized in a subsequent step. Two additional parameters might simplify things:

* `normalize`: normalizes line endings in the input, as done by `fn:doc`, `fn:parse-xml` and `fn:parse-xml-fragments` and as defined in https://www.w3.org/TR/xml/#sec-line-ends (inspired by the discussion in #212).
* `fallback`: Process characters that are not valid in the version of XML supported by the implementation (inspired by `fn:parse-json`)

### Signature

```xquery
fn:unparsed-text(
  $href    as xs:string?,
  $options as map(*)
) as xs:string?
```

### Rules

The entries that may appear in the `$options` map are as follows:

Key | Meaning
-- | --
`encoding` | The name of an encoding … (see existing rules)
`normalize-newlines` |  `false` by default. If set to `true`, two-character sequences `#xD#xA` and any `#xD` that are not followed by `#xA` are translated to single `#xA` characters.
`fallback` | A function which is called when the input contains an escape sequence that represents a character that is not valid […] (see [`fn:parse-json`](https://qt4cg.org/specifications/xpath-functions-40/Overview-diff.html#func-parse-json)).

ChristianGruen:
Closes #216. See this issue for more details on the proposed change.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -15747,6 +15747,11 @@ else $c[1] + sum(subsequence($c, 2))</eg>
          </olist>
          <p>The result of the function is a string containing the string representation of the
             resource retrieved using the URI.</p>
+         <p diff="add" at="2023-05-19">End-of-line characters are normalized as known from the
+            <xnt spec="XML" ref="sec-line-ends">end-of-line handling</xnt> in <bibref ref="xml"/>,
+            by translating both the two-character sequence <code>x0D</code> <code>x0A</code>
+            and any <code>x0D</code> that is not followed by <code>x0A</code>
+            to a single <code>x0A</code> character.</p>
       </fos:rules>
       <fos:errors>
          <p>A dynamic error is raised <errorref class="UT" code="1170"
@@ -15857,6 +15862,9 @@ else $c[1] + sum(subsequence($c, 2))</eg>
 ]]></eg>
          </fos:example>
       </fos:examples>
+      <fos:history>
+         <fos:version version="4.0">Changed in 4.0: end-of-line characters in the input are normalized.</fos:version>
+      </fos:history>
 
    </fos:function>
    <fos:function name="unparsed-text-lines" prefix="fn">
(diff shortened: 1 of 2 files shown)


michaelhkay (changes requested):
1. "As known from XML" is not very idiomatic, and more importantly, not very precise. I'd suggest a specific reference to section 2.11 End of Line Handling: "are normalized as described in [XML] section 2.11 End of Line Handling.
2. Why XML 1.0 and not XML 1.1?
3. This is a backwards incompatibility. I'm not necessarily opposed to it but the fact needs highlighting in Appendix G. Might there be applications that need to know what line endings were used? Should this be added to a new options parameter?

ChristianGruen:
> 1. "As known from XML" is not very idiomatic, and more importantly, not very precise. I'd suggest a specific reference to section 2.11 End of Line Handling: "are normalized as described in [XML] section 2.11 End of Line Handling.

Thanks. That’s what I tried to do, but I might need help how to do it properly:

image

> 2. Why XML 1.0 and not XML 1.1?

Happy to change that.

> 3. This is a backwards incompatibility. I'm not necessarily opposed to it but the fact needs highlighting in Appendix G. Might there be applications that need to know what line endings were used? Should this be added to a new options parameter?

I had already added an item to Appendix G, but I assume it’s supposed to be more comprehensive, or wrongly formatted/edited?

ndw:
This worked for me:
```
         <p>Test: <xspecref spec="XML" ref="sec-line-ends"/>.</p>
```

ndw:

PR #795, opened 2023-10-30
655 fn:sort-with

### LINKED ISSUE #655: fn:sort-with: Comparators
See https://github.com/qt4cg/qtspecs/issues/93#issuecomment-1017937220:

> One solution to more powerful sorting would a variant of fn:sort that uses a comparator function. We've resisted this in the past because we can't trust a user-supplied comparator function to be well behaved (e.g. transitive). I wonder how serious an obstacle this is?

We should try to trust; there are much more hidden pitfalls in the existing language.

`fn:sort` could be extended with a `comparator` parameter with two arguments and returning `xs:boolean`:

```xquery
(: returns John, Joe, Jim, Jack :)
sort(('Jack', 'Joe', 'Jim', 'John'), comparator := op('>'))

(: returns -1, 2, -3 :)
sort((-3, -1, 2), comparator := fn($a, $b) { abs($a) < abs($b) })
```

Obviously, if a comparator is supplied, other parameters (`collation(s)`, `key(s)`, `ascending`, see #623) must not be specified in parallel.

ChristianGruen:
Closes #655

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -19891,6 +19891,10 @@ return sort($in, $SWEDISH)</eg>
                before the second.</p></item>
             <item><p>Otherwise, the first is returned after the second.</p></item>
          </ulist>
+         <p>Users are responsible for supplying a transitive comparator; otherwise, the result
+            might not be correctly sorted. An example for a non-transitive and thus unsuitable
+            comparator is <code>fn($a, $b) { if($a mod 2 = 1) then 1 else -1 }</code>,
+            as it considers odd numbers to be greater than even numbers.</p>
          <p>Sorting is <emph>stable</emph>, which means that the relative order of the input items
             is maintained.</p>
          <p>More formally, the effect of the function is equivalent to the following implementation
@@ -19916,10 +19920,6 @@ declare function sort-with(
       <fos:notes>
          <p>An implementation is free to choose any sorting algorithm, as long as the result
             returned is stable and matches the result of the implementation in XQuery.</p>
-         <p>Users are responsible for supplying a transitive comparator; otherwise, the result
-            might not be correctly sorted. An example for a non-transitive and thus unsuitable
-            comparator is <code>fn($a, $b) { if($a mod 2 = 1) then 1 else -1 }</code>,
-            as it considers odd numbers to be greater than even numbers.</p>
       </fos:notes>
       <fos:examples>
          <fos:example>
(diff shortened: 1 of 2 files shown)


michaelhkay (commented):
1. Typo: "relationshop".
2. Why a boolean comparator function rather than a more conventional 3-valued comparator?
3. Need to explain the effect of the function when there are duplicates (rather than leaving this to be deduced by reverse-engineering the sample implementation). [Actually, after applying a bit of thought, I think I can see the logic: by constructing the two-valued comparator in any way you like, you can achieve the duplicates handling that you want. But that needs explaining.]
4. Need to explain the rules for the comparator function (transitivity etc) to ensure the function is well-behaved, and the consequences of breaching the rules.
5. How close to the reference implementation does the actual implementation need to be? For example, the reference implementation has the property that $comparator(X, Y) is only called for a pair of values where X precedes Y in the input. Does the actual implementation need to have that property?

ChristianGruen:
> 2. Why a boolean comparator function rather than a more conventional 3-valued comparator?

A less-than comparator felt easier to me to use in practice, and I didn’t see real advantages for having 3 states (maybe achieving a stable sort? But this also depends on the sort algorithm). I don’t have a strong opinion on it, though.

Thanks for the remaining comments, for which I’ll try to find answers later on.

ChristianGruen:
@ndw Before further elaborating the details, maybe we could put this on next week’s agenda (if it fits in well) to find out if the majority believes that the function would be helpful, or if the enhanced `fn:sort` function is already sufficient. Thanks!

ndw:
WIll do.

line-o:
@ChristianGruen I do like your proposal.
As mentioned in the call, I am used to sort on more than one property by returning a sequence with `fn:sort`.
Your last example could be simplified to 
```xquery
function ($a, $b) {
  compare($a/@p1, $b/@p1),
  compare($a/@p2, $b/@p2)
}
```

This also opens up to use a combined function as a comparator. Something along the lines of 
```
chain(comp1#2, comp2#2)
```

ChristianGruen:
@line-o Thanks. Should be doable.

Given the two choices, I would tend to introduce a sequence of comparators to retain the classical comparator contract, and make it symmetric with the enhanced 4.0 semantics of [`fn:sort`](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-sort), which allows us to provide a sequence of keys:

```xquery
sort(reverse(-5 to 5), (), (abs#1, string#1))
```

ChristianGruen:
@ndw …should be ready to merge in a few minutes.

ChristianGruen:
@ndw …can now be merged.

ndw:
Thanks. I think we agreed to put this in the "merge without discussion" list for next week.

ndw:
The CG agreed to merge this PR without discussion at [meeting 066](https://qt4cg.org/meeting/minutes/2024/02-20.html)

PR #797, opened 2023-10-31
Edits to parse-uri()

Arithmeticus:
Attn @ndw -- these edits pertain to 6.6.1 `fn:parse-uri` and require your careful review.

- `hierarchical` option appears in the rules of the function but not its preamble
- "This function is described..." moved to anticipate the actual narrative. 
- I was a bit thrown by the phrase "This approach...is not implementation advice" because "approach" is vague and the reader is left with the impression that a good deal of _something_ that follows is non-normative. We offer non-normative prose in the rules, but that informal prose is normally followed by an equivalent description that _is_ normative. I did not make any edits in this area, but I would recommend clarification as to exactly what parts of the narrative are normative and which are non-normative.
- The period-to-colon change at 27573 is to make sure the reader remains aware of the governing "If..." clause.
- I deleted a paragraph that attempted to distill the rather complicated description of `fn:decode-from-uri`. A mere xref provides a cleaner narrative here and a more accurate description of uri decoding, and the xref is within the same document, so should not pose a burden on the reader.
- For the `query-parameters` there was a discrepancy in the data model presented (in the preamble versus in the function rules), and I opted for the array-of-maps model. If it's supposed to be the other way (simple map), let me know and I'll revise.
- `query-segments` versus `query-parameters`; I went with the latter.

Let me know where things ain't right.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -27721,7 +27721,7 @@ declare function some(
          <p>The <emph>query separator</emph> is the value of the
          <code>query-separator</code> option, and is used to construct
          a <emph>query-parameters</emph> value as follows.
-         Start with an empty array. Tokenize the <emph>query</emph> on
+         Start with an empty map. Tokenize the <emph>query</emph> on
          the <emph>query separator</emph>. For each token, identify
          the <emph>key</emph> and the <emph>value</emph>. If the token
          contains an equal sign (<code>=</code>), the <emph>key</emph>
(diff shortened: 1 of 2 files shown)


ChristianGruen:
@Arithmeticus Thanks for the edits. Note that there’s another pull requests for `fn:parse-uri` that we should probably merge first: #770. Some of your suggestions seem to have been covered in both PRs.

Arithmeticus:
Thanks, yes, there's some overlap. Once that PR is dealt with, I'll rebase the branch behind this PR and tidy up, and re-summarize what needs to be looked at.

Arithmeticus:
Revised in light of today's merge-ment. See new commit. The fifth and sixth bullet points in my cover message no longer apply.

ChristianGruen (approved):

ndw:
I will happily, and with gratitude, accept all of these proposed changes.

ndw:

PR #798, opened 2023-10-31
479: fn:deep-equal: Input order

### LINKED ISSUE #479: fn:deep-equal: Input order
#383 is about the specific order of children of element nodes.

I think we should also provide an option to ignore the top-level order of the input items:

```xquery
(: returns true: both input arguments contain the same items, but in a different order :)
deep-equal(
  (1 to 10),
  reverse(1 to 10),
  map { 'unordered': true() }
)

(: returns false: the compared elements are different :)
deep-equal(
  <a><b/><c/></a>,
  <a><c/><b/></a>,
  map { 'unordered': true() }
)

(: returns false: the second sequence contains duplicates :)
deep-equal(
  (1, 2),
  (2, 1, 1),
  map { 'unordered': true() }
)
```

ChristianGruen:
I decided to follow my initial suggestion and add `unordered` to `fn:deep-equal` instead of adding a separate function for it. My motivation:

* It’s convenient to be able to use the new option in combination with the other options.
* It will be used more often than many other options we’ve added recently.
* It seemed pretty straightforward to add, both in terms of documentation and implementation.

Closes #479

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -13582,12 +13582,6 @@ return contains-sequence(
             for the interpretation of each option appear later.</p>
             
             <fos:options>
-               <fos:option key="unordered">
-                  <fos:meaning>Controls how the input is to be compared.
-                  </fos:meaning>
-                  <fos:type>xs:boolean</fos:type>
-                  <fos:default>false</fos:default>
-               </fos:option>
                <fos:option key="base-uri">
                   <fos:meaning>Determines whether the <code>base-uri</code> of a node is significant.
                   </fos:meaning>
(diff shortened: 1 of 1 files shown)


michaelhkay (changes requested):
The option has been misspelled "unorderd" rather than "unordered" (twice). But I don't thing negative option names are a good idea anyway: ordered=true is better than unordered=false. Perhaps if we want the default generally to be false, then use permute=true in place of unordered=true.

Efficient implementation of this function requires computation of some kind of signature or hash code such that two equal items have the same hash code. If the system is going to have to implement such a hash function, I wonder if it's worth exposing it for other purposes, e.g. detecting duplicates by deep-equality.

ndw:

PR #800, opened 2023-10-31
660: Static functions, default parameters, XPST0017

### LINKED ISSUE #660: Static functions, default parameters
In the current XQuery draft in [5.18.3 Function Parameters](https://qt4cg.org/specifications/xquery-40/xquery-40.html#id-function-parameters), it’ stated that:

> If a parameter is optional, then all subsequent parameters in the list must also be optional. In other words, the parameter list includes zero or more required parameters followed by zero or more optional parameters.

I would suggest raising `XPST0017` if that’s not the case.

ChristianGruen:
I’ve chosen XPST0017 over XPST0003 (it felt more intuitive to me).

Closes #660.

### DIFF
--- a/specifications/xquery-40/src/errors.xml
+++ b/specifications/xquery-40/src/errors.xml
@@ -1093,10 +1093,10 @@ It is a static error if the name of a feature in
          <termref def="dt-generalized-atomic-type"/>.</p>
       </error>
       
-      
-      
-      
-
+      <error spec="XQ" role="xquery" code="0148" class="ST" type="static">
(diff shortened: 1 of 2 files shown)


michaelhkay (changes requested):
The use of XPST0017 doesn't match the description of this error in the appendix. Existing uses of XPST0017 all relate to errors in function calls, not errors in function declarations. In addition, this particular error is XQuery-specific which might mean that an XQ code is more appropriate. I would go for a new error code such as XQST0148.

ChristianGruen:
Thanks, r
evised as suggested.

ndw:

PR #802, opened 2023-10-31
801: non-deterministic → nondeterministic

### LINKED ISSUE #801: nondeterministic vs non-deterministic
In the specs, both “nondeterministic” and “non-deterministic” can be found. The first one is more frequent, so I guess it’s the one that’s preferred.

ChristianGruen:
Closes #801


ndw:

PR #803, opened 2023-10-31
651: fn:log → fn:message

### LINKED ISSUE #651: fn:log → fn:message
I think we've made a mistake in the choice of name for this function. Without the namespace prefix, it looks far too much like a function that computes logarithms. Also, it would be nice in the future to find some way of allowing the math functions to be called without a namespace prefix, and this choice scuppers that possibility.

I think my choice would be fn:message.

ChristianGruen:
Closes #651


ndw:

PR #806, opened 2023-11-01
566 A few minor fixes for parse-uri

ndw:
As CG observes in [issue 566](https://github.com/qt4cg/qtspecs/issues/566#issuecomment-1787666119), there are still a couple of small problems with `fn:parse-uri()`.

1. The regular expressions used to parse the fragment identifier and query are incorrect. The URI specification allows `?` to appear in a query string and `#` to appear in a fragment identifier, so the expressions have been rewritten to match everything after the first `?` and `#`, respectively, even if they contain additional `?` or `#` characters.
2. The description of how to interpret the regular expression for parsing a Windows file path preceded by slashes was incorrect. It caused the leading "/" to be lost. That's been corrected.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -27536,13 +27536,13 @@ declare function some(
 
          <p>Strip off the fragment identifier and any query:</p>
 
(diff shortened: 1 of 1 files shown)


ndw:

PR #808, opened 2023-11-01
Tweaks to highlight new functions in F&O

ndw:
Following a discussion on the XML.com slack, this is a little lunchtime hackery...

Any function listed in the `new-functions` section of the changes appendix or identified with an `ednote` that contains the string `New in 4.0` will be marked as 🆕 in the specification. The 🆕 occurs in the ToC, the drop-down function list, and to the left of the section title.

(I'm just going to merge this as it has no spec changes and the effect won't be visible in the PR anyway.)


ChristianGruen:
Great!

PR #811, opened 2023-11-01
Highlight changed functions in the ToCs and headings

ndw:
Obviously, I should have done icons for functions that have changed as well. So now I have. I'm not convinced that the 🆙 emoji is sufficiently different from the 🆕 emoji. Maybe I should try to make them different colors or something. But it's a start.


ChristianGruen:
Awesome!

Issue #822, opened 2023-11-08
XQuery, XQFO: Edits (pool)


ChristianGruen:
XQuery spec:

* [x] 4.3.4 Context Value Reference → 4.3.4 Context Value Reference**s** (plural, in alignment with the other expressions)
* [x] Move **4.3.4** before **4.3.3 Parenthesized Expressions** (and after **4.3.2 Variable References**)
* [x] 4.19 Switch Expression → 4.19 Switch Expression**s**
* [x] Try/Catch Expressions: There’s no `CatchErrorList` anymore

XQFO spec:

* [x] Unify representation of equivalent examples, implementations, …see https://github.com/qt4cg/qtspecs/pull/828#issuecomment-1807222990
* [x] Add History sections for new functions
* [x] `9^XXX outputs "4451"` → `"3185"`
* [x] errors are signaled → raised (#783)

PR #828, opened 2023-11-11
516 Add position argument to HOF callbacks

### LINKED ISSUE #516: Add position argument to HOF callbacks
The coercion rules now allow a supplied function item to have lower arity than the signature of the declared type; the effect is that the information supplied in the additional arguments is ignored.

One of the intended use cases for this was to allow existing higher-order functions to be extended while retaining backwards compatibility. For example, in `fn:filter`, we can change the required type of the predicate function from `function(item()) as xs:boolean` to `function(item(), xs:positiveInteger) as xs:boolean`, with the second argument supplying the position of the item being tested. A function that isn't interested in the position can just ignore it, so existing calls will continue to work.

I propose that we add a position argument to the callbacks for:

```
fn:filter
fn:for-each
fn:for-each-pair
fn:partition
fn:items-after
fn:items-before
fn:items-starting-where
fn:items-ending-where
array:filter
array:for-each
array:for-each-pair
```

Other candidates include

```
fn:all
fn:some
fn:index-where
array:index-where
```

but I suggest we leave these unless someone can think of a use case.

ChristianGruen:
I have added positional parameters to the following functions:

```
array:filter
array:fold-left
array:fold-right
array:for-each
array:for-each-pair
array:index-where
fn:every
fn:filter
fn:fold-left
fn:fold-right
fn:for-each
fn:for-each-pair
fn:index-where
fn:items-after
fn:items-before
fn:items-ending-where
fn:items-starting-where
fn:iterate-while
fn:partition
fn:some
```

Comments:
* For `fn:every` and `fn:some`, the additional parameter seemed useful to me, as a positional variable has been requested for quantifier expressions in the past.
* I’ve also added positional variables to folds.
* I’ve unified and simplified the formal XPath/XQuery equivalencies in the rule sets.
* I’ve dropped some XSLT equivalencies because I felt that the XQuery representations are more concise (I certainly won't mind if they're added back).

Closes #516.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -12097,7 +12097,7 @@ return exists($break)
       </fos:rules>
       
       <fos:notes>
-         <p>The effect of the function is equivalent to the following XSLT expression:</p>
+         <p>The function is equivalent to the following XSLT expression:</p>
          <eg><![CDATA[
 <xsl:for-each-group select="$values" group-by="." collation="{$collation}">
   <xsl:sequence select="current-group()[2]"/>
@@ -13265,16 +13265,13 @@ return ends-with-sequence(
       <fos:rules>
          <p>Informally, the function returns <code>true</code> if <code>$input</code> contains a consecutive subsequence matching <code>$subsequence</code>,
             when items are compared using the supplied (or default) <code>$compare</code> function.</p>
-         <p>More formally, the effect of the function is equivalent to the following implementation in XQuery:</p>
+         <p>More formally, the function is equivalent to the following expression in XQuery:</p>
          <eg><![CDATA[
-declare function contains-sequence(
-  $input       as item()*,
-  $subsequence as item()*,
-  $compare     as function(item(), item()) as xs:boolean := fn:deep-equal#2
-) as xs:boolean {
+declare function contains-sequence($input, $subsequence, $compare) {
   starts-with-sequence($input, $subsequence, $compare) or
   exists($input) and contains-sequence(tail($input), $subsequence, $compare)
 };
+contains-sequence($input, $subsequence, $compare)
 ]]></eg>
       </fos:rules>
       <fos:notes>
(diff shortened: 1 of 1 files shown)


michaelhkay:
Looks good on a first skim but there's a lot of scope here for introducing errors so I want to check it more thoroughly. It's a little unfortunate that many of the examples that make use of a positional parameter are somewhat unconvincing as plausible use cases.

michaelhkay:
1. You often use the phrase "the effect of the function is equivalent to the following expression in XQuery" and then follow this with something that isn't actually an expression (because it includes a function declaration) - see for example fn:contains-sequence. (The expansion for contains-sequence is particularly confusing. In most cases like this you've just given an expression that can be used as the implementation of the function.)
2. In deep-equal, dropping the function signature of the internal equal-strings() function seems to lose clarity here.
3. Generally, the phrase "the effect of the function is equivalent to the following expression" means you can treat the expression as the implementation of the function. But with fn:chain, the expression returns a function, not the result of calling the function.
4. A better example for every() might be `every(*, fn($elem, $pos){name($elem) = if (($pos mod 2) = 1) then "dt" else "dd"})`
5. For fold-left and similar examples I think it would be better if the internal function were clearly indicated as such, perhaps `declare %private function fold-left-helper(....) {...}` rather than using the same name as the public function. (I'm perfectly happy to see the XSLT equivalents go, by the way).
6. fold-right(). Should this call the internal fold-right() function rather than fold-left()? (fold-right always confuses me and I haven't fully checked it).
7. fn:for-each second note (existing text) I can't see the relevance of array:build here.
8. items-before() and items-ending-where(): why supply 0 as the second argument of subsequence() rather than 1?
9. iterate-while() in the prose invokes "$action($input)", but $action has two arguments. This one seems a little different in that (if I understand correctly), $pos is not the position in a sequence, but a count of the number of calls that have been made. It might be worth drawing attention to this. Also, the "equivalent" expression calls iterate-while($input, $predicate, 1) which is surely missing the $action argument. The example that returns 3628800 surely deserves some explanation -- what on earth is this function actually doing?
10. partition(): the prose says it calls split-when() with two arguments, but there are three.
11. array:fold-left() - old text in rules hasn't been deleted but is now redundant
12. the use of "for member" in array:for-each() is equally applicable to array:filter().

ChristianGruen:
The feedback has been helpful, and I feel I’ve overdone it, out of laziness: I found inconsistencies in the existing equivalences, and I started fixing some of them, but I should rather have done it separately.

I’ve redesigned the PR to focus on the actual task. As a result, some code snippets in the rules have become unnecessarily complex. I plan to simplify them in #822, along with considering items 1.–3. and 5. of your feedback.

I’ve incorporated the remaining suggestions, and I hope the result is ready for merging.

ndw:

PR #833, opened 2023-11-15
Fix the line endings, force a single lf in text files

ndw:
Hello,

Per discussion on the list, this PR changes the way `git` handles line endings so that text files will always, exclusively have line endings delimited by a single `lf`. This should be a largely transparent change for Mac/Unix users. On Windows, it means that checked out files will have `lf` line termination. If your favorite editing tool can handle this, then you don't have to care. Even if your editor saves files with `cr`/`lf` line endings, `git` will turn them back into single `lf` endings when you commit your changes.

Hat tip to @ChristianGruen for keeping focus on this issue.

This PR has no technical changes, and effects a relatively small number of files. I'll leave it here for a day or two, then I'm going to be inclined to merge it. Object now, if you object :-)


ChristianGruen:
Thanks, Norm. If I’m not mistaken, it’s sufficient to have this following single entry in the `.gitattributes` file:

```
* -text eol=lf
```

ChristianGruen:
> If I’m not mistaken, it’s sufficient to have this following single entry in the `.gitattributes` file:

Well, I should be careful causing too much confusion. Should be fine as it is!

ndw:
With a hypen, `-text eol=lf`?

The advice I read online was that if you're going to change the way text files are delimited, you should explicitly identify binary files so that they *don't* get manged.

ChristianGruen:
> The advice I read online was that if you're going to change the way text files are delimited, you should explicitly identify binary files so that they _don't_ get manged.

…I think you are just right. Sorry for the confusion!

ChristianGruen (approved):

PR #849, opened 2023-11-21
847 Allow uri-structure-record keys to have empty sequence values

### LINKED ISSUE #847: build-uri() - is {"port":()} legal?
The (only) example in the spec of a build-uri() call specifies `{"port":()} `. But `uri-structure-record` has 

`port? as xs:string,`

which means that the empty sequence is not a valid value. Either the record structure should be changed to specify the type as `xs:string?`, or the example should be changed.

ndw:
Fix #847 

The description of `build-uri` was written with the expectation that if a key was in the map, it's value should be used. I don't really want to replace every occurrence of 

    if `x` is present in the map

with

    if `x` is present in the map and its value is not the empty sequence

So I've attempted to justify that globally with the following paragraph at the beginning of the description:

> The components are derived from the contents of the $parts map. To simplify the description below, any key whose value is the empty sequence is ignored; this is equivalent to the key not being present in the map.

I'm not hugely proud of that bit of prose though. Suggestions for improvements most welcome.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -29,8 +29,8 @@
          <fos:field name="path" type="xs:string?" required="false"/>
          <fos:field name="query" type="xs:string?" required="false"/>
          <fos:field name="fragment" type="xs:string?" required="false"/>
-         <fos:field name="path-segments" type="array(xs:string)?" required="false"/>
-         <fos:field name="query-parameters" type="array(record(key? as xs:string, value? as xs:string, *))?" required="false"/>
+         <fos:field name="path-segments" type="xs:string*" required="false"/>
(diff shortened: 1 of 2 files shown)


ChristianGruen (commented):
> So I've attempted to justify that globally with the following paragraph at the beginning of the description:
> 
> > The components are derived from the contents of the $parts map. To simplify the description below, any key whose value is the empty sequence is ignored; this is equivalent to the key not being present in the map.
> 
> I'm not hugely proud of that bit of prose though. Suggestions for improvements most welcome.

I would tend to keep the strict record definition, as `fn:parse-uri` doesn’t include empty sequences in the result either (…but both will be fine).

Next, I noticed that the types for `path-segments` and `query-parameters` need to updated.

michaelhkay:
I'd be inclined to phrase it as "In the description below, a value is considered to be present in the map if the relevant field exists and is non-empty".

ndw:

PR #855, opened 2023-11-24
844 New sequence functions: names

### LINKED ISSUE #844: New sequence functions: names
Observations:

A. What about renaming `fn:contains-sequence`, `fn:starts-with-sequence` and `fn:ends-with-sequence` to `fn:contains-items`, `fn:starts-with-items` and `fn:ends-with-items`, in alignment with `fn:items-at`, `fn:items-before`, etc.? If we add equivalent functions for arrays, it could be `array:contains-members`, etc.

B. It seems confusing to have `fn:items-starting-where` and `fn:items-after`, instead of `fn:items-starting-after`. Maybe we can think of alternative (shorter) names for `fn:items-starting-where` and `fn:items-ending-where`? – I know we’ve discussed before; I raised it again due to user feedback.

ChristianGruen:
Closes #844

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -13017,9 +13017,9 @@ return error((), 'Duplicate IDs found: ' || string-join($ids, ', '))</eg>
       </fos:history>
    </fos:function>
    
-   <fos:function name="starts-with-sequence" prefix="fn">
+   <fos:function name="starts-with-subsequence" prefix="fn">
       <fos:signatures>
-         <fos:proto name="starts-with-sequence" return-type="xs:boolean">
+         <fos:proto name="starts-with-subsequence" return-type="xs:boolean">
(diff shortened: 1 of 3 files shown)


michaelhkay:
I am having second thoughts about dropping `items-starting-where`, in particular (but also `items-ending-where`). I think items-starting-where probably has more use cases than items-after.  Note also that `items-after()` can be simply derived as `items-starting-where() => tail()`, but it's less easy to derive `items-starting-where()`. I'm now inclined to keep these with the names `subsequence-starting-where()` and `subsequence-ending-where()`.

ChristianGruen:
> I'm now inclined to keep these with the names `subsequence-starting-where()` and `subsequence-ending-where()`.

…which really gets awkward, and we'd have `array:subarray-starting-where`.

I'd very much be in favor to resort to `fn:take-while` and `fn:drop-while`/`fn:skip-while`, which are available in various other languages.

Issue #878, opened 2023-12-06
Proposed extension to subsequence


ChristianGruen:
Copied from https://github.com/qt4cg/qtspecs/issues/844#issuecomment-1841415417:

-------

I'm thinking again about integrating the items-* quartet into a heavily overloaded `subsequence` function.

Must supply zero or one of:

- $start - the start position
- $from - a predicate, such that the start is the first item to match the predicate
- - defaulting to 1

And zero or one of:

- $length - the number of items to include
- $while - a predicate, the subsequence takes items so long as the predicate is true
- $until - a predicate, the subsequence takes items up to and including the first for which the predicate is false
- - defaulting to the end of the sequence.

This omits the "items-after" combination, but that one is easily achieved using `tail(subsequence(from:="xxx"))`.

michaelhkay:
Yes, I left out the "items-after" combination partly because I don't think it's needed as much as the others, and partly because it's easy to drop the first item of the result if unwanted.

michaelhkay:
This was dropped in favour of the new `subsequence-where()` function.

PR #879, opened 2023-12-06
844 New sequence functions: names

### LINKED ISSUE #844: New sequence functions: names

ChristianGruen:
Closes #844. The `items` keyword in the function names (excluding `items-at`) has been changed to `subsequence`.

See #878 for the controversial discussion on what to do with `subsequence-(after|before|starting-where|ending-where)`.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -13009,9 +13009,9 @@ return error((), 'Duplicate IDs found: ' || string-join($ids, ', '))</eg>
       </fos:history>
    </fos:function>
    
-   <fos:function name="starts-with-sequence" prefix="fn">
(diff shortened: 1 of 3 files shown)


ndw:

Issue #843, opened 2023-11-18
Standard, array & map functions: Equivalencies


ChristianGruen:
In many threads (#135, others), we have discussed how to align the functions for sequences, arrays, and maps. This is an attempt to summarize the status quo, and I hope to keep it up-to-date in the coming weeks.

The 4.0 functions are the ones with the keyword *new* attached. If the function is followed by a question mark, there may be an existing issue for its addition, or it may be consistent to add it.

Please note that the data types have fundamental differences, so it’s not always possible to present or provide exact symmetries.

### To be discussed

Functions | Array Functions | Map Functions
--- | --- | ---
`fn:contains-subsequence`<br/>*new*: #94, #844 | `array:contains-subarray` ? | `map:contains`
`fn:ends-with-subsequence`<br/>*new*: #96, #844 | `array:ends-with-subarray` ?  | –
`fn:starts-with-subsequence`<br/>*new*: #96, #844 | `array:starts-with-subarray` ? | –
`fn:distinct-values` | `array:distinct-members` ? | –
`fn:duplicate-values`<br/>*new*: #123 | `array:duplicate-members` ?  | –
`fn:empty` | `array:empty` *new*: #229 | `map:empty` ? #827
`fn:exists` | `array:exists` *new*: #229 | `map:exists` ? #827
`fn:every` *new* | `array:every` ? | `map:every` ?
`fn:some` *new* | `array:some` ? | `map:some` ?
`fn:highest` *new* | `array:highest` ? | –
`fn:lowest` *new* | `array:lowest` ? | –
`fn:index-of` | `array:index-of` ? #260 | –
`fn:index-where` *new* | `array:index-where`<br/>*new*: #114 | `map:keys($m, $pred)`<br/>*new*: #467
`fn:items-at` *new*: #213<br/>→ `fn:get` ? #872 | `array:members-at` ? #825<br/>`array:get` | `map:get`
`fn:intersperse` *new*: #2<br/>→ `fn:join` ? #868 | `array:join`  | –
`fn:subsequence-where`<br/>*new*: #878 | `array:subarray-where` ? | –
`fn:substitute` ? #553, #583 | `array:replace` *new*;<br/>`array:substitute` ? #583 | `map:replace` *new*;<br/>`map:substitute` ? #583
`fn:slice` *new* | `array:slice` *new*  | –
– | `array:split` *new* | `map:entries` *new*
– | `array:values` *new* | `map:keys`; `map:values` *new*
[...]

PR #880, opened 2023-12-06
872 Symmetry: fn:items-at → fn:get

### LINKED ISSUE #872: Symmetry: fn:items-at → fn:get
I think that `fn:items-at` should be changed to `fn:get`:

* In #843, we try to harmonize the function names across sequences, maps, and arrays. We have `array:get` and `map:get` to retrieve single entries of the input, but we have `fn:items-at` for sequences.
* `fn:items-at` allows you to supply more than a single position, but `items-at($seq, (1, 3, 2)` can easily be rewritten to `(1, 3, 2) ! get($seq, .)` – similar to `(1, 3, 2) ! array:get($array, .)` and `(1, 3, 2) ! map:get($map, .)`.
* With #844, `fn:items-at` would be the only function left with `items` in its name.

The function signature would be as simple as:

```xquery
fn:get(
  $input  as item()*,
  $at     as xs:integer	
) as item()
```

Obviously, most people will still use `$input[$at]` – but the same applies to arrays and maps (and other functions like `fn:head`). One of the advantages of `fn:get` is that you can pass on the context item as position argument.

ChristianGruen:
Closes #872.


ChristianGruen:
Instead of this PR, I think we should put #843 on the weekly agenda first.

ChristianGruen:
I’ve decided to drop this PR. It would have been easy to resolve the conflict, but #843 is more important.

PR #901, opened 2023-12-13
895 Parameters with default values: allow empty sequences

### LINKED ISSUE #895: Parameters with default values: allow empty sequences
We need a consistent approach for defining types of optional function arguments. In most current cases, if a function argument is supplied, it must be non-empty:

```xquery
map:get(
  $map       as map(*),
  $key       as xs:anyAtomicType,
  $fallback  as function(xs:anyAtomicType) as item()*  := void#1
) as item()*

fn:starts-with-sequence(
  $input        as item()*,
  $subsequence	as item()*,
  $compare      as function(item(), item()) as xs:boolean  := fn:deep-equal#2
) as xs:boolean
```

In some cases, it’s optional:

```xquery
fn:replace(
  $value        as xs:string?,
  $pattern      as xs:string,
  $replacement  as xs:string?                                                   := (),
  $flags        as xs:string?                                                   := '',
  $action       as (function(xs:untypedAtomic, xs:untypedAtomic*) as item()?)?  := ()
) as xs:string)

(: #874 :)
fn:subsequence(
  $input   as item()*,
  $start   as xs:double?                                     := (),
  $length  as xs:double?                                     := (),
  $from    as (function(item(), xs:integer) as xs:boolean)?  := (),
  $while   as (function(item(), xs:integer) as xs:boolean)?  := (),
  $until   as (function(item(), xs:integer) as xs:boolean)?  := ()
) as item()*
```

As a result, `map:get($map, fallback := ())` is invalid, while `replace($string, $pattern, action := ())`  would be valid.

I think it’s better to enforce non-empty arguments (provided that a single item is expected).

ChristianGruen:
Closes #895

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -1341,7 +1341,7 @@
       <fos:signatures>
          <fos:proto name="round" return-type="xs:numeric?">
             <fos:arg name="value" type="xs:numeric?"/>
-            <fos:arg name="precision" type="xs:integer" default="0"/>
+            <fos:arg name="precision" type="xs:integer?" default="0"/>
          </fos:proto>
       </fos:signatures>
       <fos:properties>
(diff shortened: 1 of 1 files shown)


ndw:

PR #907, opened 2023-12-16
906 fn:deep-equal: unordered → ordered

ChristianGruen:


michaelhkay (changes requested):
This sentence seems all wrong: "If the option is set to false, the function returns
            false if for every item in $input1 there is a deep-equal item
            in $input2 that has not been matched to another item."

I suspect it would be easiest to say: if the option is set to false, the function returns true  if and only if every item in the sequence $input1 is
            deep-equal to an item at some position in the sequence $input2, and vice versa.

ChristianGruen:
Revised as suggested

ndw:

Issue #877, opened 2023-12-06
Inconsistency in XQFO comparator functions/operators with recursive rules


Arithmeticus:
The rules for [op:hexBinary-less-than()](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-hexBinary-less-than) appear to define a recursive octet-by-octet operation, but I think it flounders in rule 3, where it does not ask for rule 2 to be applied seriatim to each octet pair, but asks for an en masse comparison of two octet sequences.

Compare to 5.3.2, [Unicode Codepoint Collation](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#codepoint-collation), which describes a similar recursive item-for-item comparison. Interesting formal differences (e.g., unordered list versus ordered list).

[fn:deep-equal()](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-deep-equal) is similar, but it is also much more complex. Nevertheless, the way it breaks down the problem at the outset, to dispense immediately with the recursive factor, and deal simply with the rules for equality, is IMO admirable.

It would be nice if there were a bit more consistency in the prose and presentation of recursive rules. Do others agree, and are there other functions/operations that should be considered in this question? I'm thinking immediately only of comparator functions/operations, not functions that use recursion to filter or create. (There may be parallels, but let's start with those functions that are most similar.)

michaelhkay:
By all means try to improve matters - it's not something I want to invest a great deal of time in. There's always a risk of introducing bugs unintentionally.

michaelhkay:
I believe this issue can be closed by virtue of PR #936.

ndw:
At [meeting 109](https://qt4cg.org/meeting/minutes/2025/02-11.html) the CG agreed to close this issue without further action.

PR #936, opened 2024-01-10
877 revised rules for op:binary-less-than

Arithmeticus:
Rule 3 for `op:binary-less-than` was a bit of a mess (see #877), and needed to be expressed as a recursive operation. 

My proposed revision depends of phraseology drawn from `fn:decode-from-uri`, `fn:deep-equal`, and 5.3.2 Unicode Codepoint Collation (here slightly adjusted from unordered to ordered list).


ndw:
The CG agreed to merge this PR without discussion at [meeting 062](https://qt4cg.org/meeting/minutes/2024/01-23.html#merge-without-discussion)

Issue #948, opened 2024-01-15
fn:scan-left and fn:scan-right - produce accumulation of results


dnovatchev:
# fn:scan-left and fn:scan-right - produce accumulation of results
In XPath 4.0 so far we still don't have a convenient way to express the functionality of producing a series of accumulated (accrued) results when applying a folding function over a collection (sequence, array, ...) of items. The general use-case for this is the task to produce a sequence of running totals when applying an operation over a sequence of data points: **produce the partial sums of loan payments** over fixed periods, **produce the compounded amounts  of a deposit** with fixed interest rate over years, ..., etc.

Two functions (shamelessly borrowed from Haskell):
 
- **fn:scan-left**
- **fn:scan-right**
------------
## fn:scan-left

This function has a similar signature to that of **_`fn:fold-left`_** and produces the same final result, however it produces the complete (ordered) sequence of all partial results from every new value the accumulator gets during the evaluation of **_`fn:fold-left`_**. 

### Signature
```xq
fn:scan-left($input	as item()*,	
             $zero	as item()*,	
             $action	as function(item()*, item()) as item()*	
) as array(*)*
```

### Properties
This function is [·deterministic·](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#dt-deterministic), [·context-independent·](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#dt-context-independent), and [·focus-independent·](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#dt-focus-independent).

### Rules

The function is equivalent to the following implementation in XPath(return clause added for completeness):

```xq
let $scan-left-inner := function($seq as item()*, 
                                 $zero as item(), 
                                 $fun as function(item()*, item()) as item()*,
                                 $self as function(*)
                               ) as array(*)*                               
{
  let $result := [$zero]
   return
     if(empty($seq)) then $result
       else
         (
           $result, $self(tail($seq), $fun($zero, head($seq)), $fun, $self)  
         )
},

$scan-left := function($seq as item()*, 
                       $zero as item(), 
                       $fun as function(item()*, item()) as item()*
                     ) as array(*)*  
{
  $scan-left-inner($seq, $zero, $fun, $scan-left-inner)
}  

  return
    $scan-left(1 to 10, 0, op('+'))
```

### Examples:

```xq
 $scan-left(1 to 10, 0, op('+'))
```
produces:

```xq
[0]
[1]
[3]
[6]
[10]
[15]
[21]
[...]

PR #957, opened 2024-01-18
948 Added fn:scan-left and fn:scan-right

dnovatchev:
As discussed at https://github.com/qt4cg/qtspecs/issues/948

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -29390,14 +29390,7 @@ $scan-left := function($seq as item()*,
                <fos:expression><eg>scan-left(1 to 3, 0, op('-'))</eg></fos:expression>
                <fos:result>[0], [-1], [-3], [-6]</fos:result>
             </fos:test>
-         </fos:example>  
-         <fos:example>
-            <p> Produce the factorials of all numbers from 0 to 10</p>
-            <fos:test>
-               <fos:expression><eg>scan-left(1 to 10, 1, op('*'))</eg></fos:expression>
-               <fos:result>[1], [1], [2], [6], [24], [120], [720], [5040], [40320], [362880], [3628800]</fos:result>
-            </fos:test>
-         </fos:example>           
+         </fos:example>      
          <fos:example>
             <p>Produce the intermediate results of mapping each number in a sequence to its doubled value. 
             This example shows the necessity to place each intermediate result (sequence) into a singleton array - otherwise
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -29314,6 +29314,157 @@ path with an explicit <code>file:</code> scheme.</p>
       </fos:history>
    </fos:function>
 
+   <fos:function name="scan-left" prefix="fn">
+       <fos:signatures>
+         <fos:proto name="scan-left" return-type="array(*)*">
+            <fos:arg name="input" type="item()*" usage ="navigation"/>           
+            <fos:arg name="zero" type="item()*"/>           
+            <fos:arg name="action" type="function(item()*, item()) as item()*" usage="inspection"/>   
+         </fos:proto>
+       </fos:signatures>
+      <fos:properties>
+         <fos:property>deterministic</fos:property>
+         <fos:property>context-independent</fos:property>
+         <fos:property>focus-independent</fos:property>
+      </fos:properties>
+      <fos:summary>
+        <p>Produces the complete (ordered) sequence of all partial results from every new value 
+        the accumulator is assigned to during the evaluation of fn:fold-left.</p>
+      </fos:summary>
+      <fos:rules>
+         <p>The function is equivalent to the following implementation in XPath(return clause added for completeness):</p>
+         <eg><![CDATA[
+let $scan-left-inner := function($seq as item()*, 
+                                 $zero as item(), 
+                                 $fun as function(item()*, item()) as item()*,
+                                 $self as function(*)
+                               ) as array(*)*                               
+{
+  let $result := [$zero]
+   return
+     if(empty($seq)) then $result
+       else
+         (
+           $result, $self(tail($seq), $fun($zero, head($seq)), $fun, $self)  
+         )
+},
+
+$scan-left := function($seq as item()*, 
+                       $zero as item(), 
+                       $fun as function(item()*, item()) as item()*
+                     ) as array(*)*  
+{
+  $scan-left-inner($seq, $zero, $fun, $scan-left-inner)
+}  
+
+  return
+    $scan-left(1 to 10, 0, op('+'))         
+]]></eg>         
+      </fos:rules>
+      <fos:errors>
+         <p>As a consequence of the function signature and the function calling rules, a type error
+            occurs if the supplied function <code>$action</code> cannot be applied to two arguments, where
+            the first argument is either the value of <code>$zero</code> or the result of a previous
+            application of <code>$action</code>, and the second 
+            is any single item from the sequence <code>$input</code>.</p>      
+      </fos:errors>
+      <fos:examples>
+         <fos:example>
+            <fos:test>
+               <fos:expression><eg>scan-left(1 to 10, 0, op('+'))</eg></fos:expression>
+               <fos:result>[0], [1], [3], [6], [10], [15], [21], [28], [36], [45], [55]</fos:result>
+            </fos:test>
+         </fos:example>         
+         <fos:example>
+            <fos:test>
+               <fos:expression><eg>scan-left(1 to 3, 0, op('-'))</eg></fos:expression>
+               <fos:result>[0], [-1], [-3], [-6]</fos:result>
+            </fos:test>
+         </fos:example>  
+         <fos:example>
+            <p> Produce the factorials of all numbers from 0 to 10</p>
+            <fos:test>
+               <fos:expression><eg>scan-left(1 to 10, 1, op('*'))</eg></fos:expression>
+               <fos:result>[1], [1], [2], [6], [24], [120], [720], [5040], [40320], [362880], [3628800]</fos:result>
+            </fos:test>
+         </fos:example>           
+      </fos:examples>          
+      <fos:history>
+         <fos:version version="4.0">Proposed for 4.0</fos:version>
+      </fos:history>
+   </fos:function>
+
+   <fos:function name="scan-right" prefix="fn">
+       <fos:signatures>
+         <fos:proto name="scan-right" return-type="array(*)*">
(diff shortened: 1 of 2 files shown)


ChristianGruen:
Thanks; looks good. Some feedback:

1. If we’re going to update two variants of the function, the spec should provide convincing arguments why `fn:scan-right` is needed (the mere existence of `fn:fold-right` might be a questionable reason, as hardly anyone uses it in practice).
2. If we keep `fn:scan-right`, the pseudocode can possibly be simplified to `array:reverse(scan-left(array:reverse($input), $zero, $action))`.
2. At least one example should be added that demonstrates why it’s helpful to return the intermediate results as arrays.
4. Existing functions such as `map:for-each` return the results as a flat sequence. A note could be useful to point out why this function behaves differently.
5. I would suggest aligning the pseudocode with the one from the fold functions, as it does a similar thing.
6. As the `$action` function of `fold` now contains a positional parameter, we should either add it to `scan` as well, or argue why the two solutions differ.

michaelhkay:
Similar comments to me. I originally expected the function to return an array of sequences rather than a sequence of arrays, but I don't think there's a strong argument either way.

dnovatchev:
> Thanks; looks good. Some feedback:
> 
Thanks for the comments!

> 1. If we’re going to update two variants of the function, the spec should provide convincing arguments why `fn:scan-right` is needed (the mere existence of `fn:fold-right` might be a questionable reason, as hardly anyone uses it in practice).

I will have to dismiss this as inappropriate for a group that has dedicated itself to the cause of diversity. People who knew better in the first place, included [**`fn:fold-right`**](https://www.w3.org/TR/xpath-functions-31/#func-fold-right) in the XPath 3 official/current specification. 

And the real problem is the implementors' lacking understanding of the goals of  [**`fn:fold-right`**](https://www.w3.org/TR/xpath-functions-31/#func-fold-right). We have already discussed this thoroughly in this issue: **https://github.com/qt4cg/qtspecs/issues/670** .

If someone, instead of fixing their mistakes in implementing `fold-right`, wants to completely "eliminate the problem" by eliminating the function itself, this would be a classical example of discrimination and yielding to one's lack of competence.

So, let's do better here - we are not that "someone".

> 2. If we keep `fn:scan-right`, the pseudocode can possibly be simplified to `array:reverse(scan-left(array:reverse($input), $zero, $action))`.

There are more than one ways of expressing the semantics. Let's not waste time in trying to find "the best one". I prefer the current code as it relies only on using the very fundamental functions `fn:head` and `fn:tail`, which is more robust than being dependent on more complicated functions whose future behavior could more easily be changed. Also, the proposed  code change is quite sub-optimal as it requires to evaluate array:reverse **twice**, while the current code just performs a single linear scan of the input.

> 3. At least one example should be added that demonstrates why it’s helpful to return the intermediate results as arrays.

Yes, thank you for noticing this. I will add such example.

> 4. Existing functions such as `map:for-each` return the results as a flat sequence. A note could be useful to point out why this function behaves differently.

Yes, I will add this note, thanks!

> 5. I would suggest aligning the pseudocode with the one from the fold functions, as it does a similar thing.

I would suggest replacing the code for the fold functions which comes in two flavors - XSLT and XQuery, with just a single XPath implementation, thus eliminating the redundancy and helping shorten the spec and save a tree. Indeed, we can and should do this for all functions code implementations.

> 6. As the `$action` function of `fold` now contains a positional parameter, we should either add it to `scan` as well, or argue why the two solutions differ.

I am going to raise a separate issue for the change in the signature of fold-left and fold-right. The examples for these functions -- all but with one exception - are providing an ` $action` function that only accepts 2 (not 3) arguments. Yes, I know about the rule we introduced that when passing more than the wanted number of arguments to a function, the not-needed ones are ignored, but still passing 2 arguments to a 3-argument function is clearly an error. The prose of these functions needs to be changed - this probably is not worth the effort - it would be better to have 2 separate overloads for each of the fol functions, which eliminates this artificially-created problem.

ChristianGruen:
> I will have to dismiss this as inappropriate for a group that has dedicated itself to the cause of diversity. People who knew better in the first place, included [**`fn:fold-right`**](https://www.w3.org/TR/xpath-functions-31/#func-fold-right) in the XPath 3 official/current specification.

Improving the documentation, and making it accessible to more readers, should never be inappropriate.

> And the real problem is the implementors' lacking understanding of the goals of [**`fn:fold-right`**](https://www.w3.org/TR/xpath-functions-31/#func-fold-right). We have already discussed this thoroughly in this issue: **#670** .

Yes, there was discussion on that topic. Note that there are reasons why no implementation (or at least 5 of them; I’m not aware any other) have chosen a recursive implementation, and there are reasons why many other languages only provide single `fold` or `reduce` functions. As long as there will be no corresponding implementation, there is no benefit for a right-to-left scan.

In principle, I have no objection to adding `scan-right', but we definitely need to be able to justify it and explain to users why it makes sense.

> I would suggest replacing the code for the fold functions which comes in two flavors - XSLT and XQuery, with just a single XPath implementation, thus eliminating the redundancy and helping shorten the spec and save a tree. Indeed, we can and should do this for all functions code implementations.

You may have checked out an old version of the specification; the current one only includes XQuery pseudocode.

Maybe we can discuss in the meeting if the XQuery or XPath representations in the spec are better readable, and unify them in a second step.

dnovatchev:
> > I will have to dismiss this as inappropriate for a group that has dedicated itself to the cause of diversity. People who knew better in the first place, included [**`fn:fold-right`**](https://www.w3.org/TR/xpath-functions-31/#func-fold-right) in the XPath 3 official/current specification.
> 
> Improving the documentation, and making it accessible to more readers, should never be inappropriate.

Suggesting to remove a function that was officially published and part of the Spec for more than 10 years is not just "improving the documentation" Here is a fragment from your original comment:

>>> "**_the mere existence of fn:fold-right might be a questionable reason, as hardly anyone uses it in practice_**"

And about this:
> Yes, there was discussion on that topic. Note that there are reasons why no implementation (or at least 5 of them; I’m not aware any other) have chosen a recursive implementation, and there are reasons why many other languages only provide single `fold` or `reduce` functions.

Yes, there are reasons for everything that has happened - this doesn't justify the bad/tragic/obnoxious events like for example the Holocaust in which 6 million people were killed simply because they were Jewish ...



> Maybe we can discuss in the meeting if the XQuery or XPath representations in the spec are better readable, and unify them in a second step.

Instead of providing either an XQuery or an XSLT code we can do much better -- providing XPath code that is both valid XSLT and XQuery.

michaelhkay:
Could we please try to keep cool about this?

It's perfectly simple: there's a case for providing the function because it's present in other environments and it's symmetrical, and there's a case against because not many people need it. That's all the information we need in order to make a decision.

ChristianGruen:
> Suggesting to remove a function that was officially published and part of the Spec for more than 10 years is not just "improving the documentation"

Existing functions cannot be removed. Even if we assume that features are rarely used, this would break backward compatibility.

Maybe it was a misunderstanding. My intent was to say that we need to help users to understand why a particular function exists.

> Yes, there are reasons for everything that has ever happened - this doesn't justify the bad/tragic/obnoxious events like for example the Holocaust in which 6 million people were killed simply because they were Jewish ...

I doubt this will be a constructive discussion, so I'll leave it like that, and wait for our joint discussion.

dnovatchev:
1. One not-mentioned-yet use-case of both scan functions is as a tool to debug the execution of the respective (fold-left or fold-right) functions.

2. Another use case for scan-right is provided in https://www.cs.cmu.edu/~guyb/papers/Ble93.pdf - see the "up-sweep" at the top of page 42.

One shouldn't dismiss something based on the fact that "I don't see it" - for example  people believed that the Earth was the center of the Universe and everything was rotating around the Earth, ... until they got better telescopes and found that there were some objects that were rotating not around the earth but around other planets - such as the 4 moons of Jupiter: https://www2.jpl.nasa.gov/galileo/ganymede/discovery.html -- which of course wasn't a good reason enough to acknowledge Galileo's proven, factual truth 😢

dnovatchev:
> > I originally expected the function to return an array of sequences rather than a sequence of arrays, but I don't think there's a strong argument either way.
> 
> @michaelhkay A flat sequence would be my favorite solution. It makes things like `sum(scan-left($input, 0, op('+')))` easier. For structured results, it will be simple to wrap the intermediate functions into arrays – `scan-left($input, 1, fn { [.+.] })` – or to use the equivalent array functions, which I assume will be added in a second step.

A flat sequence is like a living 3D organism that has been smashed (and destroyed) to 2D toilet paper.

Arithmeticus:
Various questions/suggestions, in preparation for tomorrow's CG meeting. All of my comments come from comparing `fn:fold-left` to the proposed text for `fn:scan-left`.

The summary points to `fn:fold-left` as the basis of its definition, but the rules nowhere refer to that function. Inconsistent. Either remove any reference to `fn:fold-left` from the summary, or make `fn:fold-left` integral to the rules.

Summary: What does "new" mean? Text "the accumulator": what accumulator (nowhere referred to again in the rules)? Meaning of "assigned to"? Why is the prose not more closely aligned with the prose of `fn:fold-left`? (I think `fn:fold-left`'s summary is incomplete, lacking any clause indicating its output.)

Signature: Why is the fourth line not this: `$action as function(item()*, item()**, xs:integer**) as item()*` as in `fn:fold-left`? The related question might be why the integer is used in `fn:fold-left` (which may need adjustment).

Rules: 
* See above about the defined relationship of this function with `fn:fold-left`. 
* See above about the lack of `$pos as xs:integer`.
* If it is decided that `fn:scan-left` should remain defined without reference to `fn:fold-left`, the code should, nevertheless, be more closely aligned in syntax. E.g., `$seq` - > `$input`,  `$fun` - > `$action`. The fourth parameter `$self` adds a layer of indirection that is unnecessarily obfuscating; the code for `fn:fold-left`, where `fold-left-helper` refers to itself by name, is clearer. I realize that this a conscious choice, to define the function exclusively in terms of XPath, and not of a host language. But `fn:fold-left`'s presentation in XQuery more effectively communicates an abstract concept.
* I don't think the return clause should be included in the formal spec definition. Save it for the examples.

Notes: include a note that explicitly cross-references `fn:fold-left`. A reciprocal note should also be supplied for `fn:fold-left`.

Examples: begin with at least two examples that are already present in `fn:fold-left`, so readers can immediately understand the difference between the two functions. The first two examples are ideal for illustrating how `fn:scan-left` could be used to do what `fn:fold-left` cannot.

I think the question will be inevitably raised in CG discussion, why yet **another** pair of new functions? Why not extend the arity of `fn:fold-left/right` to include a parameter that specifies whether intermediate results should also be returned.

ndw:
The CG agreed to merge this PR at [meeting 063](https://qt4cg.org/meeting/minutes/2024/01-30.html)

Issue #946, opened 2024-01-12
fn:iterate-while → fn:while-do, fn:do-until


ChristianGruen:
First feedback shows that `fn:iterate-while` is helpful, but the name needs to be improved: It implies that the first iteration occurs before the invocation of the test… Which could sometimes be helpful, too.

I suggest renaming the function to `fn:while-do` (“do” is commonly used when while loops are specified), and adding `fn:do-until`.

PR #962, opened 2024-01-20
946 fn:iterate-while → fn:while-do, fn:do-until

ChristianGruen:
My first thought was to name the second function `fn:do-while`, but `fn:do-until` with an inversed predicate seemed more appropriate to me.

Issue: #946

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -18952,8 +18952,8 @@ return while-do(
                <fos:expression><eg><![CDATA[
 while-do(
   1 to 9,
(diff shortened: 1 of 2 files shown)


ndw:
The CG agreed to merge this PR at [meeting 062](https://qt4cg.org/meeting/minutes/2024/01-23.html)

dnovatchev:
If we want to preserve "while" for the language, then I am proposing this name for the function:

# whilest

A name that means the same as "while" and that even Shakespeare would like 😄

ChristianGruen:
> A name that means the same as "while" and that even Shakespeare would like 😄

No doubst ;·)

Issue #973, opened 2024-01-29
fn:parse-json, fn:json-to-xml: `number-parser`, `fallback`


ChristianGruen:
1. See #33: `number-parser` option needs to be added to `fn:json-to-xml`.
2. Similar to `fn:replace`, I would suggest using `function(xs:untypedAtomic) as item()?)` as signature for both the `fallback` and the `number-parser` option, and (for `fallback`) to invoke `fn:string#1` on the result. This way, explicit casts in the code get obsolete. Queries like the following one…

```xquery
parse-json(
  '-123',
  map { 'number-parser': fn($n) { $n => number() => abs() }
)
```

…can then be simplified to `parse-json('-123', abs#1)`.

PR #975, opened 2024-01-29
973 fn:parse-json, fn:json-to-xml: number-parser, fallback

ChristianGruen:
Issue: #973.

In addition, I fixed the description for the `fallback` option `fn:parse-json`, as it seemed incomplete to me: 

> The function is called when the JSON input contains a special character (as defined under the escape option) that is valid according to the JSON grammar, whether the special character is represented in the input directly or as an escape sequence.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -21969,7 +21969,7 @@ return $M(collation-key("a", $C))</eg></fos:expression>
                   option is present with the value <code>true</code>.
                </fos:meaning>
                <fos:type>(function(xs:untypedAtomic) as item()?)?</fos:type>
-               <fos:default>fn { char('#xFFFD') }</fos:default>
+               <fos:default>fn { char(0xFFFD) }</fos:default>
                <fos:values>
                   <fos:value value="User-supplied function"
                         >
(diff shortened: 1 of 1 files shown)


michaelhkay (commented):
Looks OK to me. For the substitute character, please avoid using a string literal containing entity references as that's not portable between XPath and XQuery. Instead use `char('...')`.

ChristianGruen:
> Looks OK to me. For the substitute character, please avoid using a string literal containing entity references as that's not portable between XPath and XQuery. Instead use `char('...')`.

@michaelhkay I’ll be glad to change it. To be sure, do you refer to `&#xFFFD;"`, i.e., to the string literal that I’ve adopted from the existing version of the spec (https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-parse-json)?

ChristianGruen:
I’ve replaced the two occurrences of `"&amp;#xFFFD;"` with `char('#xFFFD')`.

ChristianGruen:
@ndw You can merge this one as well. I’ve updated `char(0xFFFD)`, but I’ll do a separate PR for the remaining action item tasks.

ndw:
The CG agreed to merge this issue at [meeting 064](https://qt4cg.org/meeting/minutes/2024/02-06.html)

PR #977, opened 2024-01-30
Ignore this, it's just a test

ndw:
I'm trying to work out why sometimes PR succeeds when it contains markup errors...

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -29630,13 +29630,6 @@ return
   $scan-right(1 to 10, 0, op('+'))          
 ]]></eg>         
       </fos:rules>
-      <fos:errors>
-         <p>As a consequence of the function signature and the function calling rules, a type error
-            occurs if the supplied function <code>$action</code> cannot be applied to two arguments, where
-            the first argument is any item in the sequence <code>$input</code>, and the second is either
-            the value of <code>$zero</code> or the result of a previous application of
-            <code>$action</code>.</p>
-      </fos:errors>
         <fos:notes>
                 <olist>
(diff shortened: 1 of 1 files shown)


ndw:
Right. So the challenge here is that this PR passes, even though one of the files is not schema-valid because PRs don't have access to a Saxon EE licensed processor. :-(

ChristianGruen:
> Right. So the challenge here is that this PR passes, even though one of the files is not schema-valid because PRs don't have access to a Saxon EE licensed processor. :-(

Let’s hope there’ll be someone who can get hold of such licenses ;·) Well, I guess the problem is more technical.

ndw:
Yes. The problem is that a PR doesn't usually have access to the CI "secrets" that include things like access tokens and, in this case, the Saxon EE license. That's because, in principle, a user could exfiltrate those secrets with a sufficiently devious PR.

In practice, we ignore build script and stylesheet changes in PRs, so I think it would be safe(ish) to allow PRs in this repository to have access to the secrets.  And that's probably what I'll do. After I check that the Saxon EE license we're using here has reasonable metadata and an expiration date!

michaelhkay:
We have a mechanism that allows a compiled schema (SCM file) to contain an embedded key that enables validation against the schema to be done without a Saxon license, provided a master key was used when compiling the schema. Perhaps that would work here.

Michael Kay

ndw:
I don't think there's an adequately secure way to do this. Bah, humbug.

It means we'll get occasional PRs that appear to pass bug cause failures when merged. I'll just deal with them when it happens until I'm sufficiently more frustrated to try to fix this some other way. (The obvious answer occurred to me this morning: convert the XSD files into RELAX NG and validate with those, but not today.)

PR #937, opened 2024-01-10
779 hash function

Arithmeticus:
First draft of hash function, proposed in #779.

Error message left as to-do item; guidance from editors appreciated.

I opted to leave out wrapper/cryptographic functionality, such as salting, and to demonstrate via example how it could be done by a developer on their own. In my opinion what we need here is a simple atomic function that can be incorporated into other molecular functions.

I may tinker with the prose description up to CG discussion, so comments are welcome.

Issue #779, opened 2023-10-26
Hash/checksum function


Arithmeticus:
I propose a new function for the core XPath functions, here called `fn:hash()` for the sake of discussion. The goal is to give XPath users access to CRC, checksum, and cryptographic hash functions.

### Rationale

Simple checksums functions, such as those from the Fletcher family, are relatively easy to write in a host language such as XSLT. More complex ones are far more challenging to write, and may incur serious performance penalties. For example, from the TAN function library, see the [MD5 checksum/hash functions](https://github.com/textalign/TAN-2021/blob/master/functions/checksums/TAN-fn-hash-and-check.xsl). (Yes, one day I thought it would be fun to try to implement the MD5 algorithm in XSLT.) Most programming languages in which an implementation is written have access to cryptographic libraries that are highly performative.

Hash functions are widely used, and certainly important in XML-based workflows, whether as filenames, database fields, etc.

The closest comparable existing functions is `generate-id()`, but this was designed as an identifier for nodes.

In short, I believe there is a significant need that outstrips current functionality. 

In the draft below, I have adopted only the MD-5 algorithm as a core requirement, to catalyze discussion. I have assumed the user wants the string form of the output, not the raw bits. I have not tried to flesh out prose that would warn users away from security complacency. 

For discussion, [here is a list of relevant algorithms](https://en.wikipedia.org/wiki/List_of_hash_functions). I look forward to community feedback.

#### Summary
This function takes as input a string or octet sequence and returns a string representation of the results from a specified hash, checksum, or cyclic redundancy check function.

#### Signature
```
fn:hash(
   $value as union(xs:string, xs:hexBinary, 
                   xs:base64Binary)?   := fn:string(.),
   $algorithm as xs:string             := "md5"
) as xs:string?
```

#### Properties
The zero-argument form of this function is deterministic, context-dependent, and focus-dependent.

The one- and two-argument form of this function is deterministic, context-independent, and focus-independent.

#### Rules
If the zero-argument version of the function is used, the result is the same as calling the one-argument version, with `$value` set to `fn:string(.)`.

If the one-argument version of the function is used, the result is the same as calling the two-argument version, with `$algorithm` set to "md5".

The effective value of `$algorithm` is the value of the expression `fn:lower-case(fn:replace($function, '\W+', ''))`.

If `$value` is the empty sequence, or a string of zero length, the function returns the empty sequence.

[...]

PR #995, opened 2024-02-03
937 revised in light of CG feedback

Arithmeticus:
This revises #937 (catalyzed by #779) in light CG discussion that approved the PR.

* Output is now `xs:hexBinary?`
* Second parameter `$algorithm` replaced with an option map. In the specs I avoided `fos:values/fos:value`, because this would disallow for case/space normalization, and it would effectively disallow any implement-defined algorithms not on the list of three algorithms, and trigger the dynamic error described in rule 6 in the [option parameter conventions](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#option-parameter-conventions).
* Options map has only one option. It doesn't make sense to provide an option changing the kind of output.
* I think this is the first example of an options map where the `fos:meaning` has rich text (paragraphs, unordered lists). It builds and renders fine locally.
* Extra note on the output format.
* Examples are expressed as chained functions, to illustrate how to get the customary string values.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -4807,10 +4807,15 @@ return normalize-unicode(concat($v1, $v2))</eg>
       </fos:examples>
    </fos:function>
    <fos:function name="hash" prefix="fn" at="2024-01-10" diff="add">
+      <!-- output: hexbinary to string cast for output -->
+      <!-- output: hexbinary preferred -->
+      <!-- provide options map: input, output -->
+      
       <fos:signatures>
-         <fos:proto name="hash" return-type="xs:string">
+         <fos:proto name="hash" return-type="xs:hexBinary?">
             <fos:arg name="value" type="union(xs:string, xs:hexBinary, xs:base64Binary)?"/>
-            <fos:arg name="algorithm" type="xs:string?" default="'md5'"/>
+            <fos:arg name="options" type="map(*)?" default="map{}"/>
+            <!--<fos:arg name="algorithm" type="xs:string?" default="'MD5'"/>-->
          </fos:proto>
       </fos:signatures>
       <fos:properties arity="1">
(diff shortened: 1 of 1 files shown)


ChristianGruen:
Thanks for the revision. I didn't have time for a full review yet, but I wonder if lower-case is consistently the more “customary” one. Can we find a reliable reference for this?

Personally, I’ve observed both variants in the past, and the representation usually depended on how binaries were converted to strings by the responsible language or tool.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -4828,42 +4829,50 @@ return normalize-unicode(concat($v1, $v2))</eg>
...
+                           hash functions with implementation-defined semantics.</p>
+               </fos:meaning>
+               <fos:type>xs:string</fos:type>
+               <fos:default>MD5</fos:default>
+            </fos:option>
+         </fos:options>
       </fos:rules>
       <fos:errors>
-         <p>A dynamic error is raised ([TODO: error code]) if the effective value of
+         <p>A dynamic error is raised ([TODO: error code]) if the effective value of the option
```
ChristianGruen, line 4875: (next line: `<code>$algorithm</code>` → `<code>algorithm</code>`.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -4828,42 +4829,50 @@ return normalize-unicode(concat($v1, $v2))</eg>
...
                <code>$algorithm</code> is not one of the values supported by the implementation.</p>
```
ChristianGruen, line 4876:
```suggestion
               <code>algorithm</code> is not one of the values supported by the implementation.</p>
```

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -4873,42 +4882,45 @@ return normalize-unicode(concat($v1, $v2))</eg>
...
       <fos:examples>
          <fos:variable name="doc" id="v-hash-doc"><![CDATA[<doc>abc</doc>]]></fos:variable>
          <fos:variable name="salt" id="v-hash-salt" select="&quot;D;%yL9TS:5PalS/d&quot;"/>
          <fos:example>
             <fos:test>
-               <fos:expression>hash("abc")</fos:expression>
-               <fos:result>"900150983cd24fb0d6963f7d28e17f72"</fos:result>
+               <fos:expression>hash("abc") => string()</fos:expression>
+               <fos:result>"900150983CD24FB0D6963F7D28E17F72"</fos:result>
             </fos:test>
```
ChristianGruen, line 4896: The presentation of the examples…



…could be revised by…

* using newlines for each arrow operator/operand, or
* using the “vertical layout”, i.e., by inserting `</fos:example><fos:example>` after each `</fos:test>`. See for example `fn:concat`:

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -4873,42 +4882,45 @@ return normalize-unicode(concat($v1, $v2))</eg>
...
                <fos:result>"3c01bdbb26f358bab27f267924aa2c9a03fcfdb8"</fos:result>
             </fos:test>
             <fos:test>
-               <fos:expression>hash("ABC", "sha-256")</fos:expression>
+               <fos:expression>hash("ABC", map{"algorithm": "sha-256")} => string() => lower-case()</fos:expression>
                <fos:result>"b5d4045c3f466fa91fe2cc6abe79232a1a57cdf104f7a26e716e0a1e2789df78"</fos:result>
             </fos:test>
             <fos:test use="v-hash-doc">
-               <fos:expression>hash($doc)</fos:expression>
+               <fos:expression>hash($doc) => string() => lower-case()</fos:expression>
```
ChristianGruen, line 4914: One `lower-case()` conversion example may suffice.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -4873,42 +4882,45 @@ return normalize-unicode(concat($v1, $v2))</eg>
             relevant.</p>
          <p>Additional security practices, such as salting, may be applied as a preprocessing step,
             or <code>fn:hash()</code> can be incorporated into more complex functions.</p>
+         <p>In most cases, the <code>xs:hexBinary</code> output of the function will be sought in
+            string form. Because of serialization rules, casting to a string renders the hash in
+            uppercase, and the more customary lowercase requires further adjustment.</p>
```
ChristianGruen, line 4887: See my previous comment: A reference reg. lowercase could be helpful.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -4828,42 +4829,50 @@ return normalize-unicode(concat($v1, $v2))</eg>
...
-                  <bibref ref="fips180-4"/>.</p>
-            </item>
-         </olist>
-         <p>Conforming implementations <rfc2119>may</rfc2119> support other checksum and
-                  hash functions with implementation-defined semantics.</p>
+         <p>The function returns as <code>xs:hexBinary</code> the octets returned by passing
+               <code>$value</code> as an octet sequence through the hash, checksum, or cyclical
+            redundancy check function specified by the options map. The process is followed even if
+            the input octet sequence is empty. </p>
+         <p>The entries that may appear in the <code>$options</code> map are as follows. The detailed rules
```
ChristianGruen, line 4847: If I remember correctly, we’ve chosen to move the `algorithm` into a map, as a map may contain other entries that could be utilized by a proprietary algorithm. This should probably be mentioned; otherwise, one may wonder why a map needs to be specified if there’s just one option defined.

@ndw Do you have an example in mind for a custom hash algorithm with additional parameters?

ChristianGruen (changes requested):

specifications/xpath-functions-40/src/function-catalog.xml
Arithmeticus: @ChristianGruen I had hoped that the cross-reference to option parameter conventions would suffice, and not require mentioning that the map can have implementation-defined entries (already stated at the OPCs).

specifications/xpath-functions-40/src/function-catalog.xml
Arithmeticus: I've tried the `fn:concat` example, but it doesn't rendered the same, I think because of the use of variables within the examples. Thanks for the tip that the space in `fos:expression` is preserved in the specs.

Arithmeticus:
Thanks for the pointers, mostly addressed in new commit.

ChristianGruen (approved):

ndw:
The CG agreed to merge this issue without discussion at [meeting 064](https://qt4cg.org/meeting/minutes/2024/02-06.html)

Issue #150, opened 2022-09-21
fn:ranks: Produce all ranks in applying a function on the items of a sequence


dnovatchev:
We all know the value and usefulness of functions such as [**_`fn:highest()`_**](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-highest) [**_`and fn:lowest()`_**](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-lowest).

Sometimes, when we need to see all rankings of a particular sorting result, for example the rankings of a sport competition, we realize that highest and lowest are just the highest and lowest ranking-groups from all the rankings.

We define these three overloads for **`fn:ranks`**:

**fn:ranks**($input as item()*) as array(item()\*)\*

**fn:ranks**($input as item()*, $collation as xs:string?) as array(item()\*)\*

**fn:ranks**($input	 as item()\*, $collation	 as xs:string?, 	$key	 as function(item()) as xs:anyAtomicType*) as array(item()\*)\*

The rules and semantics for the arguments are the same as those for **fn:highest**, **fn:lowest**, **fn:sort**. What is different is just the result.

**Here is one possible XPath implementation and also a complete example**:

```xq
let $ranks := function(
                $input as item()*,
                $collation as xs:string?,
                $key as function(item()) as xs:anyAtomicType*) as array(item()*)*
 {
    for $v in sort(distinct-values($input ! $key(.)),  $collation)
     return [$input[$key(.) eq $v]]
 },
 
   $inp := (3, 2, 4),
   $keyfun := function($n) {$n mod 2},
   $theRanks :=  $ranks($inp, (), $keyfun),
   $theHighest := $theRanks[last()], 
   $theLowest := $theRanks[1]
 return
 
   ( "Ranks:", $theRanks,
     "=================",
     "Highest:",
     $theHighest,
     "=================",
     "Lowest:",
     $theLowest
   )
```

The result, as intended is all the rankings (in this case they are just 2 groups of equally-ranked items), then the highest and lowest, extracted as the last and first of the rankings:

```
Ranks:
[(2, 4)]
[3]
=================
Highest:
[3]
=================
Lowest:
[(2, 4)]
```

michaelhkay:
It would be useful to see some use cases for this function.

This seems to be essentially a combination of group-by() and sort(). It's essentially equivalent, as far as I can see, to

```
<xsl:for-each-group select="$input" group-by="$key(.)" collation="{$collation}">
[...]

PR #1027, opened 2024-02-17
150 fn:ranks

dnovatchev:
As proposed and discussed  here: https://github.com/qt4cg/qtspecs/issues/150


michaelhkay:
Looking at the first example, why do we want to return `[(2, 4)], [3]` rather than `[2, 4], [3]`. Generally I would have thought an array with two singleton members was more useful than an array with one member being a sequence of two items.

To make this change `return [$input[$key(.) eq $v]]` should be `return array{$input[$key(.) eq $v]}`.

In all of the examples given, the supplied key function returns a single item. But it is allowed (according to the signature) to return any sequence of atomic values. I don't think I understand the intended behaviour when it returns multiple items. The predicate `$key(.) eq $v` requires `$key(.)` to return zero or one items.

What is the effect of NaN values?

The supplied $collation is used when sorting the values, but not when deciding whether they are distinct. Is that right?

Specification style: this is the subject of a separate issue. We should either provide an expression that can act as an implementation of the function being specified, or we should provide a user-written function that has the same effect. In this case I think providing an expression will do the job. Alternatively, wouldn't an XQuery expression using group by and order by be clearer (perhaps not, since FLWOR expressions cannot have a dynamic collation).

Summary: I would suggest: _Sorts a supplied sequence based on the value of a sort key function, grouping the results so that items with the same key appear together as members of the same array._

dnovatchev:
> Looking at the first example, why do we want to return `[(2, 4)], [3]` rather than `[2, 4], [3]`. Generally I would have thought an array with two singleton members was more useful than an array with one member being a sequence of two items.
> 
> To make this change `return [$input[$key(.) eq $v]]` should be `return array{$input[$key(.) eq $v]}`.
> 
> In all of the examples given, the supplied key function returns a single item. But it is allowed (according to the signature) to return any sequence of atomic values. I don't think I understand the intended behaviour when it returns multiple items. The predicate `$key(.) eq $v` requires `$key(.)` to return zero or one items.
> 
> What is the effect of NaN values?
> 
> The supplied $collation is used when sorting the values, but not when deciding whether they are distinct. Is that right?
> 
> Specification style: this is the subject of a separate issue. We should either provide an expression that can act as an implementation of the function being specified, or we should provide a user-written function that has the same effect. In this case I think providing an expression will do the job. Alternatively, wouldn't an XQuery expression using group by and order by be clearer (perhaps not, since FLWOR expressions cannot have a dynamic collation).
> 
> Summary: I would suggest: _Sorts a supplied sequence based on the value of a sort key function, grouping the results so that items with the same key appear together as members of the same array._

@michaelhkay Thank you for these observations.

I am studying them and will respond.

dnovatchev:
> Looking at the first example, why do we want to return `[(2, 4)], [3]` rather than `[2, 4], [3]`. Generally I would have thought an array with two singleton members was more useful than an array with one member being a sequence of two items.
> 
> To make this change `return [$input[$key(.) eq $v]]` should be `return array{$input[$key(.) eq $v]}`.
> 

A good observation. Probably I wanted the key() function to be most general, but it feels difficult to find an immediate and compelling example.

> In all of the examples given, the supplied key function returns a single item. But it is allowed (according to the signature) to return any sequence of atomic values. I don't think I understand the intended behaviour when it returns multiple items. The predicate `$key(.) eq $v` requires `$key(.)` to return zero or one items.

Yes, then we would need a function such as `deep-equal`

ChristianGruen:
> A good observation. Probably I wanted the key() function to be most general, but it feels difficult to find an immediate and compelling example.

Off-topic, but maybe we can ask the same question for the scan functions: Wouldn’t singleton members be more intuitive?

dnovatchev:
> What is the effect of NaN values?
> 

Aren't `NaN` values supposed to be smaller than anything else? The answer should be: "_The effect is the same as when sorting_."

> The supplied $collation is used when sorting the values, but not when deciding whether they are distinct. Is that right?

The function `distinct-values`, used in the sample implementation, can be passed a collation, too. Not sure if the collation in the signature of `fn:ranks` should be used both for sorting and getting the distinct values from the input-sequence, or (if this at all is so important), we could have two different collations as parameters.

I prefer this to be as simple as possible. The `$colation` parameter was intended only because fn:sort needs one, not for producing the distinct values.

dnovatchev:
> Specification style: this is the subject of a separate issue. We should either provide an expression that can act as an implementation of the function being specified, or we should provide a user-written function that has the same effect. In this case I think providing an expression will do the job. Alternatively, wouldn't an XQuery expression using group by and order by be clearer (perhaps not, since FLWOR expressions cannot have a dynamic collation).
> 

A function definition **is** an expression, isn't it?

> Summary: I would suggest: _Sorts a supplied sequence based on the value of a sort key function, grouping the results so that items with the same key appear together as members of the same array._

A good one, thanks.

I will incorporate these suggestions now.

michaelhkay:
Thanks for responding to my comments.

I find it hard to believe that multiple collations are needed here; on the contrary, the way that sort keys are compared using distinct-values needs to be consistent with the way they are compared using sort. I'm also worried that there's a third comparison being done using `eq`, which uses the default collation rather than the supplied collation. I think this is also why I was uneasy about NaN - there are three different comparisons here which all potentially treat NaN differently.

I would like to suggest an alternative approach.

1. Make the signature compatible with fn:sort except that it returns `array(item())*`.

2. Take the rules of fn:sort as currently written, and modify them as needed to define `fn:ranks`: specifically, items whose sort keys all compare equal (rule 4) go into the same array in the output.

3. Change the definition of `fn:sort` so that `fn:sort($input, $collations, $keys, $orders)` returns `fn:ranks($input, $collations, $keys, $orders)?*`.

The changes needed to the fn:sort rules might be primarily, under "The result of the function is obtained as follows:" change rules 1, 3, and 4 as follows:

1. The result is a sequence of arrays S such that S?* contains the same items as the input sequence $input, but generally in a different order.

3. When a pair of corresponding sort key values of $A and $B are found to be not equal, then $A and $B appear in different arrays in the result sequence, and the array containing $A precedes the array containing $B in the result sequence if both the following conditions are true, or if both conditions are false:
* The sort key value for $A is less than the sort key value for $B, as defined below.
* The order direction in the corresponding sort key definition is "ascending".

4. If all the sort key values for $A and $B are pairwise equal, then $A and $B appear in the same array in the result sequence, and $A precedes $B in this array if and only if $A precedes $B in the input sequence.
Note:
That is, the sort is stable.

dnovatchev:
> I would like to suggest an alternative approach.
> 
> 1. Make the signature compatible with fn:sort except that it returns `array(item())*`.
> 2. Take the rules of fn:sort as currently written, and modify them as described below to define `fn:ranks`
> 3. Change the definition of `fn:sort` so that `fn:sort($input, $collations, $keys, $orders)` returns `fn:ranks($input, $collations, $keys, $orders)?*`. So we define fn:sort in terms of fn:ranks, not the other way round.
> 
> The changes needed to the fn:sort rules might be primarily, under "The result of the function is obtained as follows:" change rules 1, 3, and 4 as follows:

Thank you, @michaelhkay ,

I understand exactly what you are proposing, and yes, this is possible, however it becomes overly (and is that necessary?) complicated.

In particular, I never wanted to have a sequence of key-functions, and it seems that just one function can internally perform multiple comparisons, if that is necessary at all.

Also, by definition, `fn:ranks` is defined (to be meaningful)  over a set of (distinct) items -- while `fn:sort` returns all the input items, even in the case when they are not distinct.

As for comparing NaN values, can't we just say that NaN is less than any other item and for the purposes of this function NaN is equal to NaN? Thus no additional collation for treating NaN  would be necessary.

michaelhkay:
>however it becomes overly (and is that necessary?) complicated.

I think there's a lot of complexity in the current proposed spec, which compares values in three different ways: For example, sort and distinct-values treat two NaNs as equal, while `eq` treats them as not-equal. The proposal to define fn:sort in terms of fn:ranks is certainly a significant refactoring that may be difficult to get right, but if successful it will reduce complexity overall. (It might also be possible to define other functions such as distinct-values, duplicates, min, max, highest and lowest by reference to fn:ranks, and that would certainly be a great reduction in complexity if it can be achieved). But I agree it might be over-ambitious.

Another point, I just spotted the error condition "If the set of computed keys contains xs:untypedAtomic values that are not castable to xs:double then [the] operation will fail with a dynamic error." Why is that? All three comparisons that are used in the specification (`sort()`, `distinct-values()` and 'eq') treat `untypedAtomic` values as strings; I can't see where untypedAtomic-to-double conversion occurs.

dnovatchev:
> Another point, I just spotted the error condition "If the set of computed keys contains xs:untypedAtomic values that are not castable to xs:double then [the] operation will fail with a dynamic error." Why is that? All three comparisons that are used in the specification (`sort()`, `distinct-values()` and 'eq') treat `untypedAtomic` values as strings; I can't see where untypedAtomic-to-double conversion occurs.

This was copied from the already approved new functions: **[`fn:lowest`](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-lowest)** and **[`fn:highest`](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-highest)** .

michaelhkay:
>This was copied from the already approved new functions: fn:lowest and fn:highest

Those functions chose to follow `min()` and `max()` by treating untyped atomic values as numeric, rather than following sort which treats them as strings.

ChristianGruen:
@dnovatchev Minor observations:

* The type of `$key` is `function(item()) as xs:anyAtomicType*`. It could be helpful to add an example where the key function returns more than a single item.
* I think one `$collation` argument would be sufficient.

And editorial notes:

* A full stop is missing before `More formally,`.
* I believe `result-array` should be `result array` (but native speakers may know better).
* `key-value(s)` → `key-value combination`? (twice)
* `in increasing sort-order` → `in ascending order`
* The quotes in the text may need to be tweaked (`"` → `“` and `”`).

dnovatchev:
> @dnovatchev Minor observations:
> 
> * The type of `$key` is `function(item()) as xs:anyAtomicType*`. It could be helpful to add an example where the key function returns more than a single item.
> * I think one `$collation` argument would be sufficient.
> 
> And editorial notes:
> 
> * A full stop is missing before `More formally,`.
> * I believe `result-array` should be `result array` (but native speakers may know better).
> * `key-value(s)` → `key-value combination`? (twice)
> * `in increasing sort-order` → `in ascending order`
> * The quotes in the text may need to be tweaked (`"` → `“` and `”`).

Thanks @ChristianGruen  for these observations.

> * The type of `$key` is `function(item()) as xs:anyAtomicType*`. 

Yes,

@michaelhkay At present this is an issue with the proposed formal semantics:

```xq
let $ranks := function(
                $input as item()*,
                $key as function(item()) as xs:anyAtomicType*,
                $collation-input as xs:string?,
                $collation-keys as xs:string?
                ) as array(item())*,
                
 {
    for $v in sort(distinct-values($input ! $key(.), $collation-input),  $collation-keys)
     return array{$input[compare($key(.), $v, $collation-keys) eq 0]}
 }
```

As we see, the `compare` function is used, but it only accepts as arguments single items - not sequences.

Ideally I would like to reference not the `compare` function here, but the comparison algorithm in **`fn:sort`**  that is described under point **[5. (a. b. c. d)](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-sort)**

To do this, could you, please, give this a function name, something like `internal:compare-sort-results` - so that I can just quote it and have a link to it? This will be like how we used to have `op:same-key` in 3.0, which was not made available to the end users, but had a name and could be quoted.

michaelhkay:
Starting a new review from scratch after the recent (very helpful) changes.

1. The prose says "The distinct items from $input are sorted". But that's not what the reference implementation does. Instead it does `distinct-values($input ! $key(.), $collation-input)` - that is, it finds the distinct keys, not the distinct input items. This accounts for some of my earlier confusion. I think it would simplify the function considerably if we dropped this part. If the user wants to eliminate duplicates before calling this function, that is easy enough to do: on the principle that a function should do one thing only, I think it's better to cut out that step.

2. We're then left with the issue of how to identify items with duplicate keys in a way that is compatible with the way fn:sort does it. Short of any refactoring of the `fn:sort` specification, a simple way might be to define:

```
let $equal-sort-keys := function($k1, $k2, $coll) as xs:boolean {
    deep-equal(sort(($k1, $k2), $coll), sort(($k2, $k1), $coll))
}
```

3. I would then suggest defining the ranks function as a sort followed by a partition:

```
sort($input, $key, $collation) => 
partition(split-when := fn($group, $next){not($equal-sort-keys($key(foot($group)), $key($next)))})
```

dnovatchev:
> Starting a new review from scratch after the recent (very helpful) changes.
> 

Thank you, @michaelhkay  -- that is a huge rewrite and I need to read it carefully and have more time to fully understand it.

> 1. The prose says "The distinct items from $input are sorted". But that's not what the reference implementation does. Instead it does `distinct-values($input ! $key(.), $collation-input)` - that is, it finds the distinct keys, not the distinct input items. This accounts for some of my earlier confusion. I think it would simplify the function considerably if we dropped this part. If the user wants to eliminate duplicates before calling this function, that is easy enough to do: on the principle that a function should do one thing only, I think it's better to cut out that step.
> 

We could have a boolean parameter `$distinct-ranks = true()` so if someone needs to have duplicates in the resulting ranks, they can call the function setting this to `false()`

We will not force the user to do any deduplication, yet we will do this work for him (by default). What is really important here is the (deduplication of the) result. not the input itself. How it is actually done is an implementation detail.

> Another strong reason for dropping this step is that eliminating duplicates from the input sequence is a very different kind of operation depending on whether the input comprises nodes, atomic values, or (say) maps.
> 
> 3. We're then left with the issue of how to identify items with duplicate keys in a way that is compatible with the way fn:sort does it. Short of any refactoring of the `fn:sort` specification, a simple way might be to define:
> 
> ```
> let $equal-sort-keys := function($k1, $k2, $coll) as xs:boolean {
>     deep-equal(sort(($k1, $k2), $coll), sort(($k2, $k1), $coll))
> }
> ```

I am trying to understand this and it seems very complicated. Can't we just use: `{ deep-equal($k1, $k2) }`  ?

If we can't then I would prefer something simpler, such as:

```xq
{ count($k1) eq count($k2) 
  and every(for-each-pair($k1, $k2, fn($x, $y) {compare($x, $y) eq 0}  ) }
```

> 
> Another way might simply to be to reference `op:simple-compare()` with a reference to where it is defined (that is, in the rules of fn:sort).

Yes, this is what I suggested in my previous comment.

> 
> 4. I would then suggest defining the ranks function as a sort followed by a partition:
> 
> ```
> sort($input, $key, $collation) => 
> partition(split-when := fn($group, $next){
>             not($equal-sort-keys($key(foot($group)), $key($next), $collation))
>        })
> ```

This is probably the most complicated change, compared to the initially proposed implementation.

Here is the code for `fn:ranks` that seems simpler and easier to understand - it also is called for Example 4:

```xq
let $ranks := function(
                $input as item()*,
                $key as function(item()) as xs:anyAtomicType*,
                $distinct-ranks as xs:boolean,
                $collation-input as xs:string?,
                $collation-keys as xs:string?
                ) as array(item())*
 {
    for $v in sort(distinct-values($input ! $key(.), $collation-input),  $collation-keys)
     return 
        let $make-distinct := ( (distinct-values#2)[$distinct-ranks], fn($x, $y){$x} )[1]
         return
           array{$make-distinct($input[compare($key(.), $v, $collation-keys) eq 0],  $collation-keys)}
 },
 $toSwedish := map{
   "good" : "Bra",
   "difficult" : "svår",
   "great" : "Bra",
   "severe": "svår",
   "fine": "Bra"
 },
 $makeDistinct := true()
return
   $ranks(("good", "good", "difficult", "great", "severe", "fine"), 
          fn($word){$toSwedish($word)}, 
          $makeDistinct,
          (), "http://www.w3.org/2013/collation/UCA?lang=se")
```
This produces:

```xq
["good","great","fine"],
["difficult","severe"]
```

If we just set `$makeDistinct` above to `false()`, this time we get ranks that could contain duplicates:

```xq
["good","good","great","fine"],
["difficult","severe"]
```

dnovatchev:
Here is one more good example - each rank contains a set of all words that are anagrams to each other:

```xq
let  $makeDistinct := true(),
$decompose := fn($s as xs:string)
{
  let $chars := string-to-codepoints($s) ! codepoints-to-string#1(.),
      $freqs := fold-left($chars, map{},  
                          fn($m as map(*), $c as xs:string) 
                          {
                            if(not($m($c))) then $m => map:put($c, 1)
                              else $m => map:put($c, $m($c) +1)
                          }
                         )
      return 
        string-join((for $k in sort($freqs => map:keys()),
                         $kv in $k || $freqs($k)
                      return $kv),
                      '')       
}
 return
   $ranks(("apple macintosh", "astronomer ", "angered", "brush", "dictionary", "dirty room",
           "editor ", "enraged", "eleven plus two",  "indicatory", "laptop machines", "redo it", 
           "dormitory ", "shrub", "moon starer", "twelve plus one"), 
          fn($word){$decompose($word)}, 
          $makeDistinct, (), fn:default-collation() )
```

produces:

```xq
["astronomer ","moon starer"],
["apple macintosh","laptop machines"],
["editor ","redo it"],
["dirty room","dormitory "],
["eleven plus two","twelve plus one"],
["dictionary","indicatory"],
["angered","enraged"],
["brush","shrub"]
```

Arithmeticus:
This function should be useful, thanks for pulling it together. My comments below are based only upon the diff at qt4cg.org

`ranks()` would join three other functions with a sorting parameter `$key` with signature `function(item()) as xs:anyAtomicType*` (vel sim.): `sort()`, `lowest()`, `highest()`. 

I'm struck by places where `ranks()` departs in places from what seems to be a standard formula adopted by the others. Editorial suggestions:

* In the other three functions, the collation parameters always fall second in the list, before the key parameter.
* Rules: start with initial paragraphs about parameter defaults (see `highest()`, Rules, first three paragraphs).
* I strongly recommend adopting the prose-style approach to rules (again, `highest()` provides an excellent model). The inconsistency between `highest()` and `ranks()` in their Rules section is jarring, and diminishes the professional polish of the specs. (A senior editor may need to weigh in here.)
* If `highest()` needs only one collation parameter, why does `ranks()` need two?
* Prose infelicities need to be addressed, but that can be done after PR approval.
* The examples are well chosen, demonstrating the simple, and working up to the clearly practical.
* Example 6:
  * Add comments explaining in regular English the sorting principle intended.
  * `$chars`, just use the new 4.0 `characters()` -- simpler, easier to understand.
  * Rename `$freqs` to something that indicates what is being counted, e.g., `$char-freqs`.
  * Add to the input one word that is not an anagram of any other.
  * Add to the input one word that is a subset anagram of more than one group, e.g., "me".

One substantive point: users of this function will immediately wonder why the key that was used to do the sorting is not also returned, and consider it a serious loss. For that reason (among others), I'm still inclined to use `xsl:group-by` rather than this function for anything but the simplest of grouping procedures.

michaelhkay:
I have created an editorially revised version of the PR at PR #1062 which attempts to take the comments into account.

dnovatchev:
> This function should be useful, thanks for pulling it together. My comments below are based only upon the diff at qt4cg.org
> 
> `ranks()` would join three other functions with a sorting parameter `$key` with signature `function(item()) as xs:anyAtomicType*` (vel sim.): `sort()`, `lowest()`, `highest()`.

Thanks @Arithmeticus for the comments.

> 
> I'm struck by places where `ranks()` departs in places from what seems to be a standard formula adopted by the others. 

Yes, this was done intentionally. "Different" is not equivalent to "worse". In this particular  case, it is intended that `fn:ranks` should be simpler and easier to use than the referred other 3 functions.

Editorial suggestions:
> 
> * In the other three functions, the collation parameters always fall second in the list, before the key parameter.

The collations are intentionally the last parameters in the signature. There is a good principle that the least-frequently used parameters should be the last in the signature. In this particular case the two parameters that will always be used are `$input` and `$key`. We want to give the user the ability to call the function with specifying positionally the argument values for only these two parameters, without always having to specify in the function call `()` for any collation (collations will be unneeded in 99% of the cases)

> * Rules: start with initial paragraphs about parameter defaults (see `highest()`, Rules, first three paragraphs).
> * I strongly recommend adopting the prose-style approach to rules (again, `highest()` provides an excellent model). The inconsistency between `highest()` and `ranks()` in their Rules section is jarring, and diminishes the professional polish of the specs. (A senior editor may need to weigh in here.)
> * If `highest()` needs only one collation parameter, why does `ranks()` need two?

Because `highest()` returns results with duplicates. The `ranks()` function gives the user the option to produce ranks with distinct values. In order to do this, in the cases when the values are strings, a collation for the string comparison may be needed (if the default collation is not the right to use).

> * Prose infelicities need to be addressed, but that can be done after PR approval.
> * The examples are well chosen, demonstrating the simple, and working up to the clearly practical.
> * Example 6:
>   
>   * Add comments explaining in regular English the sorting principle intended.
>   * `$chars`, just use the new 4.0 `characters()` -- simpler, easier to understand.
>   * Rename `$freqs` to something that indicates what is being counted, e.g., `$char-freqs`.
>   * Add to the input one word that is not an anagram of any other.
>   * Add to the input one word that is a subset anagram of more than one group, e.g., "me".
> 
> One substantive point: users of this function will immediately wonder why the key that was used to do the sorting is not also returned, and consider it a serious loss. For that reason (among others), I'm still inclined to use `xsl:group-by` rather than this function for anything but the simplest of grouping procedures.

We could have a separate function for this - in most cases the real values on which the ranking is done are not needed.
This new function will return an array containing two sequences: one containing the ranks and the second containing the ranking values.

One possible name for this new function could be: **`fn:ranks-with-values`**

ChristianGruen:
> without always having to specify in the function call `()` for any collation (collations will be unneeded in 99% of the cases)

I believe to remember there was a similar discussion for `fn:sort($input, $key, $collation)`, which was eventually changed to `fn:sort($input, $collation, $key)` to be consistent with the remaining functions – and which indeed caused confusion for the reasons you mentioned. To stay consistent, and as we have keyword arguments today, I nevertheless think we should stick with that choice for future functions.

dnovatchev:
> > without always having to specify in the function call `()` for any collation (collations will be unneeded in 99% of the cases)
> 
> I believe to remember there was a similar discussion for `fn:sort($input, $key, $collation)`, which was eventually changed to `fn:sort($input, $collation, $key)` to be consistent with the remaining functions – and which indeed caused confusion for the reasons you mentioned. To stay consistent, and as we have keyword arguments today, I nevertheless think we should stick with that choice for future functions.

It is always better not "to stay consistent" with problematic work, but to solve the problems that it has.

Or is there a "ultimate authority" whose creations are not to be discussed and that cannot be improved as a rule?

If so, this would be much worse than my memories of living in a socialist country.

ChristianGruen:
> Or is there a "ultimate authority" whose creations are not to be discussed and that cannot be improved as a rule?
> 
> If so, this would be much worse than my memories of living in a socialist country.

I hope it’s the contrary, i.e., a democratic decision. None of us should be individually entitled to choose what’s right; it should always be up to the majority. Back then, the majority seemed to have voted differently than we would probably have voted.

In turn, please remember that what I or you regard as the best solution may not necessarily be the best decision for the majority. Everyone of us is limited.

dnovatchev:
> > Or is there a "ultimate authority" whose creations are not to be discussed and that cannot be improved as a rule?
> > If so, this would be much worse than my memories of living in a socialist country.
> 
> I hope it’s the contrary, i.e., a democratic decision. None of us should be individually entitled to choose what’s right; it should always be up to the majority. Back then, the majority seemed to have voted differently than we would probably have voted.
> 
> In turn, please remember that what I or you regard as the best solution may not necessarily be the best decision for the majority. Everyone of us is limited.

There is a definite problem and it has nothing to do with "democracy by majority".

People tend not to react clearly to very complicated proposals and thus proposals that are over-complicated have better chance of being accepted by the "majority".

This is what happened with the proposal for the extended functionality of `fn:sort` and I am also partly to blame (as part of that "majority" stumped by complexity) for not reacting sooner.

When I had the chance to properly read the accepted (don't think "perpetuated") prose for `fn:sort`, I realized how inflated and unnecessarily complicated this was. I asked myself: "As a developer, would you welcome this function, or would you prefer a simpler and more understandable one?" And the answer was overwhelmingly in favor of the simpler alternative.

I think we have to recognize this general decision problem we have as a group and change the way we approve proposals - especially the complex ones.

I remember that once in the past I raised the problem of having at least one person tasked beforehand to properly read a proposal and to try understand it in its entirety. Then this person to give their own opinion (in addition to the original author) about the proposal.

If we had such process in place, this could have prevented many cases where over-complicated proposals slip into the Spec because the "majority" fails to understand their intricate meaning.

ChristianGruen:
> People tend not to react clearly to very complicated proposals and thus proposals that are over-complicated have better chance of being accepted by the "majority".

That’s certainly my impression, too.

> This is what happened with the proposal for the extended functionality of `fn:sort` and I am also partly to blame (as part of that "majority" stumped by complexity) for not reacting sooner.

Just to be sure: I was referring to XQFO 3.1.

> I remember that once in the past I raised the problem of having at least one person tasked beforehand to properly read a proposal and to try understand it in its entirety. Then this person to give their own opinion (in addition to the original author) about the proposal.

Sounds good, in principle. I’m just not sure what that would change: Personally, I do my best to review every single proposal that is created by another person (maybe also because I’ll be one of those that need to implement it).

I would certainly appreciate if more people joined in, and gave feedback on my proposals before the weekly meeting, but it seems everyone’s resources are limited. After all, we all (I assume) are not paid for this work.

dnovatchev:
> > I remember that once in the past I raised the problem of having at least one person tasked beforehand to properly read a proposal and to try understand it in its entirety. Then this person to give their own opinion (in addition to the original author) about the proposal.
> 
> Sounds good, in principle. I’m just not sure what that would change: Personally, I do my best to review every single proposal that is created by another person (maybe also because I’ll be one of those that need to implement it).
> 
> I would certainly appreciate if more people joined in, and gave feedback on my proposals before the weekly meeting, but it seems everyone’s resources are limited. After all, we all (I assume) are not paid for this work.

I would prefer to spend a little bit more time reading and understanding (being assigned with this) or hearing the person assigned to do so, than realizing when it is too-late that everybody's  time  was wasted at one or more meetings due to unrealized complexity and lack of understanding.

ChristianGruen:
> I would prefer to spend a little bit more time reading and understanding (being assigned with this) or hearing the person assigned to do so, than realizing when it is too-late that everybody's time was wasted at one or more meetings due to unrealized complexity and lack of understanding.

»Everybody« implies I’m part of it, but I don’t see myself involved. Are you sure others, or even all of us, share your perspective?

When I have the impression that a feature is too complex to be accepted, I tend to ask for more time before we accept it.

What I’ve indeed suggested just recently is that we should spend time on the features that have already been added to the draft, but have not been officially accepted (https://lists.w3.org/Archives/Public/public-xslt-40/2024Feb/0016.html). I didn’t get any reply, so it could be that people don’t feel it’s necessary (or again it’s a matter of time).

dnovatchev:
> > I would prefer to spend a little bit more time reading and understanding (being assigned with this) or hearing the person assigned to do so, than realizing when it is too-late that everybody's time was wasted at one or more meetings due to unrealized complexity and lack of understanding.
> 
> »Everybody« implies I’m part of it, but I don’t see myself involved. Are you sure others, or even all of us, share your perspective?
> 
> When I have the impression that a feature is too complex to be accepted, I tend to ask for more time before we accept it.
> 
> What I’ve indeed suggested just recently is that we should spend time on the features that have already been added to the draft, but have not been officially accepted (https://lists.w3.org/Archives/Public/public-xslt-40/2024Feb/0016.html). I didn’t get any reply, so it could be that people don’t feel it’s necessary (or again it’s a matter of time).

If it is regularly the case that I don't understand well at least 50% of what someone is writing, should I constantly raise this (might well be mistaken for having a personal grudge or embarrassment) or should we deal in a more organized, systematic way? And what if I am not the only one who feels that way and who is shy to raise their voice? Doesn't this make for a significant part of the people (maybe even the majority)?

I think it is the Chair's responsibility not to ask for a vote if there is even the slightest sense of not understanding and discomfort. Maybe we are often rushed to make decisions when we are still not fully prepared to do so?

ChristianGruen:
> If it is regularly the case that I don't understand well at least 50% of what someone is writing, should I constantly raise this (might well be mistaken for having a personal grudge or embarrassment)

I welcome this personally (neutral language might helps to avoid irritations). In addition, I have repeatedly observed that my lack of native language skills lead to technical misunderstandings that I like to have clarified myself.

> Doesn't this make for a significant part of the people (maybe even the majority)?

…could very well be the case.

> I think it is the Chair's responsibility not to ask for a vote if there is even the slightest sense of not understanding and discomfort.

We should take in mind that a too strict procedure might lead to stagnancy. Several years have already passed, and we are far from finalizing version 4.

But I think we would not lose anything by spending 10 or 20 minutes of our joint time to discuss the current procedure in an upcoming meeting.

My personal hope is slightly different: I think we all should be as open-minded as possible to accept others’ thoughts and opinions. It hurts to see a PR questioned for which one has spent hours and hours to make it seemingly water-proof. However, that doesn’t prevent anyone of us to be confronted with a result that differs a lot from the initial proposal.

When saying this, I hope not to be suggestive. I don’t refer to this specific proposal; I rather have my own proposals in mind that underwent various changes before becoming accepted or eventually rejected.

ChristianGruen:
@dnovatchev Thanks for the example code. I took the liberty of pasting your [reply to the mailing list](https://lists.w3.org/Archives/Public/public-xslt-40/2024Mar/0010.html):

-----------------

Yes, any sequence of functions can be replaced by a single function.

Here is one such example:

We are given  a company's employees and each employee has a name, department and salary.

We will rank the employees first just by department, then by both department and salary - done with a single function as specified in the 2nd call to *fn:ranks* below:

let $employees := map{
    "John Smith": map{ "dept": "Sales", "salary": 50000},
    "Erin Carter": map{ "dept": "Computing", "salary": 120000},
    "Ryan Gosling": map{ "dept": "Sales", "salary": 100000},
    "Ann Gould": map{ "dept": "Computing", "salary": 150000},
    "Pete Lagard": map{ "dept": "Sales", "salary": 50000},
    "Jim Carter": map{ "dept": "Sales", "salary": 80000},
    "Greg Wilson": map{ "dept": "Computing", "salary": 120000}
  }
    return
    (
      ranks(map:keys($employees), fn($emp){$employees($emp)("dept")}),

"===============================================================================",
      ranks(map:keys($employees),
             fn($emp){$employees($emp)("dept")
                    || (let $sal := $employees($emp)("salary"),
                           $salDigits := string-length(string($sal))
                         return substring('0000000', $salDigits +1) ||
string($sal)  )})
    )

-----------------

I see that the concatenated string seems to be based on the actual value distribution of the input data (e.g., knowledge on the maximum value)…

1. How would you handle arbitrary numbers (e.g. doubles)?
2. How would you sort a secondary double sort key in a descending order?

dnovatchev:
> @dnovatchev Thanks for the example code. I took the liberty of pasting your [reply to the mailing list](https://lists.w3.org/Archives/Public/public-xslt-40/2024Mar/0010.html):

[@Christian Grün](mailto:cg@basex.org) Thanks, but I actually sent this to the mailing list and to Norm Walsh  and not to any other recipient.


image

ChristianGruen:
> [@Christian Grün](mailto:cg@basex.org) Thanks, but I actually sent this to the mailing list and to Norm Walsh  and not to any other recipient.

That’s what I wanted to say (might have been a misunderstanding?).

dnovatchev:
> I see that the concatenated string seems to be based on the actual value distribution of the input data (e.g., knowledge on the maximum value)…
> 
> 1. How would you handle arbitrary numbers (e.g. doubles)?


A general answer: A double is less precise than a decimal, and the example shows how to handle decimals - thus handling doubles can be done in a similar way

> 2. How would you sort a secondary double sort key in a descending order?

By substituting it with the difference between a suitable constant and this value.For any `x` we use `N - x` where `N` is  the largest possible value.

ChristianGruen:
> A general answer: A double is less precise than a decimal, and the example shows how to handle decimals - thus handling doubles can be done in a similar way

This is how I would sort ascending strings and descending doubles with two keys:

```xquery
let $items := (
  map { 'name': 'A', 'size': 1e33  },
  map { 'name': 'A', 'size': .1    },
  map { 'name': 'B', 'size': 0.01  },
  map { 'name': 'B', 'size': -1e99 }
)
return sort($items, (), (fn { ?name }, fn { -?size }))
```

How would you do it with a single key?

dnovatchev:
> > A general answer: A double is less precise than a decimal, and the example shows how to handle decimals - thus handling doubles can be done in a similar way
> 
> This is how I would sort ascending strings and descending doubles with two keys:
> 
> ```xq
> let $items := (
>   map { 'name': 'A', 'size': 1e33  },
>   map { 'name': 'A', 'size': .1    },
>   map { 'name': 'B', 'size': 0.01  },
>   map { 'name': 'B', 'size': -1e99 }
> )
> return sort($items, (), (fn { ?name }, fn { -?size }))
> ```
> 
> How would you do it with a single key?

There are many ways to do this.

We can even return the hash of the concatenation of the 'name' and a normalized (meaning having the same agreed upon representation) of the 'size'.

dnovatchev:
> > > A general answer: A double is less precise than a decimal, and the example shows how to handle decimals - thus handling doubles can be done in a similar way
> > 
> > 
> > This is how I would sort ascending strings and descending doubles with two keys:
> > ```xq
> > let $items := (
> >   map { 'name': 'A', 'size': 1e33  },
> >   map { 'name': 'A', 'size': .1    },
> >   map { 'name': 'B', 'size': 0.01  },
> >   map { 'name': 'B', 'size': -1e99 }
> > )
> > return sort($items, (), (fn { ?name }, fn { -?size }))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > How would you do it with a single key?
> 
> There are many ways to do this.
> 
> We can even return the hash of the concatenation of the 'name' and the normalized (meaning having the same agreed upon representation) of the 'size'.

This is the maximum double value:

dMax  =`1.7976931348623157E+308`

Use: `dMax - ?size`, then convert this to a fixed-length string with the decimal representation, then concat the result to `?name`.

There are many possible ways to compute the final single value, and I am not saying that I can immediately provide the best algorithm to do that.

The statement is that all this can be done with a single function.

dnovatchev:
I will am closing this PR because it is from my master branch and this is not good when one has more than one open PRs.

Will re-submit it from a dedicated feature branch.

Issue #340, opened 2023-02-07
fn:format-number: Specifying decimal format


ChristianGruen:
It would be nice if the decimal format for `fn:format-number` could be specified via an options argument:

```xquery
format-number(
  123.45,
  'format': map { 'decimal-separator': ',', 'grouping-separator': '.' }
)
```

michaelhkay:
Yes, the current mechanism is very clumsy. I think the original intent in XSLT 1.0 was probably to define presentation "at arm's length" so that the logic didn't need to change if the output format changed, but that can be achieved perfectly well by putting the options in a global variable.

ChristianGruen:
I think that language-specific default settings would be sensible:

```
format-number(123.45, '#.##0,00', 'de')
```

As known from the other functions for formatting numbers and dates, it could be up to the implementation to decide which languages are supported. The defaults could be overwritten by custom decimal-format declarations in the prolog to ensure that a setting is applied, even if an implementation does not support it.

michaelhkay:
I'm not convinced this would give good interoperability. Consider Arabic for example: should it default to using western or eastern decimal digits? Both are in widespread use, and the idea that everyone with a particular (country, language) combination uses the same conventions is fundamentally misguided. This doesn't matter too much if it merely affects the format of the output, but it does matter if it makes a picture string valid in one implementation and invalid in another.

ChristianGruen:
I agree, there are cases which are easier to handle and others are more sophisticated. I think the same is true for formatting integers and dates: The rules are rich and sophisticated, but for more advanced use cases (such as spelling out correct hiragana for numbers with Japanese counter words, or considering declension of numerals in Russian), you’ll be lost without writing custom code.

With ICU and Java, it’s fairly straightforward to choose language-specific formatting rules. I haven’t checked if there are flags to e.g. control formatting for Arabic numbers, and it could be that ICU has really taken the wrong path. From a German perspective, though, it’s restrictive that an implementation cannot provide sane defaults for Non-English users.

This is how ICU formats integers with different locales:

Result | Locales
--- | ---
[...]

PR #1049, opened 2024-02-28
340-partial fn:format-number: Specifying decimal format

ChristianGruen:
The PR introduces an additional `$format` argument to `fn:format-number`, which allows you to override decimal formats with custom properties.

Next, we may need to clarify if the current specification already allows processors to provide custom decimal formats (https://github.com/qt4cg/qtspecs/issues/340#issuecomment-1968856655). It’s not part of this PR.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -2262,6 +2262,8 @@
                         where the URI can be zero-length to indicate a name in no namespace.</p>
                   </item>
                </ulist>
+            </item>
+            <item>
                <p>The decimal format that is chosen is the decimal format in the static context
                   whose name matches <code>$format-name</code> if supplied, or the unnamed decimal
                   format in the static context otherwise.</p>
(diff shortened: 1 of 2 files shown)


michaelhkay:
I don't think the proposed spec states clearly that if neither $format-name nor $format is supplied, the decimal format used is the unnamed decimal format in the static context. The problem is that there are two paragraphs describing what happens (a) if $format-name is present and non-empty, and (b) if $format is present and non-empty; there should really be four paragraphs describing all four possible combinations

ChristianGruen:
> I don't think the proposed spec states clearly that if neither $format-name nor $format is supplied, the decimal format used is the unnamed decimal format in the static context. The problem is that there are two paragraphs describing what happens (a) if $format-name is present and non-empty, and (b) if $format is present and non-empty; there should really be four paragraphs describing all four possible combinations

I hoped it was clear:
* I used an ordered list to indicate that there are multiple subsequent steps.
* The last sentence of 1. says that “The decimal format that is chosen is the decimal format in the static context whose name matches `$format-name` if supplied, or the unnamed decimal format in the static context otherwise.”
* In Step 2, I refer to the result of the first step.

With 4 separate paragraphs, I think there would be lots of redundancy.

michaelhkay:
The problem is that bullet point 1 starts with the conditional "If $format-name is present and non-empty" and it's not clear how much text is scoped to that conditional. The second paragraph of bullet point 1 can be read as being within the conditional, but the last phrase "or the unnamed decimal format in the static context otherwise" is logically outside it.

Issue #1054, opened 2024-03-01
Spec fn:message #id using old name fn:log


apb2006:
- https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-message
- https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-log

PR #1057, opened 2024-03-04
1054 Spec fn:message #id using old name fn:log

ChristianGruen:
Issue: #1054


ChristianGruen:
This editorial PR won’t cause any discussion for sure, so I’m merging it.

Issue #1037, opened 2024-02-21
fn:json-to-xml: 'number-parser' option


ChristianGruen:
A function supplied via the `number-parser` option of `fn:json-to-xml` is now allowed to return zero or <del>more</del> **one** items. Analogous to the `action` argument of `fn:replace`, the result should be converted to a string by invoking `fn:string` on the result. An example:

```xquery
json-to-xml('-1', map { 'number-parser': abs#1 })
→ <fn:number>1</fn:number>

json-to-xml('1', map { 'number-parser': fn { true#0 } })
→ err:FOTY0013
```

No change is required for `fn:parse-json`.

michaelhkay:
>A function supplied via the number-parser option of fn:json-to-xml is now allowed to return zero or more items

I think it's actually zero-or-one.

ChristianGruen:
> I think it's actually zero-or-one.

Thanks, true; has been fixed in the initial comment.

PR #1058, opened 2024-03-04
1037 fn:json-to-xml: 'number-parser' option

ChristianGruen:
Issue: #1037


michaelhkay:
I think the intent is fine, but I find the sentence "The result is obtained by invoking [fn:string] for the result of the function call." confusing.

I would suggest wording along the lines: "The string value of the `number` element generated in the result will be the value obtained by calling the supplied function, and then converting its result to a string by calling fn:string#1".

ndw:
The CG agreed to merge this PR without discussion at [meeting 069](https://qt4cg.org/meeting/minutes/2024/03-12.html).

Issue #1019, opened 2024-02-11
XQFO: Unknown option parameters


ChristianGruen:
The current [option parameter conventions](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#option-parameter-conventions) are:

> 4. It is not an error if the options map contains options with names other than those described in this specification. Implementations MAY attach an ·implementation-defined· meaning to such entries, and MAY define errors that arise if such entries are present with invalid values. Implementations MUST ignore such entries unless they have a specific ·implementation-defined· meaning. Implementations that define additional options in this way SHOULD use values of type `xs:QName` as the option names, using an appropriate namespace.

The obvious consequence is that wrongly typed or unsupported options are not reported as such:

```xquery
serialize($node, map { 'format': 'html' })
```

I think we should still allow proprietary options, but raise errors when an option is neither defined in the specification nor supported by the given implementation. On the one hand, this will help users to spot typos (e.g., `byte-order-mask` or instead of `byte-order-mark`). On the other hand, options that are supported by one implementation will be rejected, which feels reasonable to me, as options usually change either the result, or the way how the input is treated.

If we believe that this change is too disruptive, we could tolerate entries with `xs:QName` keys.

michaelhkay:
Incidentally, F+O §1.6 still refers to the "function conversion rules" by their old name, with a link to XP31.

michaelhkay:
I hit an interesting example today: I wrote a test that did:

```
csv-to-arrays("a,b,c,d,e,f|p,q,r,s,t,u", 
             map{'row-delimiter':'|',  'select-columns':1 to 3}) 
```

It's easy to forget that `select-columns` is allowed on `parse-csv` but not on `csv-to-arrays`. Perhaps we should change that; but in the meantime, silently ignoring the option isn't exactly helpful.

PR #1059, opened 2024-03-04
1019 XQFO: Unknown option parameters

ChristianGruen:
Issue: #1019

### DIFF
--- a/specifications/xpath-functions-40/src/xpath-functions.xml
+++ b/specifications/xpath-functions-40/src/xpath-functions.xml
@@ -683,20 +683,20 @@ for transition to Proposed Recommendation. </p>'>
                   <note><p>This means that the implementation of the function can check for the
                presence and value of particular options using the functions <code>map:contains</code>
                and/or <code>map:get</code>.</p></note></item>
-               <item><p>It is not an error if the options map contains options with names other than those
-               described in this specification. Implementations <rfc2119>may</rfc2119> attach an 
-                  <termref def="implementation-defined">implementation-defined</termref> meaning to such entries,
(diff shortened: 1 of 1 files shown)


michaelhkay:
Thinking about this further, I'm unconvinced. 

If someone uses `saxon:double-space="x" ` as a serialization parameter, and then runs the query under BaseX, do you really want to require the implementation to throw an error? This would have a number of consequences:

(a) queries that currently run successfully would now fail.

(b) an implementation is forced to look at all options supplied, not only those it cares about.

(c) portability even across different versions/variants of the same product would become more difficult.

While I agree that the change would ensure that misspellings are caught, I'm not sure that it's a net win.

ChristianGruen:
If the proposed change seems to strict, we could change MUST to SHOULD or MAY (and thus make it completely implementation-defined):

> If an option is neither described in the specification nor supported by the implementation, a type error `FORG0013` SHOULD be raised.

ChristianGruen:
As motivated by @ndw (thanks), I’ve rewritten the rules to clarify that only those unknown options that are in no namespace must be rejected.

@michaelhkay pointed out that we could use plausibility checks. I guess the general rules are described in [2.4.6 Implausible Expressions](https://qt4cg.org/specifications/xquery-40/xpath-40.html#id-implausible-expressions). I’ve sticked with the existing presentation, as I’m not sure how this would need to be incorporated into the text.

ndw:
Looks good to me. I'm tempted to mark this as "merge without discussion" but in any event, I hope the discussion can be brief this time around :-)

ndw:
The CG agreed to merge this PR at [meeting 071](https://qt4cg.org//meeting/minutes/2024/03-26.html)

Issue #73, opened 2021-05-07
Split a string by graphemes


rhdunn:
The new `fn:characters` function is useful, but doesn't solve a problem of manipulating strings where multiple codepoints correspond to a single grapheme. For example:
1. characters with one or more combining characters;
2. emoji with skin tone variant selectors;
3. emoji with gender variant selectors;
4. multi-sequence emoji -- family, wales flag, etc.;
5. region indicator pairs for flags.

Getting this right is complex, and implementing it as a regular expression is easy to get wrong/make mistakes.

# fn:graphemes
## Summary
Splits the supplied string into a sequence of single-grapheme (one or more character) strings.
## Signature
```fn:graphemes($value as xs:string?) as xs:string*```
## Properties
This function is ·deterministic·, ·context-independent·, and ·focus-independent·.
## Rules
The function returns a sequence of strings, containing the corresponding ·grapheme· in $value. These are determined by the corresponding Unicode rules for what constitutes a ·grapheme·. The version of Unicode and the Unicode Emoji standards is ·implementation-dependent·.

If $value is a zero-length string or the empty sequence, the function returns the empty sequence.
## Examples
> The expression `fn:graphemes("Thérèse")` returns `("T", "h", "é", "r", "è", "s", "e")`, irrespective of whether the e characters use combining characters or not.
>
> The expression `fn:graphemes("")` returns `()`.
>
> The expression `fn:graphemes(())` returns `()`.
>
> The expression `fn:graphemes("👋🏻👋🏼👋🏽👋🏾👋🏿")` returns `("👋🏻", "👋🏼", "👋🏽", "👋🏾", "👋🏿")`.
>
> The expression `fn:graphemes("👪")` returns `("👪")`.
>
> The expression `fn:graphemes("👨‍🔬👩‍🔬")` returns `("👨‍🔬", "👩‍🔬")`.
>
> The expression `fn:graphemes("🇪🇪🇩🇪🇫🇷🏴󠁧󠁢󠁷󠁬󠁳󠁿🇮🇸")` returns `("🇪🇪", "🇩🇪", "🇫🇷", "🏴󠁧󠁢󠁷󠁬󠁳󠁿", "🇮🇸")`.

rhdunn:
It might also be worth adding a note to `fn:characters` about this issue and referencing `fn:graphemes` for the use xases where preserving graphemes is required.

Conal-Tuohy:
[...]

PR #1068, opened 2024-03-08
73 fn:graphemes

Arithmeticus:
First draft of `fn:graphemes`, in response to discussion at #73 .

A battery of tests will be submitted as a PR to the qt4tests repository.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -28119,6 +28119,104 @@ return every($dl/*, fn($elem, $pos) {
          <fos:version version="4.0">New in 4.0. Accepted 2022-09-20, subject to improving the description.</fos:version>
       </fos:history>
    </fos:function>
+   
+   <fos:function name="graphemes" prefix="fn">
+      <fos:signatures>
+         <fos:proto name="graphemes" return-type="xs:string*">
+            <fos:arg name="value" type="xs:string?"/>
+            <fos:arg name="options" type="map(*)?" default="map{}"/>
+         </fos:proto>
+      </fos:signatures>
+      <fos:properties arity="1">
+         <fos:property>deterministic</fos:property>
+         <fos:property>context-independent</fos:property>
+         <fos:property>focus-independent</fos:property>
+      </fos:properties>
+      <fos:properties arity="2">
+         <fos:property>deterministic</fos:property>
+         <fos:property>context-independent</fos:property>
+         <fos:property>focus-independent</fos:property>
+      </fos:properties>
+      <fos:summary>
+         <p>Splits the supplied string into a sequence of single-grapheme strings.</p>
+      </fos:summary>
+      <fos:rules>
+         <p>The <code>$options</code> argument, if present, defines additional parameters
+            controlling how the process is conducted. The <termref def="option-parameter-conventions"
+               >option parameter conventions</termref> apply.</p>
+         <p>If the one-argument version of the function is used, the result is the same as calling
+            the two-argument version, with the default <code>$options</code> settings. </p>
+         <p>The function returns a sequence of strings. Each string contains one or more <termref
+               def="character">character</termref>s that collectively constitute a single grapheme
+            cluster, as defined by <bibref ref="UNICODE-TR29"/>.</p>
+         <p>If <code>$value</code> is a zero-length string or the empty sequence, the function returns
+            the empty sequence.</p>
+         
+         <fos:options>
+            <fos:option key="grapheme-clusters">
+               <fos:meaning>
+                  <p>Specifies the name of the cluster boundaries definition to be 
+                     used.</p>
+                  <p>Conforming implementations <rfc2119>must</rfc2119> support the following
+                     options:</p>
+                  <ulist>
+                     <item><p><code>extended</code>: the Extended Grapheme Cluster Boundaries 
+                        defined by 
+                           <bibref ref="UNICODE-TR29"/>.</p></item>
+                     <item><p><code>legacy</code>: the Legacy Grapheme Cluster Boundaries 
+                        defined by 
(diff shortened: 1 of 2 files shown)


Arithmeticus:
First draft of tests: https://github.com/qt4cg/qt4tests/pull/77

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -28119,6 +28119,104 @@ return every($dl/*, fn($elem, $pos) {
...
+      <fos:rules>
+         <p>The <code>$options</code> argument, if present, defines additional parameters
+            controlling how the process is conducted. The <termref def="option-parameter-conventions"
+               >option parameter conventions</termref> apply.</p>
+         <p>If the one-argument version of the function is used, the result is the same as calling
+            the two-argument version, with the default <code>$options</code> settings. </p>
+         <p>The function returns a sequence of strings. Each string contains one or more <termref
+               def="character">character</termref>s that collectively constitute a single grapheme
+            cluster, as defined by <bibref ref="UNICODE-TR29"/>.</p>
+         <p>If <code>$value</code> is a zero-length string or the empty sequence, the function returns
```
ChristianGruen, line 28152: A minor issue: This condition could be limited to the empty sequence, as (I assume) the core split algorithm implies that a zero string will return zero characters.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -28119,6 +28119,104 @@ return every($dl/*, fn($elem, $pos) {
...
+            the empty sequence.</p>
+         
+         <fos:options>
+            <fos:option key="grapheme-clusters">
```
ChristianGruen, line 28156: Maybe just `clusters`.

I failed finding information on the two variants in the quoted reference. Could you possibly add some more details how a legacy implementations would differ from the default implementation?

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -28119,6 +28119,104 @@ return every($dl/*, fn($elem, $pos) {
...
+                  </ulist>
+                  <p>Conforming implementations <rfc2119>may</rfc2119> support other grapheme
+                     cluster boundaries with implementation-defined semantics.</p>
+               </fos:meaning>
+               <fos:type>xs:string</fos:type>
+               <fos:default>extended</fos:default>
+            </fos:option>
+         </fos:options>
+      </fos:rules>
+      <fos:examples>
```
ChristianGruen, line 28178: Examples for both the Legacy and Extended variant would be helpful to understand the difference.

ChristianGruen (changes requested):

michaelhkay:
It might be helpful in the examples to indicate what the actual codepoints in the input strings are, rather than relying on the way the browser renders them. For example, some of the characters might be in composed or decomposed form, and we can't tell just by looking at them. One solution might be to write the example as `let $s := "..." return [string-to-codepoints($s), graphemes($s)]`.

specifications/xpath-functions-40/src/function-catalog.xml
Arithmeticus: see question raised below

specifications/xpath-functions-40/src/function-catalog.xml
Arithmeticus: see question raised below

Arithmeticus:
Regarding legacy versus extended grapheme clusters: I would like to suggest that we can drop the legacy as a required option.

> The Unicode Standard provides a default algorithm for determining grapheme cluster boundaries; the default grapheme clusters are also known as **extended grapheme clusters**. For backwards compatibility with earlier versions of this specification, the Standard also defines a profile for **legacy grapheme clusters**.

Of course, nothing forbids an implementer from supporting legacy, per the options map rules.

That will simplify the specs and tests, and not saddle implementations with legacy support.  Any objections?

specifications/xpath-functions-40/src/function-catalog.xml
Arithmeticus: Here I tried merely to follow `fn:characters`. If `fn:graphemes('')` should entail `''` then `fn:characters` should do so as well, no?

ChristianGruen:
I suggest dropping the options argument. The specification should ideally not motivate implementors to do things differently unless there are compelling reasons to do so.

specifications/xpath-functions-40/src/function-catalog.xml
ChristianGruen: I agree. Consistency wins.

ChristianGruen (approved):

ndw:
The CG agreed to merge this PR at [meeting 077](https://qt4cg.org/meeting/minutes/2024/05-14.html)

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -28168,6 +28168,92 @@ return every($dl/*, fn($elem, $pos) {
...
+               <fos:result>$crlf</fos:result>
+               <fos:postamble>Carriage return + line feed, two characters, one
+                  grapheme</fos:postamble>
+            </fos:test>
+         </fos:example>
+         <fos:example>
+            <fos:test>
+               <fos:expression>graphemes(char(0x1F476) || char(0x200D) 
+|| char(0x1F6D1))</fos:expression>
+               <fos:result>char(0x1F476) || char(0x308) || char(0x1F6D1)</fos:result>
```
gimsieke, line 28232: Shouldn’t this `char(0x308)` read `char(0x200D)`?

gimsieke (commented):
Since this has already been merged, I should probably issue a new PR?

ChristianGruen:
> Since this has already been merged, I should probably issue a new PR?

Yes, that’ll probably be the easiest way.

PR #1078, opened 2024-03-13
1060 Formatting XPath/XQuery

### LINKED ISSUE #1060: Formatting XPath/XQuery
I got reminded today that the specification documents are kind of “wild”, because all code snippets use a different formatting:

* The indentation is inconsistent (the tendency seems to be 2 spaces, in accordance with the function signatures). Repeatedly, indentations are used that don’t seem to follow any conventions at all.
* Sometimes, `function`, `map` and `array` keywords are followed by a space, sometimes not. My preference would be not be too stingy; we have enough space.
* Sometimes, the `return` keyword starts in a new line, sometimes it’s attached to the previous line.

This is certainly something we cannot finalize too early, but I think we shouldn’t be too erratic in an official document, even though it’s “just code”.

Related: #1000.

ChristianGruen:
Editorial; closes #1060.

This PR attempts to unify the presentation of XPath and XQuery code. It’s not complete, but it should definitely improve the status quo.

The chosen formatting and indentation rules can certainly be discussed. My major objective was consistency: I selected rules that were used frequently enough in the given documents.

Apart from the presentation stuff, this PR fixes various minor bugs in the rules and examples.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -30757,13 +30757,13 @@ return $result
       <fos:examples>
          <fos:example>
             <fos:test>
-               <fos:expression>pin([ "a", "b", "c" ])?1 => label()?parent => array:foot()</fos:expression>
+               <fos:expression>pin([ "a", "b", "c" ])?1 ! label(.)?parent ! array:foot(.)</fos:expression>
                <fos:result>"c"</fos:result>
             </fos:test>
          </fos:example>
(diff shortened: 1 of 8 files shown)


ChristianGruen:
PS: If this PR is accepted, it’s like to cause conflicts with other PRs, so it will be best to merge it last.

ndw:
The CG agreed to merge this PR without further discussion at [meeting 071](https://qt4cg.org//meeting/minutes/2024/03-26.html)

PR #1098, opened 2024-03-19
Editorial improvements per recent discussion on #566

ndw:
1. Add a note clarifying that the fragment identifier should be (1) URI decoded and (2) ignored if it's the empty string.
2. Reworked a bit of the description in order to avoid an ambiguity in how `///abc` should be parsed. (The current spec can be satisfied either by parsing it as `//(/abc)()` or `//()/abc` and only the former is intended.)


ChristianGruen:
@ndw I’ll be happy to check out the changes!

By the way, https://qt4cg.org/dashboard/ is currently not working:

image

ndw:
Regarding the dashboard: come back in an hour or change your IP address with a VPN.

Loading the dashboard makes requests to the GitHub API. The browser runs as an unautheticated user because I couldn't be bothered to work out the login dance and force everyone using the dashboard to login. GitHub rate limits unauthenticated API requests to 100 per hour.

That's bad if you reload a lot, especially if there are a lot of open PRs. :-(

ChristianGruen:
* If the string matches `^///*([^/]+)?$`, the authority is the first match group and the string is empty.

I think this new rule contradicts with what you indicated in https://github.com/qt4cg/qtspecs/issues/566#issuecomment-2004241476 (didn’t you want to assign the value to the string, not the authority?):

>> Otherwise, if the string matches `^///*([^/]+)?(/.*)?$`, the authority is the first match group and the string is the second match group.
> That reduces the string to `/abc`. That becomes the path and the path segments are ('', 'abc').

Next, I wonder if it should be `^///*([^/]+)?$` or `^///*([^/]+)$` ?

* If the string matches `^///*([^/]+)?(/.*)?$`, the authority is the first match group and the string is the second match group.

Same here… Dropping the question marks may reduce confusion (but I may overlook something): `^///*([^/]+)(/.*)$`.

I’m afraid; I have worked through the rules again, but I still have trouble reproducing the current test cases. It seems this latest change basically yields the same results.

ChristianGruen (changes requested):
* If the string matches `^///*([^/]+)?$`, the authority is the first match group and the string is empty.

I think this new rule contradicts with what you indicated in https://github.com/qt4cg/qtspecs/issues/566#issuecomment-2004241476 (didn’t you want to assign the value to the string, not the authority?):

>> Otherwise, if the string matches `^///*([^/]+)?(/.*)?$`, the authority is the first match group and the string is the second match group.
> That reduces the string to `/abc`. That becomes the path and the path segments are ('', 'abc').

Next, I wonder if it should be `^///*([^/]+)?$` or `^///*([^/]+)$` ?

* If the string matches `^///*([^/]+)?(/.*)?$`, the authority is the first match group and the string is the second match group.

Same here… Dropping the question marks may reduce confusion (but I may overlook something): `^///*([^/]+)(/.*)$`.

I’m afraid; I have worked through the rules again, but I still have trouble reproducing the current test cases. It seems this latest change basically yields the same results.

ndw:
Replying to @ChristianGruen 's comments above.

> * If the string matches ^///*([^/]+)?$, the authority is the first match group and the string is empty.
> 
> I think this new rule contradicts with what you indicated in https://github.com/qt4cg/qtspecs/issues/566#issuecomment-2004241476 (didn’t you want to assign the value to the string, not the authority?):

I don't see which comment you think contradicts the rule. But from first principles, I think the string `//example.com` should be parsed so that `example.com` is the authority and the string is empty. The number of leading slashes is hard to predict in the wild. I think this function assumes that `////////` is the same as `//`, so `/////example.com` is still an authority of `example.com` and an empty string. You could make the argument that `///example.com` is two slashes, an empty authority, and then the path `/example.com`, but I think that would be wrong more often than right.

(In principle, `///////example.com` is just invalid. A lot of the complexity here is trying to take a random string that someone's trying to parse as a URI and making the least surprising choices.)

> Next, I wonder if it should be ^///*([^/]+)?$ or ^///*([^/]+)$ ?
>
>    If the string matches ^///*([^/]+)?(/.*)?$, the authority is the first match group and the string is the second match group.
>
> Same here… Dropping the question marks may reduce confusion (but I may overlook something): ^///*([^/]+)(/.*)$.
>
> I’m afraid; I have worked through the rules again, but I still have trouble reproducing the current test cases. It seems this latest change basically yields the same results.

I'm sorry, I don't understand the question. But `^///*([^/]+)?(/.*)?$` is designed to produce the following results:

* `////` -> authority = `()`, string = `/`
* `////host` -> authority = `host`, string = `()`
* `////host/path` -> authority = `host`, string=`/path`
* `////host/` -> authority = `host`, string = `/`

Although, the way the rules are written, `///host` will already have matched item 3, so it won't be tested against item 4 and, on that basis, the `?` is redundant, I think. So:

4. If the string matches `^///*([^/]+)(/.*)?$`, the authority is the first match group and the string is the second match group.

would be fine.

ChristianGruen:
> I don't see which comment you think contradicts the rule.

I think it was: 

> That reduces the string to /abc. That becomes the path and the path segments are ('', 'abc').

But I need to say sorry; it’s been a while, and I may spend more time on this again. 

I’m not sure how we could proceed with the two functions without spending too much of our time. Maybe we should spend an hour together or something  and compare your implemention with the rules of the spec?

ndw:
I'm not sure how to proceed either. I should never have let this lie fallow for so long. Spilt milk.

I'm trying to do my two outstanding actions (as best I can) then make my implementation match what I have asserted are the correct answers, then compare them with the results you posted on 03-19 and see what the differences are.

I think we were converging last time we did this, fingers crossed.

ChristianGruen:
> I think we were converging last time we did this, fingers crossed.

Sounds good! And I hope my comment was not too disruptive (which was my impression when I read it again after having posted it).

ndw:
I'm withdrawing this in favor of #1244

PR #1108, opened 2024-03-22
Describe a less aggressive %-encoding for fn:build-uri

ndw:
My proposal seemed to meet with general approval, so here is my attempt to implement it in the spec.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -30084,12 +30084,43 @@ path with an explicit <code>file:</code> scheme.</p>
         is made to determine if a password or standard port are present,
         the <code>authority</code> value is simply added to the string.)</p>
 
-        <p>If the <code>path-segments</code> key exists in the map, then the
-        path is constructed from the parts, with non-URI characters encoded:
-        <code>string-join($parts?path-segments ! encode-for-uri(.), $options?path-separator)</code>,
-        otherwise the value of the <code>path</code> key is used.
-        If the <code>path</code> value is the empty sequence,
-        the empty string is used for the path. The path is added to the URI.</p>
+        <ulist>
+           <item><p>If the <code>path-segments</code> key exists in the map,
+           then the path is constructed from the segments.</p>
+           <p>To construct the path, each
+           segment is encoded, then they are joined together, separated by the path
+           separator, to form the path.
+           The encoding performed replaces any control characters (code points less than 0x20)
+           and exclusively the following characters with their
+           percent-escaped forms: <code> </code> (space) <code>%</code> (percent
+           sign), <code>/</code> (solidus), <code>?</code> (question mark),
+           <code>#</code> (number sign), <code>+</code> (plus sign),
+           <code>[</code> (left square bracket), and <code>]</code> (right
+           square bracket).</p>
+
+           <p>This is a compromise. The <code>fn:parse-uri</code> function removes
+           percent-escaping when it constructs the path segments because that
+           behavior is often most convenient when dealing with
+           common cases involving <code>file:</code> URIs. But the decoding process is not lossless (consider
+           that both “<code>%3D</code>” and “<code>=</code>” will be realized as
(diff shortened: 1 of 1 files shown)


ChristianGruen:
Thanks:

* Thanks to the changes, various test cases now succeed, whereas others fail (e..g `fn-build-uri-010b`, which expects `%5C` instead of `\` in the result). I guess this is expected?
* Will the proposed %-encoding only apply to path segments? If yes, some test cases need to be adjusted. For example, in [fn-build-uri-from-parse-020](https://github.com/qt4cg/qt4tests/blob/368142d44f5911e07b2e8e42634af5cd9360ec74/fn/build-uri.xml#L296-L314), the result would need to be `ldap://[2001:db8::7]/c=GB?=objectClass%3Fone` instead of `ldap://[2001:db8::7]/c=GB?=objectClass?one`
* Due to the proposed change for decoding fragments in [another pull request](https://github.com/qt4cg/qtspecs/pull/1098/files#diff-7625c07ae8131ff65c3caa677b188ed2b9b66237312d11c05a2fa2838c6f5c67R29221), I assume that the `fragment` part should now be encoded as well.

@ndw Do you think it’s better for me to wait until you’ve unified the spec, the test cases and your implementation, or do you benefit from interim feedback?

ChristianGruen (changes requested):
Thanks:

1. Thanks to the changes, various test cases now succeed, whereas others fail (e..g [fn-build-uri-010b](https://github.com/qt4cg/qt4tests/blob/368142d44f5911e07b2e8e42634af5cd9360ec74/fn/build-uri.xml#L662-L679), which expects `%5C` instead of `\` in the result). I guess this is expected?
2. Will the proposed %-encoding only apply to path segments? If yes, some test cases need to be adjusted. For example, in [fn-build-uri-from-parse-020](https://github.com/qt4cg/qt4tests/blob/368142d44f5911e07b2e8e42634af5cd9360ec74/fn/build-uri.xml#L296-L314), the result would need to be `ldap://[2001:db8::7]/c=GB?=objectClass%3Fone` instead of `ldap://[2001:db8::7]/c=GB?=objectClass?one`
3. Due to the proposed change for decoding fragments in [another pull request](https://github.com/qt4cg/qtspecs/pull/1098/files#diff-7625c07ae8131ff65c3caa677b188ed2b9b66237312d11c05a2fa2838c6f5c67R29221), I assume that the `fragment` part should now be encoded as well.

@ndw Do you think it’s better for me to wait until you’ve unified the spec, the test cases and your implementation, or do you benefit from interim feedback?

ndw:
I appreciate the interrim feedback. I thought I'd wait until we have at least a show of support for the PR before I commit updated test results.

ChristianGruen (changes requested):
I’ve moved my comment to the review section:

1. Due to the changes, various test cases now succeed, whereas others fail (e.g. [fn-build-uri-010b](https://github.com/qt4cg/qt4tests/blob/368142d44f5911e07b2e8e42634af5cd9360ec74/fn/build-uri.xml#L662-L679), which expects `%5C` instead of `\` in the result). I guess this is expected?
2. Will the proposed %-encoding only apply to path segments? If yes, some test cases need to be adjusted. For example, in [fn-build-uri-from-parse-020](https://github.com/qt4cg/qt4tests/blob/368142d44f5911e07b2e8e42634af5cd9360ec74/fn/build-uri.xml#L296-L314), the result would need to be `ldap://[2001:db8::7]/c=GB?=objectClass%3Fone` instead of `ldap://[2001:db8::7]/c=GB?=objectClass?one`
3. Due to the proposed change for decoding fragments in [another pull request](https://github.com/qt4cg/qtspecs/pull/1098/files#diff-7625c07ae8131ff65c3caa677b188ed2b9b66237312d11c05a2fa2838c6f5c67R29221), I assume that the `fragment` part should now be encoded as well.

ndw:
The CG agreed to merge this PR at [meeting 079](https://qt4cg.org/meeting/minutes/2024/05-28.html).

Issue #1110, opened 2024-03-23
New error codes


ChristianGruen:
The XQFO spec includes various “[TODO: error code]” comments. Should we add error codes when finalizing PRs, or does a master plan exist to add them at the very end?

michaelhkay:
Best to add them now as they need to appear in tests.

ChristianGruen:
It seems there are only 3 codes missing at the moment. Feel free to correct my suggestions (I’ve never understood the semantics behind the error codes…).

* [fn:hash](https://qt4cg.org/specifications/xpath-functions-40/Overview-diff.html#func-hash): `FOHA0001` ?
* [fn:op](https://qt4cg.org/specifications/xpath-functions-40/Overview-diff.html#func-op): `FOOP0001` ?
* [XQuery, Map Test](https://qt4cg.org/specifications/xquery-40/xquery-40-diff.html#id-map-test): …

michaelhkay:
>I’ve never understood the semantics behind the error codes

Not so much semantics, as culture and tradition: they were introduced to the spec by an IBM old-timer, for whom 8 character alphanumeric codes are the instinctively natural way of distinguishing error conditions.

ChristianGruen:
> The only tricky design decision is whether to allocate a new code or re-use an existing one. The current spec offers very little guidance on that. The key question to ask yourself is: if an application wanted to catch this error, how specific would it want it to be?

True. A custom error code for `fn:hash` seems appropriate; for `fn:op`, an existing code should work as well.

Newer EXPath modules use codes like `file:is-dir`, which are probably more intuitive, but I guess it’s too late to initiate that for XQFO as long as we don’t offer aliases for the existing codes.

PR #1112, opened 2024-03-25
1110-partial New error codes

ChristianGruen:
Issue: #1110.

* [fn:hash](https://qt4cg.org/specifications/xpath-functions-40/Overview-diff.html#func-hash): I added `FOHA0001` as error code.
* [fn:op](https://qt4cg.org/specifications/xpath-functions-40/Overview-diff.html#func-op): I used `XPTY0004` as error code, as the allowed operators could also be defined as string enumeration.
* [XQuery, Map Test](https://qt4cg.org/specifications/xquery-40/xquery-40-diff.html#id-map-test): Not included in this PR.


ndw:
The CG agreed to merge this PR without discussion at [meeting 072](https://qt4cg.org/meeting/minutes/2024/04-09.html).

PR #1141, opened 2024-04-10
1140 Replace 'search' with 'target' for indexing functions

### LINKED ISSUE #1140: Use $target instead of $search for indexing functions
Back in February, when we discussed `array:index-of`, DN observed that the argument name `$search` could be interpreted as performing some sort of action. The alternative `$target` was proposed instead as being more "noun like".

ndw:
Close #1140 

Having created the issue, per my outstanding action, I thought I'd take a quick look to see how extensive the change would be. AFAICT (though I confess to not looking exceedingly carefully), only two functions are effected. Here, for your consideration, is a PR that resolves the issue.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -12378,7 +12378,7 @@ let $newi := $o/tool</eg>
       <fos:signatures>
          <fos:proto name="index-of" return-type="xs:integer*">
             <fos:arg name="input" type="xs:anyAtomicType*"/>
-            <fos:arg name="search" type="xs:anyAtomicType"/>
+            <fos:arg name="target" type="xs:anyAtomicType"/>
             <fos:arg name="collation" type="xs:string?" default="fn:default-collation()"/>
          </fos:proto>
       </fos:signatures>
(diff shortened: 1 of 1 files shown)


ChristianGruen (commented):
`$target` sounds a bit like we are addressing something that is going to change. But I’m not a native one, so… just fine ;)

michaelhkay:
The only noun I could come up meaning "the thing you are looking for" was "$quest". That might be considered a bit esoteric. I think `$target` is preferable to `$search` but I'm not wild about either.

ndw:
@LinguaCelta suggests `$term` as a possible alternative.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -17149,7 +17149,7 @@ return tokenize(normalize-space($s), ' ')[. castable as xs:IDREF]</eg>
          <p>There is no guarantee that a generated unique identifier will be distinct from any
             unique IDs specified in the source document.</p>
          <p>There is no inverse to this function; it is not directly possible to find the node with
-            a given generated ID. Of course, it is possible to search a given sequence of nodes
+            a given generated ID. Of course, it is possible to target a given sequence of nodes
```
benibela, line 17152: was that sentence intended?

benibela:
How about `$needle` ?

specifications/xpath-functions-40/src/function-catalog.xml
ndw: No. Oops.

ndw:
No, I think `$needle` only works if you're familiar with the idiom to "search for a needle in a haystack". And even then, it's so out of context that I think it would require more explanation rather than less!

ChristianGruen:
Maybe `$item`… Whatever ;·)

Arithmeticus:
`$value` and `$item` are names used in comparable C# functions I've used.

ndw:
The CG agreed to merge this PR at [meeting 073](https://qt4cg.org/meeting/minutes/2024/04-16.html).

PR #1149, opened 2024-04-12
1084 Add fos:default-description to support prose descriptions of defaults

### LINKED ISSUE #1084: Incorrect rendition of option defaults
In the F&O spec, when rendering the default value of an option, code font is being used for narrative prose: see for example defaults for the `delivery-format` and `base-output-uri` options of `fn:transform`

ndw:
Close #1084 

This won't render correctly in the PR, but hopefully the diff is clear enough to decide if this is the approach we want to take.


ndw:
The CG agreed to merge this PR without further discussion at [meeting 073](https://qt4cg.org/meeting/minutes/2024/04-16.html).

Issue #1179, opened 2024-04-30
Editorial: `array:values`, `map:values`


ChristianGruen:
Triggered by #1087:

1. Align `map:values` with `array:values`
2. Revise notes.
3. Rename the functions.

@michaelhkay You suggested that `content` might be a better name – rather than `item` – for retrieving the sequence-concatenation of values in map lookups in https://github.com/qt4cg/qtspecs/issues/1169#issuecomment-2074378446. Should we rename the functions to `array:contents`, `map:contents` or use `array:items` and `map:items`?

michaelhkay:
"content" does seem a better name than "values". The current name "values" suggests that it simply returns the values that are already present, without any hint of a flattening operation.

PR #1185, opened 2024-05-02
1179 array:values, map:values → array:get, map:get

ChristianGruen:
Issue: #1179

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -18961,7 +18961,7 @@ function-annotations(local:inc#1)</eg></fos:expression>
                <fos:expression><eg>
 let $old := %local:deprecated('0.1', '0.2') fn() {}
 let $ann := function-annotations($old)
-return map:keys($ann) || ': ' || string-join(map:contents($ann), ', ')
+return map:keys($ann) || ': ' || string-join(map:get($ann), ', ')
 </eg></fos:expression>
                <fos:result>local:deprecated: 0.1, 0.2</fos:result>
             </fos:test>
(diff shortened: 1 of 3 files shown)


michaelhkay:
I don't feel 100% comfortable with this: particularly for maps. It feels wrong that `map:contains(M, X)` is true if the map contains the key X, while map:contents(M) returns the "values". Surely any object should contain its contents?

I also feel that we're trying to impose a symmetry between maps and arrays where possibly none exists.

There's an underlying problem with over-use of the word value. A map is a set of entries containing a key and a value, in which both the key and the value are values. I would suggest we try and solve that problem first.

If we do want to treat arrays and maps symmetrically, then we should treat an array as if it were a map with consecutive, ordered, integer-valued keys. On that reasoning, perhaps:

* Both arrays and maps contain entries
* Entries comprise a key and a member
* Both the key and the member are values

And as regards map:values() and array:values(), perhaps we should ask whether we want these functions at all? They are useful in the common case where the members/values are single items, but in all other cases the behaviour is a potential trap for the unwary. Perhaps a name that warns of the dangers would be better: like `members-as-sequence()`?

ChristianGruen:
> I don't feel 100% comfortable with this: particularly for maps. It feels wrong that `map:contains(M, X)` is true if the map contains the key X, while map:contents(M) returns the "values". Surely any object should contain its contents?
>
> I also feel that we're trying to impose a symmetry between maps and arrays where possibly none exists.

I somewhat agree. The symmetry exists – due to the `MAP-OR-ARRAY?KEY` syntax – but we still haven’t found a good name for the result. The intent of this PR was mainly to align the wording with the new lookup modifier syntax, which currently uses `items` for returning values of maps (and which may use `content` instead, as discussed). If we don’t want a term here, we could discard the explicit keyword specifier for the 3.1 lookup behavior.

> There's an underlying problem with over-use of the word value. A map is a set of entries containing a key and a value, in which both the key and the value are values. I would suggest we try and solve that problem first.

The key/value term combination is very popular for maps in general (and associated arrays, etc.), but it’s true that this doesn’t simplify our work. Next, “map values” were introduced with version 3.1 of the languages, so many documentations and publications refer to it.

> And as regards map:values() and array:values(), perhaps we should ask whether we want these functions at all?

I would really like to keep those, in one way or the other. There have been repeated requests in the past for an equivalent functional solution to the `$xyz?*` syntax (and I have seen intuitive, but wrong code like `$xyz('*')`).

benibela:
>I also feel that we're trying to impose a symmetry between maps and arrays where possibly none exists.

In a theoretical sense, an array is just a map with integer keys

michaelhkay:
Following the discussion yesterday, I'd like to suggest for arrays:

```
array:to-sequence()
```
I think this expresses what the function does fairly clearly and intuitively, and signals that with a non-trivial array, you are going to lose some structure.

For maps, it's less obvious; and indeed, it's less obvious that the function is useful.

So I'm going to suggest that map:values should return the KVP values as an array, and the existing function is then achieved as map:values() => array:to-sequence().

This gives us two functions whose effect, I think, is reasonably clear from their names.

ChristianGruen:
I'm torn. I don't think people would use `$map => map:values() => array:to-sequence()` as an alternative for `$map?*`.

As this issue was initiated by the introduction of the enhanced lookup syntax, we should keep both in mind here: How would you rename the current `items` specifier?

ChristianGruen:
My current favorite would be to drop the two extra functions and generalize `array:get` and `map:get` by making the position argument optional. The advantages I see: 

* We don’t need to discuss terminology.
* As `$map?x` is equivalent to `$map => map:get('x')`,`$map?*` and `$map => map:get()` look closely related.
* The `get` functions already return a sequence, which means they cannot be adapted in the future to return multiple results in an array.

A drawback might be that the functions couldn’t be enhanced anymore to accept a sequence of positions (unless we treat `map:get($map)` and `map:get($map, ())` differently, which we could do, as other functions do it). But I believe we have already decided that we don’t want to do this anyway.

ChristianGruen:
If no one thinks it’s a bad idea, I’ll incorporate my proposal to drop the two functions `array:values` and `map:values` and incorporate the functionality in `array:get` and `map:get`.

ChristianGruen:
Now ready for review.

ChristianGruen:
Should be revised once we have a solution for https://github.com/qt4cg/qtspecs/issues/1338.

ChristianGruen:
Dropped for now.

PR #1220, opened 2024-05-15
73 copy&paste typo in fn:graphemes (combining diaeresis should be ZWJ)

gimsieke:
I wonder which font I need to use in order to see U+1F476 U+200D U+1F6D1 as a single grapheme. When I naively put the character sequence into an HTML page, the two glyphs will be rendered individually. How can we make sure that the grapheme will be rendered as intended for most people when they read the spec?

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -28949,7 +28949,7 @@ return every($dl/*, fn($elem, $pos) {
             <fos:test>
                <fos:expression>graphemes(char(0x1F476) || char(0x200D) 
 || char(0x1F6D1))</fos:expression>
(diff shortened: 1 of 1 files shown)


ChristianGruen (approved):

ndw:
Thank you @gimsieke

Arithmeticus:
Thanks for attending to the typo, @gimsieke . One of the challenges in explaining how `fn:graphemes` works with control/nonprinting characters is that we can't really render them visually. The zero-width joiner is invisible, and the two characters on either side will look just like a pair of adjacent characters. Hence the example does anything but show a rendering. Here's how Unicode tries to make the example more apparent: https://www.unicode.org/Public/15.1.0/ucd/auxiliary/GraphemeBreakTest.html#samples

ndw:
The CG agreed to merge this PR without discussion at [meeting 078](https://qt4cg.org/meeting/minutes/2024/05-21.html).

PR #1229, opened 2024-05-20
Rework record descriptions per ACTION QT4CG-070-01

ndw:
I think this PR completes the work I started before and that we reviewed briefly some time ago.

Note: because this PR involves changes to the schemas and the stylesheets, the PR build is going to be...funky. I don't know if the diffs will be useful either. There are no (intentional) technical changes in this PR. You can preview the built result by look at https://qt4cgtest.nwalsh.com/branch/action-qt4cg-70-01/xpath-functions-40/Overview.html

A few notes:

1. Instead of just reusing `fos:options`, I added `fos:record-description` with a very similar content model.
2. I couldn't completely remove some of the `<fos:type>` elements defined in the globals section at the top of the `function-catalog.xml` because it causes IDREF failures when that file is parsed.
3. I tinkered a bit with the styling, and moved a bunch of inline table styles into CSS

### DIFF
--- a/specifications/css/xpath-functions-40.css
+++ b/specifications/css/xpath-functions-40.css
@@ -74,15 +74,15 @@ div.proto        {
                  }
 
 div.record        { 
-    padding: .5em;
-		border: .5em;
-		border-left-style: solid;
-		page-break-inside: avoid;
-		margin: 1em auto;
-		border-color: #ffaaff;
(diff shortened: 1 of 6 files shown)


ndw:
The CG agreed to merge this PR at [meeting 078](https://qt4cg.org/meeting/minutes/2024/05-21.html).

Issue #1193, opened 2024-05-07
Parsing Functions: Empty input


ChristianGruen:
I was asked why some of the parsing functions allow empty input and others don’t:

Function | Input
--- | ---
`fn:parse-integer` | `xs:string`
`fn:parse-uri` | `xs:string`
`fn:parse-ietf-date` | `xs:string?`
`fn:parse-QName` | `xs:string`
`fn:parse-xml` | `xs:string?`
`fn:parse-xml-fragment` | `xs:string?`
`fn:parse-html` | `(xs:string \| xs:hexBinary \| xs:base64Binary)?`
`fn:parse-json` | `xs:string?`
`fn:parse-csv` | `xs:string?`

I would assume there is no rationale behind this, and that we should always allow empty input.

michaelhkay:
Generally the specs are a pretty inconsistent mess as regards handling of empty sequence on input; it goes well beyond the parse-XX functions.

ChristianGruen:
In the particular case, it seems that all pre-4.0 functions allow empty sequences, whereas only some of the new functions do. But it’s true, when cleaning this up, we should probably consider the complete function set.

michaelhkay:
See also #748

PR #1231, opened 2024-05-21
1193 Parsing Functions: Empty input

ChristianGruen:
Issue: #1193 (covers only the obvious 4.0 inconsistencies)


ChristianGruen:
In addition (@ndw), I’ve renamed the input argument of `fn:parse-uri` from `$uri` to `$value`, in alignment with `fn:parse-ietf-date`, `fn:parse-QName` and other functions.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -29994,8 +29996,8 @@ some $boolean in (
 
    <fos:function name="parse-uri" prefix="fn" diff="add">
       <fos:signatures>
-         <fos:proto name="parse-uri" return-type-ref="uri-structure-record">
-            <fos:arg name="uri" type="xs:string"/>
+         <fos:proto name="parse-uri" return-type-ref="uri-structure-record?">
```
ChristianGruen, line 29999: @ndw Sorry, so many questions… The current syntax doesn’t seem to be allowed yet. How can I write down optional record types?

ndw: I'll have to investigate. What you've done is sort of what I'd have expected to work.

ChristianGruen:
@ndw Just a reminder: I think that’s something I cannot fix by myself. And I also think you’ll be busy enough with other things…

ndw:
I'll try to look next week. (I'll have a few hours each morning before Balisage begins.) There's almost no chance I can look at this before then.

ndw:
The markup is apparently:
```
 <fos:proto name="parse-uri" return-type-ref="uri-structure-record" return-type-ref-occurs="?">
```
The `return-type-ref` attribute is an `IDREF` so you have to specify the occurrence indicator separately. :-(

There was an inconsistency between the DTD and the XML Schema that prevented this markup from working. I've fixed that, so if you rebase this commit off of the current head, you should be able to use this markup.

Apologies for the delay.

ChristianGruen:
@ndw Great; that did it.

image

PR #1233, opened 2024-05-22
517 Major edits to fn:chain, clarification only

Arithmeticus:
@dnovatchev For your review. The current version of `fn:chain` in the specs has unnecessary verbiage, is out of sync with other comparable spec entries, and has passages that cry out for clarity and concision. I do not think I have made any edits that change any substantive points. A summary:

* Summary simplified
* Rules: prose for the recursive process brought into conformity with other functions in the specs that describe recursive processes.
* Error conditions: two entries, and clarification of the conditions when they are triggered.
* Former note 1 deleted (excessively wordy, introduces unfamiliar or unnecessary terms/concepts).
* Former notes 3 and 4 moved to rules, distilled into more concise prose.
* Former note 5 (now 2) recast to convey what I think was your original intent, and allows the reader to more quickly compare `fn:chain` to two comparable methods.

Signature, properties, XPath rule, and examples are unchanged.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -19439,16 +19439,16 @@ return fold-right(
                   to the following rules. <olist>
                      <item>
                         <p>If <code>$f</code> has arity 1, let <code>$input-array</code> be all
-                           items in <code>$input</code> wrapped in an array.</p>
+                           items of <code>$input</code> wrapped in an array.</p>
                      </item>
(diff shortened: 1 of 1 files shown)


Arithmeticus:
Also, @dnovatchev if it wasn't fully clear from the previous note, these are all friendly amendments. I'm trying to act as both your advocate and that of the reader.

ndw:
The CG agreed to merge this without discussion at the Prague f2f.

PR #1244, opened 2024-05-28
566-partial Rewrite parse-uri

ndw:
What happened was, I discovered I'd messed with the code in two different branches.

In the course of trying to straighten that out, I came to the conclusion that the *real* difficulty in parsing URIs is what to do about all the special cases around `file:`. And that the approach I'd taken in the previous draft was unnecessarily complicated.

I think this is better.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -30071,7 +30071,8 @@ some $boolean in (
          present. Some portions of the URI are identified by matching
          with a regular expression. This approach is designed to make
          the description clear and unambiguous; it is not implementation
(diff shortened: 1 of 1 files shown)


ChristianGruen (approved):
Looks good in general.

I’ll try to reimplement the latest version of the spec, but I think I won’t make it before beginning of July.

ndw:
Okay. I've now pushed updated tests to the tests repository. We can come back to this after the face-to-face and see if we want to merge it or wait until you've implemented it.

ndw:
The CG agreed to merge this PR at [meeting 086](https://qt4cg.org/meeting/minutes/2024/07-16.html).

Issue #1281, opened 2024-06-17
invisible-xml() return type


michaelhkay:
It has been pointed out that the return type of `fn:invisible-xml()`, currently `fn(xs:string) as item()`, could be more precisely given as `fn(xs:string) as document-node()`.

ndw:
Yes, that's my `QT4CG-081-04: NW to update the function signature of fn:invisible-xml` I believe.

michaelhkay:
Fixed by #1282

PR #1282, opened 2024-06-17
Revise fn:invisible-xml

### LINKED ISSUE #991: Invisible-xml - missing details
The spec for invisible-xml doesn't say whether the parsing function returns an element node or a document node.

It should also say, for completeness, that the parsing function is "nondeterministic with respect to node identity" - that is, if you parse the same input twice, its undefined whether you get the same node twice, or different nodes.

ndw:
1. Resolve `QT4CG-080-04`: NW to revise p:invisible-xml, fix #991 
2. Resolve `QT4CG-081-04`: NW to update the function signature of fn:invisible-xml
3. Resolve `QT4CG-081-04`: NW to describe why the grammar option can be empty on fn:invisible-xml

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -31426,6 +31426,10 @@ let $scan-right := function(
          representation that indicates an error occurred and may
          provide additional error information.</p>
 
+         <p>If the function is called twice with the same arguments, it is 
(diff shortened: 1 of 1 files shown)


ndw:
Fix #1281

ndw:
The CG agreed to merge this PR at [meeting 083](https://qt4cg.org/meeting/minutes/2024/06-25.html).

PR #1297, opened 2024-06-24
Minor correction to fn:scan-right  - typo

dnovatchev:
In the first example the expression was:

```
scan-right(1 to 5, 0, op('+'))
```

but must be:

```
scan-right(1 to 10, 0, op('+'))
```


ndw:
This PR includes not just the scan-right typo, but also the BLAKE3 changes and perhaps others. Can you resubmit this as a clean PR that makes only the scan-right change, please?

dnovatchev:
Closed as this incidentally contains commits for other issues.
Will resubmit again.

Thanks Norm for catching this

PR #1358, opened 2024-08-01
959 fn:unix-time

### LINKED ISSUE #959: Milliseconds ↔ xs:dayTimeDuration, Unix time ↔ xs:dateTime
We should extend the constructor functions to convert integers (millisecond, and the Unix time, starting from `1970-01-01T00:00:00Z`) to `xs:dayTimeDuration` and `xs:dateTime` instances…

```xquery
xs:dateTime(12345),
xs:dayTimeDuration(12345)
```

…and it should be possible to convert the values back to integers.

Related: https://docs.basex.org/wiki/Conversion_Module#Dates_and_Durations

ChristianGruen:
Closes #959

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -8541,9 +8541,9 @@ xs:dayTimeDuration($arg2) div xs:dayTimeDuration('PT1S')
          <p>Returns a dateTime value for a Unix time.</p>
       </fos:summary>
       <fos:rules>
-         <p>The function returns a dateTime value for the Unix time specified by <code>$value</code>
-            in milliseconds. If the value is absent or an empty sequence, <code>0</code> is used.
-            The Unix time is defined in <bibref ref="ieee1003.1-2024"/>.</p>
+         <p>The function returns a dateTime value in UTC timezone for the Unix time specified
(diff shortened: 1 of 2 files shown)


michaelhkay:
I suggest following current-dateTime() by declaring the result type as xs:dateTimeStamp and including the paragraph

If the implementation supports data types from XSD 1.1 then the returned value will be an instance of xs:dateTimeStamp. Otherwise, the only guarantees are that it will be an instance of xs:dateTime and will have a timezone component.

michaelhkay:
Perhaps there should be a citation to something like https://pubs.opengroup.org/onlinepubs/9699919799/functions/time.html that gives an official definition

Are you sure that the supplied calculation handles leap seconds correctly?

ChristianGruen:
I'll be glad to incorporate the suggestions.

> Are you sure that the supplied calculation handles leap seconds correctly?

As far as I remember, Unix time does not account for leap seconds; it assumes that every day has 86,400 seconds. So I would say that the presented calculation is correct, but I could add a note on this.

michaelhkay (changes requested):
The examples suggest the intent that the returned value will always be in UTC timezone, but this is not stated in the rules.

ChristianGruen:
> The examples suggest the intent that the returned value will always be in UTC timezone, but this is not stated in the rules.

:+1: I’ve added UTC timezone in the rules.

PR #1380, opened 2024-08-12
1320 Attempt to resolve a bug in parse-uri

### LINKED ISSUE #1320: fn:parse-uri: Observations
@ndw I decided to only give feedback on the first test that fails; maybe it makes things easier. Next, I chose this repository (instead of `qt4tests`), as you are the better person to judge if bugs are to be fixed in the tests or the spec.

The test case [fn-parse-uri-012](https://github.com/qt4cg/qt4tests/blob/5b0750b035d336040b4944dba6225e50bd004839/fn/parse-uri.xml#L239-L253) uses `fn:parse-uri("file:///c:/path/to/file")` and returns:

```xquery
{
  "uri": "file:///c:/path/to/file",
  "scheme": "file",
  "hierarchical": true(),
  "path": "/c:/path/to/file",
  "filepath": "c:/path/to/file",
  "path-segments": ("", "c:", "path", "to", "file")
}
```

Following the current rules, I would have expected `filepath` to be `/c:/path/to/file`:

```
If URI matches `^([a-zA-Z][A-Za-z0-9\+\-\.]*):(.*)$`:
• SCHEME: file
• STRING: ///c:/path/to/file

If the scheme is known to be file and the string matches "^/*(/[a-zA-Z][:|].*)$":
• STRING: /c:/path/to/file

If the scheme is file or the empty sequence, and filepath is the empty sequence, filepath is also the whole string:
• FILEPATH: /c:/path/to/file
```

Would you propose to revise the test or the specification?

ndw:
Fix #1320 

(A speculative PR in the hope that my proposed resolution is acceptable.)

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -30570,6 +30570,8 @@ fold-left($input, false(), fn($result, $item, $pos) {
          the <emph>string</emph> is the first match group and the
          <emph>query</emph> is the second match group. Otherwise,
          the string is unchanged and the <emph>query</emph> is the empty
+         sequence. If the query is the empty
+         string, it is discarded and the <emph>query</emph> is the empty
          sequence.</p>
             </item>
          </ulist>
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -30650,6 +30650,11 @@ fold-left($input, false(), fn($result, $item, $pos) {
                      and <emph>string</emph> are both the first match group.</p>
                   </item>
                   <item>
+                     <p>If the <emph>string</emph> begins <code>^//*[A-Za-z]:/</code> then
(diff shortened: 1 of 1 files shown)


specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -30649,6 +30649,11 @@ fold-left($input, false(), fn($result, $item, $pos) {
                      matches <code>^/*(//[^/].*)$</code>: then <emph>filepath</emph>,
                      and <emph>string</emph> are both the first match group.</p>
                   </item>
+                  <item>
+                     <p>If the <emph>string</emph> begins <code>^//*[A-Za-z]:/</code> then
```
ChristianGruen, line 30653: It may be clearer to write:

```suggestion
                     <p>Otherwise, if the <emph>string</emph> matches <code>^/+[A-Za-z]:/.*$</code> then
```

ChristianGruen (commented):
@ndw Thanks, this is a huge step forward.

PR #1388, opened 2024-08-14
1387 Clarify the encoding rules

### LINKED ISSUE #1387: fn:build-uri: Observations
@ndw I’ll again start with a single test case, [fn-build-uri-from-parse-020](https://github.com/qt4cg/qt4tests/blob/b5c9431b6e5c7e24729a4de19aac1754b0b34c4a/fn/build-uri.xml#L296-L314):

```xml
<test-case name="fn-build-uri-from-parse-020">
    <description>Builds an example from the specification</description>
    <created by="Norm Tovey-Walsh" on="2023-03-10"/>
    <test>fn:build-uri(map {
  "uri": "ldap://[2001:db8::7]/c=GB?objectClass?one",
  "scheme": "ldap",
  "authority": "[2001:db8::7]",
  "host": "[2001:db8::7]",
  "path": "/c=GB",
  "query": "objectClass?one",
  "query-parameters": map {
    "": "objectClass?one"
  },
  "path-segments": ("", "c=GB")
})</test>
    <result>
      <assert-eq>"ldap://[2001:db8::7]/c=GB?objectClass?one"</assert-eq>
    </result>
  </test-case>
```

I would expect `ldap://[2001:db8::7]/c=GB?=objectClass%3Fone` as result, as the spec says…

> If the `query-parameters` key exists in the map, its value must be a map. A sequence of strings is constructed from the values in the map. For each key and each *value* associated with that *key* in turn:
> * If the *key* is the empty string, the string constructed is the *value* encoded with `encode-for-uri`.

…and `encode-for-uri('?')` gives us `%3F`. How do you handle that in your code?

ndw:
I decided to put this in a separte PR so that it can be reviewed in isolation. I don't expect it to be controversial, exactly, but it may require adjustment.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -31389,7 +31389,7 @@ description of <code>fn:resolve-uri</code>.</p>
 <p>This example uses an invalid query separator so raises an error.</p>
 <fos:test>
 <fos:expression><eg>parse-uri("https://example.com:8080/path?s=%22hello world%22;;sort=relevance", { "query-separator": ";;" } )</eg></fos:expression>
-<fos:error-result error-code="FOXX0000"/>
+<fos:error-result error-code="FOUR0002"/>
 </fos:test>
 </fos:example>
 
@@ -31560,9 +31560,28 @@ path with an explicit <code>file:</code> scheme.</p>
         is made to determine if a password or standard port are present,
         the <code>authority</code> value is simply added to the string.)</p>
 
+        <p>The <code>fn:parse-uri</code> function removes
+        percent-escaping when it constructs the
+        <code>path-segments</code>, <code>query-parameters</code>, and
+        <code>fragment</code> properties. That’s often the most
+        convenient behavior but, in order to reconstruct a URI from them,
+        special escaping rules apply. These rules protect delimiters
+        without encoding additional characters unnecessarily.
+        The rules for <code>path-segments</code>,
+        <code>query-parameters</code>, and <code>fragment</code>s are
+        slightly different because the URI encoding conventions are
+        slightly different in each case.</p>
+
+        <p>An application with more stringent requirements can
+        construct a <code>path</code> or <code>query</code> that
+        satisfies the requirements and leave the
+        <code>path-segments</code> and/or
+        <code>query-parameters</code> keys out of the map.</p>
+
         <ulist>
            <item><p>If the <code>path-segments</code> key exists in the map,
            then the path is constructed from the segments.</p>
+
            <p>To construct the path, each
            segment is encoded, then they are joined together, separated by the path
            separator, to form the path.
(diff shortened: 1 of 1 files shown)


ChristianGruen (commented):
I wonder what could be done to simplify the rules a little:

1. I would really appreciate if we had fixed set of characters that are encoded. Would you mind dropping the `path-separator` and `query-separator` options, or do you think they are essential?
2. If we need to keep them, I assume the `path-separator` would need to be considered as well when encoding path segments. In addition, we’ll have to write more tests.
3. Maybe the lists of characters to be encoded could be represented as simple strings (regex or literals).

ChristianGruen:
> I'm really sympathetic to trying to simplify the set of characters but it's tricky because the URI spec/user expectations intersection is different in the different cases. I'll see if I can think of a simpler presentation.

That would be great. I’ve been thinking about the time we’ve already spent on those two functions, and with every new rule I wonder if we’ll manage to build something that’s going to be solid (unfortunately, we don’t have a third implementer around who might encounter things that we both haven’t noticed yet). Removing the convenience options might improve the situation: If someone really needs more advanced features to build URIs, it’s fairly easy to write some custom code.

Either way, URIs are tough…

ChristianGruen (changes requested):
I took the liberty of marking this with “request changes”. I think (as proposed) we should either simplify the rules again or consider `path-separator`, too.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -31478,40 +31483,55 @@ path with an explicit <code>file:</code> scheme.</p>
...
-
         <p>The path is added to the URI.</p>
 
         <p>If the <code>query-parameters</code> key exists in the map, its value
-        must be a map. A sequence of strings is constructed from the values in the map.
+        must be a map. A sequence of strings is constructed from the values in the map.</p>
+
+        <p>To construct the string, each <emph>key</emph> and <emph>value</emph>
+        is encoded.
+        The encoding performed replaces any control characters (code points less than 0x20)
```
Arithmeticus, line 31493: code points > codepoints

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -31478,40 +31483,55 @@ path with an explicit <code>file:</code> scheme.</p>
...
+        <code>query</code> key.</p>
+        <p>If there is a query, it is added to the URI with
+        a preceding <code>?</code> (question mark).</p>
 
         <p>If the <code>fragment</code> key exists in the map, then
-        the value of that key is encoded with <code>encode-for-uri</code> and added to the URI with
-        a preceding hash mark (<code>#</code>).</p>
+        the value of that key is encoded and added to the URI with a
+        preceding hash mark (<code>#</code>).
+        The encoding performed for the fragment replaces any control characters (code points less than 0x20)
```
Arithmeticus, line 31526: code points > codepoints

ndw:
The CG agreed to close this PR without further action at [meeting 089](https://qt4cg.org/meeting/minutes/2024/09-10.html)

PR #1405, opened 2024-08-28
1404 Change fn:invisible-xml grammar parameter to xs:string?

### LINKED ISSUE #1404: Invisible-xml function: why is $grammar of type item()?
In the `invisible-xml` function, the $grammar argument has type `item()?`. There's no evident reason why it isn't `xs:string?`; there's no clue in the specification or example what it might mean to supply something that isn't a string (or an empty sequence).

Note that declaring it as `item()` rather than `xs:string?` suppresses atomization of the supplied value.

ndw:


ChristianGruen (approved):

ndw:
In private conversation, @michaelhkay observed that it would be possible to go further with the function signature to
```
<fos:arg name="grammar" type="(xs:string | document-node(element(ixml)) | element(ixml))?" default="()"/>
```
I don't know if that's more useful than distracting or not.

There's also the broader question of whether to allow both document and element. I don't honestly like having both, but I think it would be a mistake to remove either, so...

ChristianGruen:
> There's also the broader question of whether to allow both document and element

I would prefer to only have one of them. Otherwise, users may be confused what they will get.

ndw:
Okay, I think that's enough pressure in that direction. If you've got an XML version of an iXML grammar that you loaded or constructed in a document, I guess `grammar="$doc/ixml"` is simpler than allowing `document-node()`. 

Beware, however, that in XQuery, if the default namespace is defined to be something other than the empty string, you'll have to work a little harder. I'd have put the `ixml` grammar in a namespace, but that ship has sailed.

michaelhkay (approved):

PR #1412, opened 2024-09-02
Fix typo in uri-structure-record

### LINKED ISSUE #1411: uri-structure-record gives type of path-segments property as xs:string?, not xs:string*
The current spec [defines `uri-structure-record`](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#uri-structure-record) as having a
```
path-segments? 	as xs:string?
```
property, which doesn't make sense and also contradicts the example given later:
> The unescaped form is easily accessible from path-segments:
> ```
> ("", "path", "to", "a/b")
> ```
It should be `xs:string*`.

ndw:
Fix #1411


ndw:
The CG agreed to merge this PR without discussion at [meeting 089](https://qt4cg.org/meeting/minutes/2024/09-10.html)

PR #1413, opened 2024-09-02
Dispose of action QT4CG-080-05, add absolute to parse-uri

ndw:
This PR disposes of my action to add `absolute` to the output of `fn:parse-uri`.

The rules that define an absolute URI: "has a scheme", "does not have a fragment identifier", "is a hierarchical scheme" don't fit neatly into a single place, so there are two different places where the setting is considered.

Because of this, I think it's simpler to set `absolute` to `true` when the URI is absolute, but not to `false` when it isn't. That could be rectified with a bit more prose.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -30591,6 +30591,10 @@ fold-left($input, false(), fn($result, $item, $pos) {
             </item>
          </ulist>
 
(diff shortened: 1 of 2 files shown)


ChristianGruen (approved):

ndw:
The CG agreed to merge this PR without discussion at [meeting 089](https://qt4cg.org/meeting/minutes/2024/09-10.html)

PR #1414, opened 2024-09-03
XSLT spec abstract, introduction

Arithmeticus:
As currently written, preliminary parts of the the XSLT specs are out of sync with other specs, or have inconsistencies.

* The section called Abstract mostly describes what's different in 4.0. That despite a section shortly later specially devoted to the topic.
* The section called Abstract does not lead with its best foot forward: the language does a _lot_ more than XML-to-XML transforms.
* The section called "What is XSLT?" never answers the question posed in the header. 
* The section called "What is XSLT?" begins with a sentence that both repeats earlier material and is not germane to the header.

In the attached PR, I have proposed moving several paragraphs. I have not intentionally edited them. I have also taken a stab at defining XSLT. I have tried to calibrate the Abstract so that it is both sparing, in the spirit of the XQFO abstract.

I don't think that what I've put in here is any way finalized. But I hope it is a step in the right direction, and a good basis for further refinement.


michaelhkay (approved):

ndw:
The CG agreed to merge this issue without discussion at [meeting 090](https://qt4cg.org/meeting/minutes/2024/09-17.html).

PR #1423, opened 2024-09-04
1387b Clarify parse-uri/build-uri encoding rules, and remove options

ndw:
This is a slightly more radical alternative to #1388 

In addition to clarifying the encoding rules (maintaining slightly different rules for path segments, query parameters, and fragment identifiers), it removes the `path-separator` and `query-separator` options.


ChristianGruen:
Thanks, Norm. I haven't implemented it yet, but I am convinced it improves the status quo, and I would be happy to see it merged (without discussion unless someone wants to add something).

PR #1429, opened 2024-09-08
1403 Align type tests

### LINKED ISSUE #1403: Align AnyMapTest, AnyArrayTest and with ElementTest
I always wondered why ElementTests for _any element_ can be written as `element()` which is equivalent to `element(*)` but `as map()` and `as array()` are not permitted and always have to include the asterisk (`map(*)`) .

This is still the case even in XPath4 and I think we should have a look, if this can be omitted.
It does not add any value IMHO, but maybe I am overlooking something important.

https://qt4cg.org/specifications/xquery-40/xquery-40.html#doc-xquery40-ElementTest
https://qt4cg.org/specifications/xquery-40/xquery-40.html#doc-xquery40-ArrayTest
https://qt4cg.org/specifications/xquery-40/xquery-40.html#doc-xquery40-MapTest

The "fix" would be as simple as

```
[255]   AnyMapTest   ::=  "map" "(" "*"? ")"
```

```
[265]   AnyArrayTest ::=  "array" "(" "*"? ")"
```

-- update after comment from @michaelhkay --

The same does _not apply_ to FunctionTest as `as function()` specifies a zero-arity function wheras `as function(*)` matches functions of any arity.

line-o:
fixes #1403

Allow Array and Map tests to omit the asterisk to match any array or map.

- `array()` <> `array(*)`
- `map()` <> `map(*)`

### DIFF
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -5223,8 +5223,8 @@ name.</p>
 
 
 
-               <p>The <nt def="MapTest">MapTest</nt>
-                  <code>map(*)</code> matches any map. The <nt def="MapTest">MapTest</nt>
+               <p>The <nt def="AnyMapTest">AnyMapTest</nt>
+                  <code>map()</code> or <code>map(*)</code> matches any map. The <nt def="TypedMapTest">TypedMapTest</nt>
                   <code>map(K, V)</code> matches any map where <phrase diff="del" at="issue730">the type of </phrase>every key
                   is an instance of <code>K</code> and <phrase diff="del" at="issue730">the type of </phrase>every value is an
   instance of <code>V</code>.</p>
@@ -5243,6 +5243,11 @@ name.</p>
   </p>
 
                <ulist>
+                  <item>
+                     <p>
+                        <code>$M instance of map()</code>  returns <code>true()</code>
+                     </p>
+                  </item>
                   <item>
                      <p>
                         <code>$M instance of map(*)</code>  returns <code>true()</code>
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -5651,7 +5651,7 @@ name.</p>
 
 
                <p>The <nt def="AnyArrayTest">AnyArrayTest</nt>
-                  <code>array(*)</code> matches any
+                  <code>array()</code> or <code>array(*)</code> matches any
   array. The <nt def="TypedArrayTest"
                      >TypedArrayTest</nt>
                   <code>array(X)</code> matches any array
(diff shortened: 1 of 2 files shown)


michaelhkay (changes requested):
`<g:string>*?</g:string>` is wrong. It needs to be `<g:optional><g:string>*</g:string></g:optional>`

michaelhkay:
I think it would also be a good idea to make clear in 3.3.2.8 and 3.3.2.9 that the two forms are equivalent.

line-o:
@michaelhkay I try to make that clear in 3.3.2.8 (see screenshot) and similarly in 3.3.2.9

Screenshot 2024-09-08 at 16 07 53

Do you have a suggestion how to make that point clearer?
Adding `map() and map(*) are equivalent.` to `The AnyMapTest map() or map(*) matches any map.` seems redundant.

michaelhkay:
Perhaps in 3.3.2.9 clause 1(b) change `B is array(*)` to `B is an AnyArrayTest (that is, array() or array(*))` (and similarly for maps). I'l leave it to you. But when there are two equivalent ways of saying the same thing, I think we need an occasional reminder that what we say about one form applies equally to the other.

line-o:
I did some work on section 3.3 Subtype Relationships that should drive the point home.

specifications/xquery-40/src/expressions.xml
```diff
@@ -5648,17 +5653,19 @@ name.</p>
...
 
                <ulist>
+                  <item>
+                     <p>
+                        <code>[ 1, 2 ] instance array()</code> returns <code>true()</code>
+                     </p>
+                  </item>
                   <item>
                      <p>
                         <code>[ 1, 2 ] instance array(*)</code> returns <code>true()</code>
```
ChristianGruen, line 5671: The one that was already in the spec:

```suggestion
                        <code>[ 1, 2 ] instance of array(*)</code> returns <code>true()</code>
```

specifications/xquery-40/src/expressions.xml
```diff
@@ -5648,17 +5653,19 @@ name.</p>
...
+                  The <nt def="TypedArrayTest">TypedArrayTest</nt> <code>array(X)</code> matches any array
+                  in which every array member matches the <nt def="SequenceType">SequenceType</nt><code>X</code>.
+               </p>
 
                <p>Examples:</p>
 
                <ulist>
+                  <item>
+                     <p>
+                        <code>[ 1, 2 ] instance array()</code> returns <code>true()</code>
```
ChristianGruen, line 5666: The copied one:

```suggestion
                        <code>[ 1, 2 ] instance of array()</code> returns <code>true()</code>
```

line-o:
It just occurred to me that if we would to accept this change we should also allow `record()` to be equal to `record(*)`.

line-o:
Since this change was rejected, should I close it myself?

ndw:
At [meeting 092](https://qt4cg.org//meeting/minutes/2024/10-01.html), the CG declined to accept this PR.

ndw:
It would have been fine if you had, but I didn't notice your question before I closed it :-)

PR #1431, opened 2024-09-11
1372 Unknown option: FORG0013 → XPTY0004

### LINKED ISSUE #1372: Unknown option: FORG0013 → XPTY0004
I would suggest dropping https://qt4cg.org/specifications/xpath-functions-40/Overview.html#ERRFORG0013 and replacing it with a type error, `XPTY0004`. This way, it can be perfomed during the type checks (which may also be the best solution if we define options as records).

I’ll be glad to create the PR for the spec and the tests.

ChristianGruen:
Closes #1372


michaelhkay (commented):
I think you need to use an `<xerrorref>` rather than `<errorref>` to get the cross-spec reference rendered correctly. Otherwise, fine.

ndw:
The CG agreed to merge this issue at [meeting 090](https://qt4cg.org/meeting/minutes/2024/09-17.html).

PR #1432, opened 2024-09-11
1379 Initializing expression: Allow self references

### LINKED ISSUE #1379: Circular dependencies: XQDY0054 vs. XPST0008 vs. optional errors
It remains a challenge (for me) to understand how the specification expects implementations to handle circular dependencies. Sorry in advance for mixing up the specification, tests and the behavior of Saxon (it basically demonstrates my confusion):

### Error Codes

The definition of `err:XPST0008` is:

> It is a [static error](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-static-error) if an expression refers to an element name, attribute name, schema type name, namespace prefix, or variable name that is not defined in the [static context](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-static-context), except for an ElementName in an [ElementTest](https://qt4cg.org/specifications/xquery-40/xquery-40.html#doc-xquery40-ElementTest) or an AttributeName in an [AttributeTest](https://qt4cg.org/specifications/xquery-40/xquery-40.html#doc-xquery40-AttributeTest).

The definition of `err:XQDY0054` is:

> It is a [dynamic error](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-dynamic-error) if a cycle is encountered in the definition of a module’s dynamic context components, for example because of a cycle in variable declarations.

### Queries

The following test case [K-InternalVariablesWith-15b](https://github.com/qt4cg/qt4tests/blob/4659dfdcc92ac57c9a7f482a1ec9f3f509020f52/prod/VarDecl.xml#L1527C21-L1537) expects `XPST0008` as error:

```xquery
declare variable $var1 := $var1;
true()
```

I would expect `XQDY0054` as error, and `true()` to be a valid alternative, as `$var1` is indeed defined in the static context. Think of a similar query (for which I would as well expect `XQDY0054` or `true()` to be correct results):

```xquery
declare variable $var1 := $var2;
declare variable $var2 := $var1;
true()
```

Saxon outputs  the following error for the first query…

> [XPST0008] Circular definition of global variable: $var1 uses $var1.

…which is confusing in itself, as the error code and the message do not really match. The output of the second query is consistent:

> [XQDY0054] Circular definition of global variable: $var2 uses $var1, which uses $var2.

Another test case that expects `XQDY0054` is [K-InternalVariablesWith-17a](https://github.com/qt4cg/qt4tests/blob/4659dfdcc92ac57c9a7f482a1ec9f3f509020f52/prod/VarDecl.xml#L1582-L1591):

```xquery
declare variable $var := local:func1();
declare function local:func1() { local:func2($var) };
declare function local:func2($arg2) { 1 };
true()
```

Again, I would expect `true()` to be a valid alternative, in `K-InternalVariablesWith-17` (the original XQuery 1.0 test), the rational for reporting an error was…

> A prolog variable having a circular dependency, by having a variable reference in a call site argument. This is an error even though the variable isn't used, because implementations cannot skip reporting static errors.

…but we now have a dynamic error.

### Suggestions

1. `XPST0008` should only be raised if no declaration exists for that variable in the query prolog.
2. `XQDY0054` should be optional for cases in which the affected code is never evaluated.

I would like to hear some feedback on this, and I will be happy to move the corresponding questions to `qt4tests` and/or the Saxon bug tracker if we believe that the spec is comprehensive enough.

ChristianGruen:
With function declarations, recursive functions can be declared:

```xquery
declare function local:factorial($x) {
  if($x > 1) then $x + local:factorial($x - 1) else $x
};
local:factorial(5)
```

We should allow the same for variable declarations:

```xquery
declare variable $factorial := fn($x) {
  if($x > 1) then $x + $factorial($x - 1) else $x
};
$factorial(5)
```

I believe it is sufficient to simplify the definition of *initializing expressions* and drop the exception “other than the variable being declared”.

Related: https://www.w3.org/Bugs/Public/show_bug.cgi?id=15791 (the static dependency check was given up before due to `fn:function-lookup`).

Closes #1379

### DIFF
--- a/specifications/xquery-40/src/query-prolog.xml
+++ b/specifications/xquery-40/src/query-prolog.xml
@@ -1322,7 +1322,7 @@ return $node/xx:bing]]>
         declaration includes an expression (<code>VarValue</code> or <code>VarDefaultValue</code>),
         the expression is called an <term>initializing expression.</term> The static context for an
         initializing expression includes all functions, variables, and namespaces that are declared
-        or imported anywhere in the Prolog, other than the variable being declared.</termdef>
+        or imported anywhere in the Prolog.</termdef>
     </p>
     
     <p diff="add" at="2022-11-17">If a required type is defined, then the value obtained by 
--- a/specifications/xquery-40/src/query-prolog.xml
+++ b/specifications/xquery-40/src/query-prolog.xml
@@ -1169,6 +1169,10 @@ return $node/xx:bing]]>
         and means that the rules for binding to variables are the same as the rules for binding
         to function parameters.
       </change>
+      <change issue="1379" PR="1432" date="2024-09-12">
+         In earlier versions, the static context for the <termref def="dt-initializing-expression"/>
+         excluded the variable being declared. This restriction has been lifted.
+      </change>
     </changes>
 
     <scrap>


michaelhkay (changes requested):
Please add a "changes" section flagging the fact that there's a specification change here.

michaelhkay:
>By the way, it can be tricky to know the number of PR before changes are pushed.

Yes, I've sometimes done an extra commit/push just to deal with that. A bit more effort for the author to provide a minor convenience for the reader. If it's inconvenient we can always leave it till a bulk wrap-up later.

ndw:
The CG agreed to merge this issue at [meeting 090](https://qt4cg.org/meeting/minutes/2024/09-17.html).

PR #1434, opened 2024-09-11
1373 XQFO: Editorial

### LINKED ISSUE #1373: XQFO: Editorial
* Wrong example: `hash("password123" || $salt, "sha-1234567")`
* The following test lacks a `use="v-deep-equal-at"` attribute:

```xquery
deep-equal(
  $at//name[@first="Bob"], 
  $at//name[@last="Barker"],
  options := { 'items-equal': op('is') } 
)
```
* `graphemes($crlf)`: Result should be `char('\r') || char('\n')` instead of `$crlf` (otherwise, test doesn’t work)
* fix `fn:unix-time` example: `1969-12-31T23:59:59.999Z` (thanks @johnlumley)
* TODO for `fn:replicate` can be dropped
* `fn:parse-uri`: Remove example for `FOUR0002`.

…to be continued

ChristianGruen:
Closes #1373

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -13849,9 +13849,6 @@ filter($input, fn($item, $pos) { $pos gt 1 })
          multiple references to the same node. So, for example, <code>fn:count(fn:replicate(/, 6)|())</code>
          returns <code>1</code>, because the <code>fn:replicate</code> call creates duplicates, and the
             union operation eliminates them.</p>
-         <p>[TODO: the use of type <code>xs:nonNegativeInteger</code> for the second argument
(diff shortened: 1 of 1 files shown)


michaelhkay (changes requested):
In the fn:graphemes() examples, I think the declaration of variable $crlf can now be dropped?

ChristianGruen:
> In the fn:graphemes() examples, I think the declaration of variable $crlf can now be dropped?

It was still used in the expression; but it indeed seems superfluous to have a variable if it is only referenced once, so I have removed it as suggested.

ndw:
The CG agreed to merge this PR without further discussion at [meeting 091](https://qt4cg.org/meeting/minutes/2024/09-24.html).

PR #1436, opened 2024-09-11
1323 Function parameters names: $href → $uri

### LINKED ISSUE #1323: Function parameters names: $uri vs. $href
I was asked about the difference between the function parameter names `$href` and `$uri`…

* `fn:doc($href)` (“Retrieves a document using a URI supplied as an `xs:string`…”); `fn:unparsed-text($href)`; others
* `fn:collection($uri)`, `fn:uri-collection($uri)`; others

…and I couldn’t give a convincing answer. At least in BaseX, both `fn:doc` and `fn:collection` can be used interchangeably to address single resources (but things change when the target contains multiple resources).

Would it be reasonable to rename `$href` to `$uri`, or are there reasons why we need to differ those two?

If we rename the parameter, `fn:resolve-uri` may be the only function using `$href` (maybe even here, `$uri` could work?).

ChristianGruen:
Closes #1323

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -7103,7 +7103,7 @@ some $t in $value ! tokenize(.) satisfies
    <fos:function name="resolve-uri" prefix="fn">
       <fos:signatures>
          <fos:proto name="resolve-uri" return-type="xs:anyURI?">
-            <fos:arg name="href" type="xs:string?"/>
+            <fos:arg name="uri" type="xs:string?"/>
             <fos:arg name="base" type="xs:string?" default="()"/>
          </fos:proto>
       </fos:signatures>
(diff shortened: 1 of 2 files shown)


michaelhkay (commented):
Technically a relative reference is not a URI, so I think there is some merit in using the (weird but traditional) name `href` for something that can be a relative reference as well as a URI.

ChristianGruen:
> Technically a relative reference is not a URI, so I think there is some merit in using the (weird but traditional) name `href` for something that can be a relative reference as well as a URI.

I had always thought that URIs can be relative (as e.g. described here: https://www.w3.org/DesignIssues/Relative.html).

If we use `href` for parameters that may be relative, shouldn’t we also use it for `fn:collection`?

I had always my problems with `href`. It stands for “hypertext reference”, which is fine for `<a href=.../>`, but seems indeed weird for `fn:doc` function calls. But my perspective may reflect a minority.

ndw:
There are no "right" answers here, there are just a bunch of different, imperfect alternatives.

I tend to think that the pattern of `<a href=` and `<link href=` has inculcated the idea that "if it's a document you can point to and get, then it's an href". On that basis, I prefer `doc($href)` over `doc($uri)` but I'm not going to argue about it.

On `resolve-uri` I have a strong preference for `$href` because I think one of the most common uses of `resolve-uri` is to take a thing you got out of an `href` attribute and resolve it. So the thing you're putting there is an href. But I'm not going to argue about it.

Whatever we do for `doc()`, we should do the same thing for `doc-available()`, `unparsed-text()`, and `unparsed-text-lines()`. I will argue about that :-) )

I don't have any motivation for liking one or the other on `zip:binary-entry`.

ChristianGruen:
Thanks for the insights.

Maybe `fn:doc($input)`/`fn:collection($input)` or `fn:doc($source)`/`fn:collection($source)` would be more neutral?

After all, it’s just the names of (mandatory) parameters that no one cared about before version 4, so I’ll be fine with either solution.

ChristianGruen:
Obsolete.

PR #1437, opened 2024-09-11
1325 Variadic System Functions limited to `fn:concat`

### LINKED ISSUE #1325: Variadic System Functions: Principles?
In the current spec, the following functions are variadic:

* `fn:concat`
* `fn:codepoints-to-string`
* `fn:distinct-ordered-nodes`

The advantage of the variadic representation is that a user can omit additional parentheses; the drawback is that the function cannot be enhanced with parameters later on. In my point of view, `fn:distinct-ordered-nodes` might be a candidate for that in the future (on the other hand, it will only be used by a very small user group one or the other way).

Other candidates for variadicity could be (among others):

* `fn:count`
* `fn:exists`, `fn:empty`
* `fn:head`, `fn:tail`, `fn:trunk`, `fn:foot`
* `fn:one-or-more`, `fn:exactly-one`, `fn:zero-or-one`
* `fn:innermost`, `fn:outermost`, `fn:unordered`, `fn:reverse`
* `fn:boolean`, `fn:not`
* `fn:identity`
* `fn:data`, `fn:has-children` (the 0-arity case would need to be preserved due to its special semantics)

Note that function calls like `fn:exactly-one(E1, E2)` can be reasonable in practice, as `E1` may return an empty sequence.

I wonder if we can define a principle which functions should be variadic?

ChristianGruen:
Closes #1325

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -4068,14 +4068,13 @@ translate(value := '٢٠٢٣', replace := '٠١٢٣٤٥٦٧٨٩', with := '01234
    <fos:function name="codepoints-to-string" prefix="fn">
       <fos:signatures>
          <fos:proto name="codepoints-to-string" return-type="xs:string">
-            <fos:arg name="values" type="xs:integer*" default="()"/>
+            <fos:arg name="values" type="xs:integer*"/>
          </fos:proto>
       </fos:signatures>
       <fos:properties>
          <fos:property>deterministic</fos:property>
          <fos:property>context-independent</fos:property>
          <fos:property>focus-independent</fos:property>
-         <fos:property>variadic</fos:property>
       </fos:properties>
       <fos:summary>
          <p>Returns an <code>xs:string</code> whose characters have supplied <termref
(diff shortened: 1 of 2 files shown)


michaelhkay (commented):
1. In 1.5.3, the sentence "More strictly, there is an infinite set of this function, with arity ranging from 0 to infinity." Seems poor grammar. There is an infinite set of functions having the same name and different arities.
2. More generally, if we have a generalised mechanism for variadic functions then I don't see why we wouldn't want to use the mechanism for the functions codepoints-to-string and distinct-ordered-nodes.

ChristianGruen:
> 1. In 1.5.3, the sentence "More strictly, there is an infinite set of this function, with arity ranging from 0 to infinity." Seems poor grammar. There is an infinite set of functions having the same name and different arities.

Fixed, thanks.

> 2. More generally, if we have a generalised mechanism for variadic functions then I don't see why we wouldn't want to use the mechanism for the functions codepoints-to-string and distinct-ordered-nodes.

I think that…

* the current solution feels arbitrary: What makes these two functions so different. We still lack any consistent or reasonable principle which functions should be variadic and which shouldn’t.
* We may want to enhance the two functions with additional parameters, which wouldn’t be possible by making it variadic. For `codepoints-to-string`, we could introduce a fallback option (e.g., to return `U+FFFD` for invalid XML 1.0 codepoints); for `distinct-ordered-nodes`, we could have a future option to return nodes of different documents ordered by a given rule.

See https://github.com/qt4cg/qtspecs/issues/1325#issuecomment-2342708773 for the original reasoning.

In my point of view, `fn:identity` would be the only candidate that would offer real advantage as a variadic function (it would allows us to use `fn:identity#2`, `fn:identity#3`, etc.), but I haven’t added it here.

ndw:
At [meeting 092](https://qt4cg.org/meeting/minutes/2024/10-01.html), the CG agreed to merge this PR.

PR #1438, opened 2024-09-11
1322 fn:collation-available (editorial)

### LINKED ISSUE #1322: fn:collation-available (editorial)
Minor observations:

1. As the function input is a plain URI, I would propose to rename `$collation` to `$uri` (see `fn:collection` and other functions).
2. I believe we should make `fn:collation-available($uri, ())` and `fn:collation-available($uri)` equivalent. Could we change the default value of `$usage` to `()`?
3. It would be helpful to have (successul and failing) examples for the `$usage` argument added to the specification. Maybe things get easier once test cases exist.

ChristianGruen:
Closes #1322

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -23400,7 +23400,7 @@ return map:build($titles/title, fn($title) { $title/ix })
    <fos:function name="collation-available" prefix="fn">
       <fos:signatures>
          <fos:proto name="collation-available" return-type="xs:boolean">
-            <fos:arg name="uri" type="xs:string"/>
+            <fos:arg name="collation" type="xs:string"/>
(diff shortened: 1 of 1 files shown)


michaelhkay (commented):
I don't see the benefit in changing the parameter name from `collation` to `uri`. It's `collation` everywhere else.

ChristianGruen:
> I don't see the benefit in changing the parameter name from `collation` to `uri`. It's `collation` everywhere else.

I may have been misled by functions like `fn:doc-available`, which check if the argument *refers* to an available document. Similarly, the argument of `fn:collation-availabe` is not really a collation, but a string/URI referring to a potentially available collation.

However, it is true that this is similar for to the other occurrences of the `$collation` parameter, so I have reverted the change.

ndw:
The CG agreed to merge this PR without further discussion at [meeting 091](https://qt4cg.org/meeting/minutes/2024/09-24.html).

PR #1439, opened 2024-09-11
1235 Function Identity: Treating function items with identical bodies

### LINKED ISSUE #1235: Function Identity: Treating function items with identical bodies
(One) requirement resulting from #908 (in particular, https://github.com/qt4cg/qtspecs/issues/908#issuecomment-1891524815):

The following examples reflect the status quo:

\# | Function | Result
-- | -- | --
1 | `deep-equal(<a/>, <a/>)` | `true`
2 | `let $f := fn { <a/> } return deep-equal($f, $f)` | `true`
3 | `deep-equal(fn { 1 }, fn { 1 })` | `true` or `false`
4 | `deep-equal(fn { <a/> }, fn { <a/> })` | `false`

#### Notes

1. The result is `true` (regardless of the identity of the compared nodes).
2. The result is `true` (regardless of the identity of the nodes that result from evaluating the functions)
3. The result is allowed to be `true` or `false`, depending on the optimization strategies of a processor.
4. Only in this case, the result must currently be `false`.

We should allow `deep-equal` to return `true` for function items that have the same arguments and bodies. An implementation should be allowed to use the same internal representation for multiple occurrences of such functions.

ChristianGruen:
Closes #1235

### DIFF
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -10620,14 +10620,20 @@ return $incrementors[2](4)]]></eg>
                      function references using this function name and arity return the same function.
                   For more details see <specref ref="id-named-function-ref"/>.</p>
                   </item>
-                  <item><p>An optimizer is permitted to rewrite expressions in such a way that repeated
-                  evaluation is avoided if it can be established that the result will be the same each time,
-                  and this may be done without consideration of function identity. For example, if the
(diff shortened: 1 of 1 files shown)


michaelhkay (approved):

ndw:
The CG agreed to merge this PR at [meeting 091](https://qt4cg.org/meeting/minutes/2024/09-24.html).

PR #1440, opened 2024-09-12
1387 Another tweak to build-uri

ndw:
In #1387 @ChristianGruen observes that the `+` seems to be a special case in the `tel:` scheme. In fact, I think we concluded that the special case is that you shouldn't encode the path segment of a non-hierarchical URI.

This PR attempts to implement that.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -31614,7 +31614,7 @@ path with an explicit <code>file:</code> scheme.</p>
            <item><p>If the <code>path-segments</code> key exists in the map,
            then the path is constructed from the segments.</p>
 
-           <p>To construct the path, each
+           <p>To construct the path, if the URI is hierarchical, each
            segment is encoded, then they are joined together, separated by <code>/</code> (solidus)
            characters, to form the path.
            The encoding performed replaces any control characters (codepoints less than 0x20)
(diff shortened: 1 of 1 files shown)


ChristianGruen (changes requested):
After having read the PR again, I think it needs to be improved a little:

`'a+b/c' => parse-uri()` returns a non-hierarchical path. With the current rules, the result of `'a+b/c' => parse-uri() => build-uri()` would be `a b`.

Maybe we should:

1. apply the rule to all path segments and
2. not consider the `hierarchical` attribute, but rather check if a supplied scheme is known to be non-hierarchical?

ChristianGruen (approved):

ChristianGruen (changes requested):
Copied from my comment, and slightly revised:

`'a+b/c' => parse-uri()` returns a non-hierarchical path. With the proposed rules, if I understand it correctly, the result of `'a+b/c' => parse-uri() => build-uri()` would be `a b`.

Maybe we should:

1. apply the rule to all path segments and
2. not consider the `hierarchical` attribute, but rather check if a scheme is supplied and if it is known to be non-hierarchical?

ndw:
Yes, you've persuaded me. And I've tried to change the rules accordingly. It was not easy to work out how best to describe those rules. Comments welcome.

ChristianGruen (commented):
@ndw Looks and feels good.

1. I noticed only now that the new rule of what’s *non-hierarchical* is pretty similar to what’s found further above for schemes. Maybe it would be reasonable to use exactly the same rule (if we believe that this makes sense)?

> A URI is considered to be non-hierarchical if either the hierarchical key is present in the $parts map with the value false() or if the scheme is known to be non-hierarchical. (In other words, schemes are hierarchical by default.)

2. I now get `jar:file:/C:/Program Files/test.jar!/foo/bar` instead of `jar:file:/C:/Program%20Files/test.jar!/foo/bar` for the test case `fn-build-uri-from-parse-028`. Do you think we can/should live with that?

ChristianGruen (approved):

ChristianGruen:
I’ve readded `Propose Merge without Discussion`, maybe we can add it to today’s list.

ndw:
The CG agreed to merge this issue without discussion at [meeting 090](https://qt4cg.org/meeting/minutes/2024/09-17.html).

PR #1460, opened 2024-09-24
1323b Function parameters names: $uri → $source

### LINKED ISSUE #1323: Function parameters names: $uri vs. $href

ChristianGruen:
Second attempt. Closes #1323 and #1436.

In addition, includes a change to `fn:json-doc($uri)`, and fixed an editorial bug for the `fn:escape-html-uri` function.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -5955,7 +5955,7 @@ return normalize-unicode(concat($v1, $v2))</eg>
             printable characters of the US-ASCII coded character set, specifically the <termref
                def="codepoint"
                >codepoints</termref> between 32 and 126 (decimal) inclusive. Each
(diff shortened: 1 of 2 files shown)


ChristianGruen:
@ndw I hope this one can be merged. I didn’t see any merge conflicts, but I’ve rebased it nevertheless.

Issue #689, opened 2023-09-08
fn:stack-trace: replace with $err:stack-trace


ChristianGruen:
The current specification contains a diagnostic function called [`fn:stack-trace`](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-stack-trace). Many other languages provide a similar function: The returned output can possibly help to understand which function calls led to an error during the evaluation of a code.

Still, I have strong doubts that it is a good decision to include this function in the standard:

The specification gives you a vast amount of freedom how to implement and optimize things. As a result, it’s completely feasible and reasonable to rewrite the following code…

```xquery
declare function local:double($f) {
  $f * 2
};
(1 to 6) ! local:double(.)
```

…to `(1 to 6) ! (. * 2)` at compile time. If a user adds a `fn:stack-overflow` call in the function body, s·he would expect to find the function invocation of the original code representation in the output. As always, there are technical solutions to achieve this (store additional information in the evaluation tree on the original query; suppress optimizations when `fn:stack-trace` is found), but all of them can affect the runtime behavior and lead to different evaluation trees, hiding possible bugs in the implementation (which can be a reason to call `fn:stack-trace` at all).

A standard should provide a minimum amount of assurance that a function behaves similarly across different implementations. At this time, I don’t believe we can’t give that guarantee.

Related: #55, #686

– As an alternative, a stack trace could optionally be created by an implementation when an error is triggered.

Arithmeticus:
I doubt that I am representative of the type of person who should influence this question. That said, if `fn:stack-trace` were made available, I would use it relatively often, certainly in development work and possibly in production.

ChristianGruen:
@Arithmeticus Thanks. Out of interest: What would be productive use cases for the function? Would it be beneficial or limiting for you to have the stack trace generated by errors instead?

Arithmeticus:
I have had to develop applications in environments where it is difficult (sometimes very difficult) to intercept the message listener when invoking an XSLT subprocess, and the ability to write in XSLT the stack trace to a secondary result document (`<xsl:result-document>`) would be useful. It's hard for me to say further, because I don't know what the output of `fn:stack-trace` would be, and it's only by playing around with it that I can discern other ways it might be used.

The option with errors would be a benefit. That is, bundling as an available variable within e.g. `<xsl:catch>`? Seems to make more sense.

ChristianGruen:
To revive this discussion:

[...]

PR #1470, opened 2024-10-01
689 fn:stack-trace: replace with $err:stack-trace

ChristianGruen:
Issue: #689

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -30826,41 +30826,6 @@ fold-left($input, false(), fn($result, $item, $pos) {
       </fos:changes>
    </fos:function>
 
-   <fos:function name="stack-trace" prefix="fn">
-      <fos:signatures>
-         <fos:proto name="stack-trace" return-type="xs:string"/>
-      </fos:signatures>
-      <fos:properties>
-         <fos:property>nondeterministic</fos:property>
-         <fos:property>context-independent</fos:property>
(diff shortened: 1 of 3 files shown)


michaelhkay:
My concerns would be:

(a) a stack trace can be useful even if there is no error. It seems cumbersome having to raise and catch and error just to get a stack trace. 

(b) the way the information is being provided -- as a variable and as a map entry -- doesn't exactly make it easy for the implementation to construct and format the information lazily on demand. Providing it through a function is much more likely to be performant.

(c) XSLT 4.0 needs to be kept in sync.

ChristianGruen:
> (a) a stack trace can be useful even if there is no error. It seems cumbersome having to raise and catch and error just to get a stack trace.

I followed the discussion in the issue, which resulted in a comment that it was consensus in Prague to remove the function (https://github.com/qt4cg/qtspecs/issues/689#issuecomment-2147376442).

> (b) the way the information is being provided -- as a variable and as a map entry -- doesn't exactly make it easy for the implementation to construct and format the information lazily on demand.

An alternative would be to return a function that returns the stack trace.

michaelhkay:
I would be inclined to make it available (only) as a function within the map, so you do `$err:map?stack-trace()`.

ChristianGruen:
Revised as discussed in [Meeting 094](https://qt4cg.org/meeting/minutes/2024/10-15.html); I kept `xs:string?` as return type.

ndw:
The CG agreed to merge this issue at [meeting 098](https://qt4cg.org/meeting/minutes/2024/11-12.html).

Issue #1471, opened 2024-10-01
JSON Serialization: Sequences on Top Level


ChristianGruen:
Extracted from #576:

All serialization methods except for `json` allow sequences to be output on top level. We should also allow this for JSON data. This way, constructs like…

```xquery
declare option output:item-separator '&#xa;';
declare option output:method 'text';

for $json in ({ 1: 2 }, { 3: 4 })
return serialize($json, { 'method': 'json' })
```

…can be simplified to:

```xquery
declare option output:method 'json';
{ 1: 2 },
{ 3: 4 }
```

We should still not allow sequences within JSON structures. It would be inconsistent to have different output rules depending on the number of items in a sequence (`(1)` and `[1]` would not be serialized identically).

PR #1472, opened 2024-10-01
1471 JSON Serialization: Sequences on Top Level

ChristianGruen:
Issue: #1471

### DIFF
--- a/specifications/xslt-xquery-serialization-40/src/errors.xml
+++ b/specifications/xslt-xquery-serialization-40/src/errors.xml
@@ -144,7 +144,7 @@ has two keys with the same string value, unless the
 
 <error spec="SE" code="0023" class="RE" type="serialization" 
 diff="add" at="2014-11-06">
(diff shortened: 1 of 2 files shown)


ChristianGruen:
…@ndw I re-committed all my PRs, but I guess there’s nothing I can do about the failing tests by myself?

ndw:
Sorry about the PR build failure. Not your fault. I fixed it and reran the job.

michaelhkay:
I'm not entirely convinced this is the right approach. If the user wants to output something that conforms to json-lines, this almost achieves it (using the default item-separator) except that an empty sequence is output as "null", rather than as an empty file. That glitch seems to be inviting trouble.

ChristianGruen:
The PR will be revised and a new serialization parameter for [JSON Lines](https://jsonlines.org/) will be added.

ChristianGruen:
Outdated; see #1497

PR #1497, opened 2024-10-15
1471 JSON Serialization: json-lines

### LINKED ISSUE #1471: JSON Serialization: Sequences on Top Level

ChristianGruen:
Closes #1471 and #1472.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -19121,6 +19121,18 @@ return tokenize(normalize-space($s), ' ')[. castable as xs:IDREF]</eg>
                </tr>
                <tr>
                   <td>
+                     <code>json-lines</code>
+                  </td>
+                  <td>
+                     <code>xs:boolean?</code>
+                  </td>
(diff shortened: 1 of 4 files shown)


michaelhkay:
Requires XSLT changes to accept the new serialization parameter in xsl:output and xsl:result-document

joewiz:
Great idea! Thanks for this!

ndw:
The CG agreed to merge this PR at [meeting 096](https://qt4cg.org/meeting/minutes/2024/10-29.html).

Issue #1318, opened 2024-07-09
Function Coercion: Records, Maps, Arrays


ChristianGruen:
**Conclusion (2024-09-11):** As we plan to keep coercion rules for records, we should add rules for arrays and maps as well: If `$v as xs:int+` is successful for `(1, 2)`, `$v as array(xs:int)` should be successful for `[ 1, 2 ]`.

-------------------

I like to question the coercion rule that encompasses record tests:

> 6. If *R* is a [RecordTest](https://qt4cg.org/specifications/xquery-40/xquery-40.html#doc-xquery40-RecordTest) and *J* is a map, then J is converted to a new map as follows:

> * The keys in the supplied map are unchanged.

> * In any map entry whose key is an `xs:string` equal to the name of one of the field declarations in R, the corresponding value is converted to the required type defined by that field declaration, by applying the coercion rules recursively (but with XPath 1.0 compatibility mode treated as false).

I would like us to drop this rule. I believe that both the instance checks and the conversions of large maps can get very expensive. In addition, it may even require recursive rebuilds of map structures if a supplied record test includes nested record tests.

I think it’s completely fair to expect users to deliver maps in a way that matches record definitions, and it could even be counterintuitive if map updates take place as a consequence of a simple function call.

If we want to stick to that coercion rule, we should also define coercion rules for maps and arrays.

michaelhkay:
The motivation for this rule was primarily connected with numeric fields. If the required argument to a function is `record(longitude as xs:double, latitude as xs:double)` then it seems unreasonable not to allow the value `{"longitude": 125, "latitude": 53}` to be supplied.

Note that we already incur the same costs for large sequences: and records are typically bounded in size by the record definition, whereas sequences are unbounded.

ChristianGruen:
> it seems unreasonable not to allow the value `{"longitude": 125, "latitude": 53}` to be supplied.

Following that reasoning, we should definitely include coercion rules for arrays and maps as well.

> Note that we already incur the same costs for large sequences: and records are typically bounded in size by the record definition, whereas sequences are unbounded. And the cost of type-checking is there anyway even if the outcome is to reject the value rather than coercing it.

I would agree only partially: Definitions of records can be nested, which makes coercions more expensive if the type of a descendant value does not match. I may be wrong, but I believe it could be confusing that a trivial function call triggers a potentially recursive update of a data structure. Shouldn’t we rather try to define a robust record constructor instead of moving all the complex semantics in the coercion rules?

[...]

PR #1501, opened 2024-10-16
1318 Function Coercion: Records, Maps, Arrays

ChristianGruen:
Issue: #1318

Coercion rules added for maps and arrays.

### DIFF
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -7256,7 +7256,9 @@ declare record Particle (
                         is a map, then <var>J</var> is converted to a new map as follows:</p>
                      <olist>
                         <item><p>Each key in the supplied map is converted to the required
-                           map key type by applying the coercion rules.</p></item>
+                           map key type by applying the coercion rules. If the resulting map would
+                           contain duplicate keys, a type error is raised
+                           <errorref class="TY" code="0004"/>.</p></item>
                         <item><p>The corresponding value is converted to the required
                            map value type by applying the coercion rules recursively.</p></item>
                      </olist>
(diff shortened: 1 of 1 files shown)


ChristianGruen:
It turns out that our proof-of-concept implementation already rejected duplicates by raising an error.

@michaelhkay Do you think it will be sufficient to add a note that `XPTY0004` is raised if the resulting map would contain duplicates (possibly enriched with an example, like `{ 1.2: true(), 1.2000001: true() }` being coerced to `map(xs:float, xs:boolean)`)?

michaelhkay:
I feel it should be a dynamic error rather than a type error since in general it can't be detected by type analysis? It's more like the error you get when the required type is `xs:integer` and the supplied value is the `xs:untypedAtomic` value "xyz".

ChristianGruen:
I thought of cases like `for $i as xs:byte in (1 to 128) return $i`, which currently raise `XPTY0004`. Well, they should: There seem to be cases in which we raise `FORG0001`, which doesn’t look correct (but no one seemed to care over the years).

ChristianGruen:
After further reflection, I think there are many cases in which it will be possible to type maps and arrays statically (for example if they are passed externally, and analyzed before evaluating the query). But isn’t `XPTY0004` raised both statically and dynamically anyway? I think of queries like…

```xquery
let $r as xs:integer := if(year-from-date(current-date()) > 0) then 'a' else 1
return $r
```

michaelhkay:
All type errors can be raised statically or dynamically.

`for $i as xs:byte in (1 to 128) return $i`

is a dynamic error FORG0001 according to the spec, but it gets into fuzzy errors-and-optimization territory.

ChristianGruen:
Revised: Added a rule and a note for duplicate map keys. I have used `XPTY0004`, as it seemed more generic.

Issue #1458, opened 2024-09-19
Arguments that have a default value but don't accept ()


michaelhkay:
The following functions define an argument that has a default value, but don't allow () to be supplied explicitly:

```
lang
id
element-with-id
idref
parse-html
```

As determined by running the following query against the function catalog:

`//*:arg[@default and not(ends-with(@type,'?') or ends-with(@type, '*'))]/(ancestor::*:function/(@prefix||':'||@name))`

There are others which this query doesn't pick up, mainly where the defaultable argument is a function, including:

array:build
map:build
array:get
map:get
fn:highest
fn:lowest

ndw:
In fact, the prose has become a little tangled. Looking at `id` as an example, we find:

> `E` is in the target document. The target document is the document containing `$node`, or the document containing the context value (`.`) if the second argument is omitted.

That reads a little oddly because if the second argument is omitted, it will now get an explicit default value, and that doesn't feel quite like "omitted".

I'm not sure if your observation implies a proposal that they should allow `()` as an explicit value, but assuming it does...

I'm not sure. You'd get:

* `id('foo')` => `id('foo', .)`
* `id('foo', .)` => `id('foo', .)`
* `id('foo', ())` => `id('foo', .)`

I think I'd prefer not to provide a default and have the semantics as they're currently written (because that was the status quo when they were written) or to provide a default and to make it an error to pass `()` explicitly. The `id` function *needs* a context, if you pass an expression that happens to evaluate to the empty sequence, that's more likely to be an error than to be "use the document that contains the current context value". IMHO.

I think similar reasoning holds for `lang`, `element-with-id`, and `idref`.

For `parse-html` the option is a map. If we allow an explicit empty sequence, we have to decide what it means. On thirty seconds thought, I think the simpler thing is just not to allow it there either.

ChristianGruen:
This one may catch some more cases:

```xquery
//*:arg[@default and matches((@type, @type-ref), '^fn|[^?*]$')]
  /ancestor::*:function ! (@prefix || ':' || @name)
```

ChristianGruen:
Regarding `fn:lang` etc., I agree with Norm (see, among others, https://github.com/qt4cg/qtspecs/issues/951).

michaelhkay:
Yes, I think you're probably right that where the default is ".", we shouldn't apply the default if () is supplied explicitly.

[...]

PR #1502, opened 2024-10-16
1458 Arguments that have a default value but don't accept ()

ChristianGruen:
Issue: #1458


ChristianGruen:
Everyone feel free to remove the label »Merge without Discussion« if we should discuss this one in the next meeting (I have only fixed the function signatures, but I have not revised the prose, as motivated by Norm in https://github.com/qt4cg/qtspecs/issues/1458#issuecomment-2360360990).

ndw:
At [meeting 095](https://qt4cg.org/meeting/minutes/2024/10-22.html), the CG agreed to merge this PR without discussion.

Issue #868, opened 2023-12-04
fn:intersperse → fn:join, array:join($arrays, $separator)


ChristianGruen:
With `string-join`, you can create a string for multiple strings, optionally interspersed with a separator. `array:join` can be used to create an array from multiple arrays.

`fn:intersperse`, which has been added to the XQuery 4 draft, does something very similar, and early feedback indicates that the function is useful, but easily to overlook due to its name.

I propose to unify the functions by…

1. renaming `fn:intersperse` to `fn:join`;
2. adding a parameter to `array:join`: `$separator as array(*)* := ()`; and
3. allowing a separator sequence for `fn:string-join`: `$separator as xs:string* := ()`.

#### Examples

Query | Result | Info
-- | -- | --
`string-join((1, 2))` | `'12'` | *existing syntax*
`string-join((1, 2), '-')` | `'1-2` | *existing syntax*
`string-join((1, 2), ('-', '+'))` | `'1-+2'` | *new*
`array:join([ [ 1 ], [ 2 ] ])` | `[ 1, 2 ]` | *existing syntax*
`array:join([ [ 1 ], [ 2 ] ], [ '-' ])` | `[ 1, '-', 2 ]` | *new*
`array:join([ [ 1 ], [ 2 ] ], [ '-', '+' ])` | `[ 1, '-', '+', 2 ]` | *new*
`join((1, 2))` | `(1, 2)` | `intersperse((1, 2))`<br/>(*or just* `(1, 2)`)
`join((1, 2), '-')` | `(1, '-', 2)` | `intersperse((1, 2), '-')`
`join((1, 2), ('-', '+'))` | `(1, '-', '+', 2)` | `intersperse((1, 2), ('-', '+'))`

ChristianGruen:
> Could also call it `sequence-join` or `items-join`, then it is more consistent with `string-join`

I suggested `fn:join` as we already have `array:join` and various other sequence functions that don’t include `sequence`.

> `items-join` variant would also be consistent with `items-after`

The `items-` functions will probably be renamed:

https://qt4cg.org/pr/855/xpath-functions-40/Overview.html#func-subsequence-after
https://github.com/qt4cg/qtspecs/issues/844#issuecomment-1826276165

> or it could be a sequence.
> `array:join([1,2,3], '-')` to make it `[1,'-',2,'-',3]`

I thought about this, but then it wouldn’t be possible to include multiple arrays.

joewiz:
[...]

PR #1504, opened 2024-10-16
868 fn:intersperse → fn:join, array:join($arrays, $separator)

ChristianGruen:
Issue: #868

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -27140,6 +27140,11 @@ array:of-members(
             </fos:test>
          </fos:example>
       </fos:examples>
+      <fos:changes>
+         <fos:change issue="868" date="2024-10-29">
+           <p>Optional <code>$separator</code> added.</p>
+         </fos:change>
+      </fos:changes>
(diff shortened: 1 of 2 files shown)


ChristianGruen:
> I'm not sure I understand the use case for making the second argument of string-join into a sequence.

It’s an alignment with intersperse/join; and I did not see a reason not to do it, as we already allow it to be an empty sequence.

wendellpiez:
`fn:chain-sequence()` as an alternative name for `fn:join()` ?

ChristianGruen:
I have done the following changes (Sorry for the repeated commits): 

* `fn:string-join`: reverted to accept at most one separator item.
* `fn:join`: renamed to `fn:sequence-join`
* `array:join`: Signature and equivalent XPath code changed to take charge of empty-sequence separators (thanks Reece).

michaelhkay (commented):
The array:join function needs a changes section.

ChristianGruen:
> The array:join function needs a changes section.

…added

ndw:
The CG agreed to merge this PR at [meeting 096](https://qt4cg.org/meeting/minutes/2024/10-29.html).

PR #1521, opened 2024-10-21
Update the changed/new marks in the ToC

### LINKED ISSUE #1357: Rendering of new vs. updated features
In previous versions of the XQFO drafts, it was possible to differentiate between NEW and UPDATED functions. It would be nice if this distinction could be preserved:

ndw:
Fix #1357 

This is a purely cosmetic PR. I think @ChristianGruen is right that it is unfortunate that we've lost the distinction between "new" and "only updated" functions. I'm trying to make that work again, in a way that's more visually distinct.

I have:

1. Use `Δ` for changed sections. 
2. Use `Δ➕` for new functions. (I'm making no effort to determine if non-function sections are new or changed; I doubt that it's either worth the effort or likely to be correct. One could argue that `➕` alone is sufficient, but I liked the consistency this way. Even new functions do change between drafts.)
3. Per @michaelhkay, a function is "new" if it doesn't appear on F&O 3.1.
4. I found it distracting that the delta symbol preceded the ToC entry. That's purely aesthetic, I guess, but I've put them at the end.

You can see the results here: https://qt4cgtest.nwalsh.com/branch/iss-1357/xpath-functions-40/Overview.html

### DIFF
--- a/specifications/css/xpath-functions-40.css
+++ b/specifications/css/xpath-functions-40.css
@@ -5,7 +5,7 @@
 }
 
 #toc .toc-new::after {
-    content: " Δ➕";
+    content: " ➕";
     font-weight: normal;
     color: blue;
 }
--- a/specifications/css/xpath-functions-40.css
+++ b/specifications/css/xpath-functions-40.css
@@ -4,6 +4,18 @@
   padding-top: 4em ! important;
 }
 
+#toc .toc-new::after {
+    content: " Δ➕";
+    font-weight: normal;
+    color: blue;
+}
+
+#toc .toc-chg::after {
+    content: " Δ";
+    font-weight: normal;
+    color: blue;
+}
+
 #function-finder {
   position: fixed;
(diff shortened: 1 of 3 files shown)


ChristianGruen (approved):
Great, thanks. A reasonable choice to place the icons after the string.

My minimalistic voice says that a single ➕ for new functionality could be sufficient.

ndw:
Okay. I made that change and pushed it. Give the process 15 minutes or so and the results should be at the same URI I pointed to above.

ChristianGruen:
Looks good.

Maybe ➕ should be added in the first Change section, which currently says:

> Sections with significant changes are marked Δ in the table of contents.

ndw:
Yes. I was mostly playing with the possible formats. But on the assumption that this PR will be accepted, I'll make that update!

ndw:
At [meeting 095](https://qt4cg.org/meeting/minutes/2024/10-22.html), the CG agreed to merge this PR.

Issue #1584, opened 2024-11-18
Review the XML Schema and RELAX NG schemas for XSLT 4.0 for compatibility


ndw:
It's likely that we've allowed them to drift.

ndw:
I hypothesize that this is a requirement, but we can leave it until near(er) the end.

michaelhkay:
I believe we have done this review and although there is a requirement to keep it under review, there is no open issue as such, so I propose we close it.

ndw:
At [meeting 160](https://qt4cg.org/meeting/minutes/2026/04-14.html), the CG agreed to close this issue without further action.

PR #1585, opened 2024-11-18
Update RELAX NG grammar for XSLT

ndw:
This PR updates the RELAX NG grammar to be (more) consistent with the XSD grammar (and consequently more correct for XSLT 4.0)

1. Relaxes the definition of QNames so that prefixes declared with `fixed-namespaces` can be supported
2. Adds the `fixed-namespaces` attribute to `xsl:stylesheet`, `xsl:transform`, and `xsl:package`
3. Adds the `xsl:switch` element
4. Adds the `xsl:array` and `xsl:array-member` elements
5. Adds `shallow-copy-all` to the possible values for `on-no-match`
6. Adds `separator` to `xsl:apply-templates`
7. Removes `select` attribute from `xsl:copy`
8. Adds `as` attribute to `xsl:sequence`
9. Updated the declarations for `xsl:accumulator-rule`, `xsl:array`, `xsl:array-member`, `xsl:attribute`, `xsl:catch`, `xsl:comment`, `xsl:map-entry`, `xsl:matching-substring`, `xsl:message`, `xsl:namespace`, `xsl:non-matching-substring`, `xsl:on-empty`, `xsl:on-non-empty`, `xsl:otherwise`, `xsl:param`, `xsl:processing-instruction`, `xsl:sequence`, `xsl:sort`, `xsl:value-of`, `xsl:variable`, `xsl:when`, and `xsl:with-param` so that they accept *either* a `select` attribute or a sequence constructor

More changes may also be required. A comprehensive comparision of the RNC and XSD schemas is needed, #1584

### DIFF
--- a/specifications/xslt-40/src/schema-for-xslt40.rnc
+++ b/specifications/xslt-40/src/schema-for-xslt40.rnc
@@ -1,18 +1,30 @@
-# XSLT 3.0 Relax NG Schema
+# XSLT 4.0 Relax NG Schema
 # 
 # Copyright (c) 2010-2016, Mohamed ZERGAOUI (Innovimax)
+# Portions © 2024, XQuery and XSLT Extensions Community Group
 # 
 # All rights reserved.
 # 
-# Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
-# Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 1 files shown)


ndw:
The CG agreed to merge this PR without discussion at [meeting 099](https://qt4cg.org/meeting/minutes/2024/11-19.html).

Issue #1588, opened 2024-11-18
Move the Streamability chapter?


ndw:
This is a minor thing, but it annoys me every single time. Open up the XSLT spec and search for any instruction and the first hit in the ToC is always 19.8.4.x "Streamability of [instruction]" which has never been the reason I was looking for the instruction.

Seems we could move 19 to 27, putting it before Serialization or we could tinker with the markup so that 19.8.4.x didn't appear in the ToC.

michaelhkay:
Yes, it's something I've got used to, but it is a little bit quirky. 

We could do something like the function finder in F+O, with a list of XSLT elements that appears *before* the table of contents.

ndw:
Well, yes, I'd considered that on several occasions. Now I've done it. (#1589 which I'm going to proactively merge as soon as it builds since (a) I can't imagine anyone objecting and (b) it's difficult to stage build changes.)

I *still* think moving the streamability chapter is a good idea. Or maybe move all of 19.8.4.x into an appendix?

ndw:
*shrug*

I didn't suggest that moving it was a perfect solution, simply that it would make some number of things better. I'm not going to die on this hill.

gimsieke:
It doesn't annoy me that much. You click on the first match link 'streamability of xsl:apply-templates' and you can be assured that there's a link to xsl:apply-templates in the heading that you're taken to. You reach your goal much faster than, for instance, in the interactive Saxon documentation where you need to skip many 'changes' sections before you discover the section you were looking for in the first place.

michaelhkay:
We agreed to close this at meeting 102

PR #1589, opened 2024-11-19
Implement an instruction/function finder in XSLT

ndw:
See also #1588


ndw:
Imma gonna ask for forgiveness instead of permission...

Issue #1610, opened 2024-11-26
Some cross references are incorrect


ndw:
If you look in, for example XSLT, at references to the "FO40" spec, they actually attempt to link the w3.org location where it would have been if it was a REC. Not sure what the fix is, but...

ndw:
I looked into this a little bit. I fixed (#1614) the case I happened to notice yesterday, that FO40 function references in the XSLT specification linked to the 3.1 F&O spec.

Looking around a little more, we have a bunch of machinery in place to make links to 2.0(!), 3.0, 3.1, and 4.0 specifications with different defaults in different contexts. Does it make more sense to yank all that out and make this set of specifications self contained? Are there any reasons to link back to previous versions of the specifications?

michaelhkay:
>Are there any reasons to link back to previous versions of the specifications?

Not with fine-grained links, I think. On the rare occasions it might be needed, one can always put in a section number by hand, since the specs are stable.

michaelhkay:
Talking of links, and links to functions in particular, it would be nice to use `<function>` markup consistently across the specs.

michaelhkay:
Can this now be closed?

ndw:
I think so.

PR #1614, opened 2024-11-27
Fix xfunction refs in XSLT

ndw:
I changed `xfunction` refs so that they point to the right URI for 40 functions.

(Partial fix for #1610 )


ndw:
The CG agreed to merge this PR without discussion at [meeting 101](https://qt4cg.org/meeting/minutes/2024/12-03.html).

PR #1630, opened 2024-12-03
Two minor corrections of `minus-sign` spec

GuntherRademacher:
Per #1250, `minus-sign` is now a string rather than a single character.

This change:

 - corrects that in one place where it was still said to be a character
 - changes the formulation from "represent" to "mark" a negative number.

Sorry for opening a branch in this repo. I did this accidentally, omitting the fork that I originally wanted to create.


michaelhkay:
Thanks for pointing these out. I think that the two sections concerned actually need more extensive changes to align them properly with the format-number specification (e.g. the description of the `percent` property also needs updating). I will attempt to formulate these.

michaelhkay:
Superseded by issue #1634

michaelhkay:
Superseded by #1638

PR #1663, opened 2024-12-19
Remove DTD/stylesheet distractions at the top of the schema

### LINKED ISSUE #374: Can't view the XSD for XSLT in the browser
If you attempt to open `https://qt4cg.org/specifications/xslt-40/schema-for-xslt40.xsd` in the browser (in Firefox), you'll get:

```
Error loading stylesheet: An unknown error has occurred (805303f4)
http://www.w3.org/2008/09/xsd.xsl
```

In a Chrome-derived browser I get a blank screen on which even the context menu doesn't work. Digging about in the inspect window leads me to

```
Unsafe attempt to load URL http://www.w3.org/2008/09/xsd.xsl from frame with URL
https://qt4cg.org/specifications/xslt-40/schema-for-xslt40.xsd. Domains, protocols and ports must match.
```

I conclude that the problem is trying to load the XSL for XSD from a different domain. Boo. I guess we should copy those stylesheets to `qt4cg.org`, or remove the stylesheet PI, or ignore the whole thing on the assumption that we'll eventually publish these specifications in some W3C location and the probem will go away. Maybe.

ndw:
We decided to close #374 without further action. This PR just cleans up the relevant schema file by removing the DTD and stylesheet related comment(s).

### DIFF
--- a/specifications/xslt-40/src/schema-for-xslt40.xsd
+++ b/specifications/xslt-40/src/schema-for-xslt40.xsd
@@ -1,15 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--* <!DOCTYPE xs:schema PUBLIC "-//W3C//DTD XMLSCHEMA 200105//EN"
-       "http://www.w3.org/2001/XMLSchema.dtd" [
-<!ENTITY % schemaAttrs "
-  xmlns:xs   CDATA #IMPLIED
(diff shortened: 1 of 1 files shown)


ndw:
The CG agreed to merge this PR without further discussion at [meeting 104](https://qt4cg.org/meeting/minutes/2025/01-07.html).

Issue #1685, opened 2025-01-09
Pipeline Operator


ChristianGruen:
This is an attempt to find a solution for the discussion in #755, which was originally about defining an expression to bind the context value. It serves as a summary for an upcoming PR.

We have two operators in the language that can be used for pipelining:
  1. With the simple map operator `!`, single items of an input can be bound to the context value.
  2. With the arrow operator `=>`, an input can be bound as first argument in a function call.

The current restrictions are:
  A) There is no way to bind a sequence with 0 or more than 1 items to the context value.
  B) We can only bind the input to the first function argument.

In addition, we have introduced the mapping arrow expression `=!>` to bind single items of an input to the first function argument.

We could generalize and simplify the situation by introducing a dedicated and very basic pipeline operator: `A -> B` evaluates `A` to a value, which is bound to the context value before evaluating `B`.

With the operator, restriction A) would be resolved. Restriction B) would be tackled indirectly, as `->` and `!` can often be combined. For example, the following examples from the specification could be simplified…

```xquery
(: current vs. simplified syntax :)
$s => tokenize() =!> fn { `"{.}"` }()
$s -> tokenize(.) ! `"{.}"`

(: current vs. simplified syntax :)
(1 to 5) =!> xs:double() =!> math:sqrt() =!> fn($a) { $a + 1 }() => sum()
(1 to 5) ! xs:double(.) ! math:sqrt(.) ! (. + 1) -> sum(.)
```

…and we could drop `=!>` in favor of the new operator.

An equivalent representation for the focus function `fn { E }` would be `fn($c) { $c -> E }`.

michaelhkay:
>We could generalize and simplify the situation by introducing a dedicated and very basic pipeline operator: A -> B evaluates A to a value, which is bound to the context value before evaluating B.

I'm strongly in favour of that.

> With the operator, restriction A) would be resolved. Restriction B) would be tackled indirectly, as -> and ! can often be combined.

The main problem is that they have different precedence. But we can live with that. In most of your examples it doesn't seem to be an issue.


> For example, the following examples could be simplified…
> (: current and simplified syntax :)
> $s => tokenize() =!> fn { `"{.}"` }()
> $s -> tokenize(.) ! `"{.}"`
> 
> (: current and simplified syntax :)
> (1 to 5) =!> xs:double() =!> math:sqrt() =!> fn($a) { $a + 1 }() => sum()
[...]

PR #1686, opened 2025-01-09
1685 Pipeline Operator

ChristianGruen:
Issue: #1685

The PR introduces the pipeline operator `->`. If we decide to add it, we could drop `=!>` in a second step and update various examples in the text.

### DIFF
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -851,7 +851,7 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
       <g:ref name="AssignmentExpr" if="scripting" lookahead="3"/>
       <g:ref name="ExitExpr" if="scripting" lookahead="2"/>
       <g:ref name="WhileExpr" if="scripting" lookahead="2"/>
-      <g:ref name="OrExpr" if="xpath40 xquery40"/>
+      <g:ref name="PipelineExpr" if="xpath40 xquery40"/>
     </g:choice>
   </g:production>
   
@@ -1413,6 +1413,11 @@ ErrorVal ::= "$" VarName
   <!-- [ start OrExpr etc -->
 
   <g:exprProduction name="OperatorExpr" node-type="void">
+    <g:level>
+      <g:binary name="PipelineExpr" condition="&gt; 1">
+        <g:string>-></g:string>
+      </g:binary>
+    </g:level>
     <g:level>
       <g:binary name="OrExpr" condition="&gt; 1">
         <g:string>or</g:string>
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -851,7 +851,7 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
       <g:ref name="AssignmentExpr" if="scripting" lookahead="3"/>
       <g:ref name="ExitExpr" if="scripting" lookahead="2"/>
       <g:ref name="WhileExpr" if="scripting" lookahead="2"/>
-      <g:ref name="PipelineExpr" if="xpath40 xquery40"/>
+      <g:ref name="OrExpr" if="xpath40 xquery40"/>
     </g:choice>
   </g:production>
   
@@ -1414,11 +1414,6 @@ ErrorVal ::= "$" VarName
 
   <g:exprProduction name="OperatorExpr" node-type="void">
     <g:level>
-      <g:binary name="PipelineExpr" condition="&gt; 1">
-        <g:string>-></g:string>
-      </g:binary>
-    </g:level>
-    <g:level>
       <g:binary name="OrExpr" condition="&gt; 1">
         <g:string>or</g:string>
       </g:binary>
@@ -1557,6 +1552,11 @@ ErrorVal ::= "$" VarName
           <g:ref name="Rbrace"/>
         </g:sequence>
       </g:postfix>
+    </g:level>
+    <g:level>
+      <g:binary name="PipelineExpr" condition="&gt; 1" if="xpath40 xquery40">
+        <g:string>-></g:string>
+      </g:binary>
     </g:level>
     <g:level>
       <g:postfix name="ArrowExpr" prefix-seq-type="*" condition="&gt; 1" if="xpath40 xquery40">
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -22086,14 +22086,16 @@ raised <errorref
          <example>
             <ulist>
                <item>
-                  <p>Tokenizes a string, creates a new string from the tokens and returns
-                     <code>"a/b/c"</code>:</p>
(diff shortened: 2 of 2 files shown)


michaelhkay:
I am in favour of adding this operator. My only reservation is about its operator precedence [Incidentally, it needs adding to the precedence table in A.5].

I can see the logic of a low precedence - the result of any expression can be piped into any other expression. But my instinct would have been to use the same precedence as `=>`, so that the two operators have equal weight and can be used interchangeably in a pipeline: `x => f(y)` can be replaced by `x -> f(., y)`.

ChristianGruen:
> I can see the logic of a low precedence - the result of any expression can be piped into any other expression. But my instinct would have been to use the same precedence as `=>`, so that the two operators have equal weight and can be used interchangeably in a pipeline: `x => f(y)` can be replaced by `x -> f(., y)`.

I believe the analog with `!` should be regarded as much more important, as the two operators can be expected to be frequently combined. If we made `->` and `=>` interchangeable, we could not write:

```xquery
'1 2' -> tokenize(.) ! number()
```

Instead, it would need to be:

```xquery
('1 2' -> tokenize(.)) ! number()
```

This seems counterintuitive, as this syntax implies that the parens are necessary to control the precedence, which is actually not the case.

To look at it the other way: Can we think of use cases in which it would be confusing if `=>` and `->` had a different precedence?

michaelhkay:
I guess the biggest difference between `->` and `=>` is that `=>` is not a regular binary operator - the RHS is not a general expression. So if we write `X => Y -> Z`, it doesn't matter what  the precedence of `->` is, it will inevitably parse as `(X => Y) -> Z`. So the question becomes, how should `X -> Y => Z` parse?

Most of the time, the result will be the same either way. For example `"1 2 3" -> tokenize(.) => count()` produces the answer 3 for both options. That's because the result of (X -> Y) is the result of Y. The only case I can see that's different is if the function call on the RHS of `=>` includes function arguments that depend on the focus, and I'm having trouble constructing a realistic example that does that.

So it seems to be a case rather like `A/B/C`, where it's only in very obscure edge cases that `(A/B)/C` and `A/(B/C)` will produce different answers.

I'm not sure I understand your example

`'1 2' -> tokenize(.) ! number()`

If `->` and `=>` had the same operator precedence, this would still be lower than `!`, so it wouldn't affect this example, surely?

The main difference is going to show when the `->` arrow is mixed with operators such as arithmetic. On your proposal:

`1 + 2 => math:sin()` means `1 + ( 2 => math:sin() )`

while

`1 + 2 -> math:sin(.)` means `(1 + 2) -> math:sin(.)`

I'm worried that the difference will lead to user confusion.

ChristianGruen:
> `'1 2' -> tokenize(.) ! number()`
> 
> If `->` and `=>` had the same operator precedence, this would still be lower than `!`, so it wouldn't affect this example, surely?

True. I rather wanted to point out that…

```xquery
'1 2' => tokenize() ! number()
```

…is not a valid syntax (it requires parens before the simple map operator), but that we should allow `'1 2' -> tokenize(.) ! number()`.

> `1 + 2 => math:sin()` means `1 + ( 2 => math:sin() )`
> while 
> `1 + 2 -> math:sin(.)` means `(1 + 2) -> math:sin(.)`
> I'm worried that the difference will lead to user confusion.

Yes, I see. I think I will change the grammar to:

```
CastExpr     ::= PipelineExpr ("cast" "as" CastTarget "?"? )?
PipelineExpr ::= ArrowExpr ( "->" ArrowExpr )*
```

michaelhkay:
```
CastExpr     ::= PipelineExpr ("cast" "as" CastTarget "?"? )?
PipelineExpr ::= ArrowExpr ( "->" ArrowExpr )*
```

Thanks. I think that's the best solution. (We might also want to rename ArrowExpr to reflect the functionality rather than the visual form of the operator. But finding a good name is difficult, so let's put that on one side).

ChristianGruen:
@djbpitt I have revised the example section, maybe you find it helpful: https://qt4cg.org/pr/1686/xquery-40/xpath-40-autodiff.html#id-pipeline-operator

djbpitt:
Thank you for sharing these. The XQuery alternatives are welcome, but where the same logic could, alternatively, be expressed in XPath using just simple map (`!`) and fat arrow (`=>`), it might be helpful to give those as well. This would foreground the situations that cannot be expressed in that way, that is, situations that demonstrate a need for a new thin arrow (`->`) operator.
---

```
'a b c' -> tokenize(.) -> count(.) -> concat('count=', .)
```

Except for the last step I think this could be expressed idiomatically as:

```
'a b c' => tokenize() => count()
```

The last step, though, makes a strong case because neither `!` nor `=>` could be used effectively. A version like:

```
('a b c' => tokenize() => count()) ! concat('count=', .)
```

works, but it requires parentheses around everything before the `!`, which is unnatural and hard to read.

---



(1 to 10) ! math:pow(2, .) -> sum(.)
Unless I've misunderstood something, this is equivalent to:
(1 to 10) ! math:pow(2, .) => sum()
This does illustrate that -> can be used here, which is enough to make it a correct example, but it doesn't seem to demonstrates why it's needed. I suppose it isn't necessary to demonstrate a need to end-users of the spec, but those who are familiar with 3.1 might nonetheless be most interested in uses that aren't handled well with ! and =>. For that reason, I'd suggest giving this fat-arrow alternative alongside the XQuery one, which would emphasize, at least implicitly, that the examples without such alternatives are those where the alternatives in question don't exist, or where they would be awkward and hard to read and understand.__
(1 to 4)
-> for-each(., op('+'))
-> for-each-pair(4 to 7, ., op('>'))
-> some(.)
If I replace the 4.0 op() functions with lambda addition and comparison functions and some() with count() just to get the code to run under 3.1, this seems equivalent to:
(1 to 4) => for-each(function($a) {$a + $a}) => for-each-pair(4 to 7, function($a, $b) {$a lt $b})
=> count()
The only significant change I had to make was to swap the order of the arguments to for-each-pair() and invert the comparison operator. An alternative to for-each-pair() in this example, one where the order of the arguments isn't under user control, at least in a natural way, might make a more persuasive argument for the need for a new -> operator.
The following also appears to work:
(1 to 4) 
! (. + .)
=> for-each-pair(4 to 7, function($a, $b) {$a lt $b})
=> count()__
$dictionary/word
-> (if(count(.) < 10) then . else (.[1 to 9], '…'))
-> string-join(., '; ')
I like this last example because I don't think there's a way with just ! and => to pass a sequence into count() when the count() function is inside an "if" clause.
I hope this is helpful.
Best,
David
On Thu, Jan 16, 2025 at 8:53 AM Christian Grün <notifications@github.com> wrote:
[@djbpitt](https://github.com/djbpitt) I have revised the example section, maybe you find it helpful: https://qt4cg.org/pr/1686/xquery-40/xpath-40-autodiff.html#id-pipeline-operator—

ChristianGruen:
Thanks for the feedback; it is good to know that the revised examples are helpful.

The alternative syntax should be both valid XPath and XQuery (but it may be counterintuitive that `for` and`let` are also available in XPath). I omitted `!` and`=>` in this syntax, as I felt that the presentation is more consistent this way; but I could certainly add a second alternative to some examples.

PR #1701, opened 2025-01-14
Add dedication to MSM (action QT4CG-088-01)

ndw:
I've had this action on my plate for a while. Having written a dedication, there's a follow-up question of where to put it. Having it in only one specification isn't wrong, but it seems slightly odd given that MSM contributed to them all. In the end, I decided to put a full dedication in the XPath specification and link to it from the others. 

My rationale for the XPath spec is that it's probaly one that everyone reads. Another possibility was the Data Model as it's "foundational" but I think it's less read than XPath. 

The published PR won't be write because there are tooling changes required. I've attached a couple of screen shots, one of the full dedication in XPath:

------
Screenshot 2025-01-14 at 14-00-08 XML Path Language (XPath) 4 0 WG Review Draft
------

And another of the link from the other specs (from XSLT, I think, but they're all the same).

------
Screenshot 2025-01-14 at 13-59-08 XSL Transformations (XSLT) Version 4 0
------

### DIFF
--- a/etc/status-general.xml
+++ b/etc/status-general.xml
@@ -71,4 +71,10 @@
 <!-- ************************************************************************** -->
       &patent-policy-paragraph;
 
+<note role="dedication" id="dedication">
+<p>The publications of this community group 
+<loc href="../xquery-40/xpath-40.html#dedication">are dedicated</loc> to our co-chair,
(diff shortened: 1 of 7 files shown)


ChristianGruen (approved):
Thank you.

PR #1749, opened 2025-01-31
Don't fix the function finder on small devices

ndw:
This is also related to 1747. It "fixes" the problem that @ChristianGruen reported where the function finder obscured content on mobile (narrow) devices. I've changed things so that it isn't at a fixed location on narrow devices. It still appears above the ToC, but it scrolls as normal.


ChristianGruen:
Great

ndw:
The title 'Don't fix the function finder' was...confusing!

PR #1756, opened 2025-02-03
Make DeltaXML diffs on the main build too

ndw:
This PR should build DeltaXML diffs of the EXPath specs...and when merged, should build them on the main build as well.


ChristianGruen:
@ndw Looks promising. Should we already merge it?

ndw:
I will in a moment. I'm using this PR to test parts of the PR build process 😄

PR #1760, opened 2025-02-03
Remove hand-generated diffs from the builds

### LINKED ISSUE #1757: Build cleanup: remove the "by hand" diffs
Unless I'm mistaken, the 'by hand' diffs, the ones that are created from explicit diff markup added by the editors, have not been consistently maintained for some time.

We still have places that point to them, and I think this could be confusing.

I propose that we pull all of that machinery out and remove references to them.

ndw:
Fix #1757 

The PR build isn't going to be very informative, but I'll leave this one open in case anyone wants to review the source code diffs.

I did not attempt to remove the XML markup from the specs. Perhaps we should, but I think we'd want to manage that carefully to avoid an absolute mountain of merge conflicts.


ndw:
The CG agreed to merge this PR without discussion at [meeting 108](https://qt4cg.org/meeting/minutes/2025/02-04.html).

PR #1769, opened 2025-02-05
Add links from processing model diagrams

ndw:
Completes action QT4CG-108-02

I’ve added link targets where necessary. I didn’t try to link closer than that paragraph level, partly because I think that’s the context the reader needs, but also partly because we don’t copy ID values from all elements.

There’s no definition of DM4.

(Review of the link targets and comments on what (if anything) the remaining boxes and labels should link to most appreciated.)


Arithmeticus:
In the explanatory paragraph below figure 1, we are promised "an area labeled external processing," but there is no such label in the figure. 

Why the two alternative paths for DM1? The prose for DM1 doesn't describe an unmediated path from XML to XDM. Is the Infoset missing in the DM1 portion of the diagram?

DM4 lacks a link.

I'm uncertain whether the node shapes are consistent. For example, it makes sense that Host Environment and Execution Engine might share the same shape, but why does Other/Direct Generation have that same shape? Especially since the prose account suggests that Other/Direct Generation is an alternative counterpart to the XML shape that is the source of DM1. I propose that "Other/Direct Generation" is a descriptor of the activity described by edge DM3, and not its source (or the subject of DM3).

It is hard for me to correlate the prose description for DQ1-5 with the diagram. Each one muddles the other, to my mind.

Are "operation tree" and "abstract syntax tree" interchangeable terms?

Is SQ3 intentionally missing?

I also think that the edges/arrows should be explained or glossed, if possible by means of a transitive verb construction. But that might run into problems. Below is a list of my attempt to interpret the meaning of each of the labeled edges, with X representing the start of the arrow and Y its target.

DM1/2/3/4: X is converted [by what?] into Y
SQ1: X is parsed [by what?] into Y
SI1/SQ2/DQ3: X provides Y
DQ1: X is depended upon by Y
DQ2: ??
DQ4: X creates Y (but that doesn't make sense, because it's a double-headed arrow -- what's that about?)
DQ5: ?? (double-headed arrow, meaning?)
SQ4: ??
SQ5: X is normalized into Y

Hopefully some of these comments are useful for improving the diagram and prose.

michaelhkay:
>Hopefully some of these comments are useful for improving the diagram and prose.

It seems to me that redrawing the diagram to use a different graphics technology is one thing, rewriting the processing model section so it actually makes sense is a different undertaking altogether.

ndw:
> In the explanatory paragraph below figure 1, we are promised "an area labeled external processing," but there is no such  label in the figure.

Fixed.

> Why the two alternative paths for DM1? The prose for DM1 doesn't describe an unmediated path from XML to XDM. Is the Infoset missing in the DM1 portion of the diagram?

Fair point. I think I was trying to separate the schema-validated and not-schema-validated paths. But the original diagram lumped them together by marking the intermediate box Infoset/PSVI, so I've done the same.

> DM4 lacks a link.

I persuaded myself last week that DM4 didn't occur in the prose. I was wrong, so I've added the link. (I assume I got distracted by all of the occurrences of "DM40" in the prose.)

> I'm uncertain whether the node shapes are consistent. For example, it makes sense that Host Environment and Execution Engine might share the same shape, but why does Other/Direct Generation have that same shape? Especially since the prose account suggests that Other/Direct Generation is an alternative counterpart to the XML shape that is the source of DM1. I propose that "Other/Direct Generation" is a descriptor of the activity described by edge DM3, and not its source (or the subject of DM3).

I'm happy to take suggestions. My choices were:

1. Notes, little boxes with folded corners, for the "documenty type" things.
2. Rectangles for the things that are "objrects" in the processing model.
3. Rounded boxes for processes and abstractions
4. 3d boxes for the contexts because the original diagrams show them with additional lines.

> It is hard for me to correlate the prose description for DQ1-5 with the diagram. Each one muddles the other, to my mind.

My action was to sort out the diagrams. I haven't attempted to improve the prose.

> Are "operation tree" and "abstract syntax tree" interchangeable terms?

That's a good question.

> Is SQ3 intentionally missing?

That's about loading modules in XQuery so it only appears in the XQuery specification.

> I also think that the edges/arrows should be explained or glossed, if possible by means of a transitive verb construction. 

It's true that there are some additional words on the arrows in the original diagrams. I'm not sure how best to do that. Just adding all those words to each arrow is going to make the diagram messy. Maybe it'd be possible to have a little legend. But at least the edges link to the prose now.

> Hopefully some of these comments are useful for improving the diagram and prose.

Very useful, thank you. But tackling the prose is a different task.

ndw:
At the meeting last week, I was asked to add JSON to the diagrams and try again to sort out the tooltips.

I did both of those on Friday. We've looked at the diagrams twice, I'm guessing we can merge this without further discussion, but, you know, shout if I'm mistaken.

ndw:
The CG agreed to merge this PR without discussion at [meeting 110](https://qt4cg.org/meeting/minutes/2025/02-18.html).

PR #1835, opened 2025-02-24
add zero-width assertions to regular expressions

liamquin:
Proposal for issues !998 and !1006 to add zero-width assertions - lookahead, lookbehind, and word boundary.

Word boundaries use the already-defined \w and \W from XML Schema.

The syntax for lookahead and lookbehind assertions supports the two most common variants, one using < and > and the other using `(*positive_lookahead:expr)`, which is at least amenable to Web searches, and doesn’t need escaping in XSLT or XQuery.

Note that word boundary \< \b \B \> assertions can be rewritten in terms of lookahead and lookbehind assertions.

Perl has a more powerful form of \b and \B that can match grapheme clusters, the Unicode linebreaking algorithm, and more, but supporting that would require language and script based mechanisms; if the graphemes() function is added, it would be worth considering. For now, i made it an error to write \b{...} so that the support could be added later if wanted, and also so that copying regular expressions into XPath would raise an error for the unsupported feature.

I will reopen !998 - if this is accepted i can produce test cases. Of course, i’m also happy to edit/rewrite etc.  The syntax is widely supported, although \K is i think not in libpcre (but, libpcre has looser restrictions on negative backward assertions).

### DIFF
--- a/specifications/xpath-functions-40/src/xpath-functions.xml
+++ b/specifications/xpath-functions-40/src/xpath-functions.xml
@@ -3351,6 +3351,205 @@ It is recommended that implementers consult <bibref ref="UNICODE-TR18"/> for inf
                   </p>
                   
                </div4>
+               <div4 diff="add" at="issue998" id="regex-look-assertions">
+                  <head>Lookahead and Lookbehind Assertions</head>
+               
+		   <!--* There is something to be said for what some libraries do,
+		       * and documenting all the zero-width assertions together,
+		       * including ^ ! \b \B \< \> and so on, as well as  lookahead/behind.
+		       * But that would not be a clean change. - LQ
+		       *-->
+		   <p>A lookahead assertion allows matching of a pattern only when followed by a match to the assertion.
+		   Similarly, a lookbehind assertion matches something only when it comes immediately after another item.
+		   The input matches by an assertion is not included in the match, and for this reason in some systems
+		   these are considered examples of zero-width assertions.</p>
+
+		   <p>It is conventional in regular expression languages to provide multiple forms of asserrtions.
+		   One form uses an abbreviated syntax, and another a longer more readable version.  Some languages
+		   provide additional forms not included here. The short forms, although most widely supported and described
+		   elsewhere, use <code>&lt;</code> and <code>&gt;</code> characters which can be inconvenient in XML-based
+		   host languages such as XML, where those characters must be escaped. In those cases, the longer forms are
+		   likely to be more convenient; in all cases the abbreviated form and the long form are identical in function
+		   and entirely interchangable.</p>
+
+		   <p>There are some restrictions on lookbehind assertions. Lookbehind assertions must not contain capture groups.
+		   Implementations MAY firther restrict the pattern, forbidding occurrence indicators
+		   such as <code>*</code>, <code>?</code>, <code>+</code>, or <code>{...}</code>,
+		   and may limit the length of what is matched by a lookbehind assertion to, for example, 100 characters.
+		   A positive lookbehind that matches a single character, such as <code>\s</code>, MUST be supported.
+		   See also <code>\K</code> for a way to avoid these restrictions.</p>
+
+		   <p>In streaming modes, Lookahead and lookbehind assertions must only be applied to input that is grounded.</p>
+
+
+		   <div5 id="regex-positive-lookahead">
+		       <head>Positive Lookahead Assertions</head>
+
+		       <p><code>a(?=b)</code></p>
+		       <p><code>a(*positive_lookahead:b)</code></p>
+
+		       <p>A pattern <code>a</code> followed by a positive lookahead assertion <code>b</code>
+		       matches <code>a</code> only if the lookahead assertion <code>b</code> also matches.</p>
+
+		       <p>The two forms are equivalent. The longer form is provided partly for clarity and partly
+		       because the conventional character <code>&lt;</code> for this purpose must be escaped in XML-based
+		       host languages such as XSLT.</p>
+
+		       <p>Input matched by an assertion is not itself included in the match unless it is explicitly
+		       matched. However, Capture buffers within positive lookahead assertions are populated with
+		       whatever they match.</p>
+
+		       <p>Examples:</p>
+
+		       <p><code>boy(?=girl)</code> will match <code>boy</code> only if immediately
+		       followed by the string <code>girl</code>.</p>
+
+		       <p><code>replace("high school antics", "high(?=\s*school)", "low")</code> results
+		       in <code>low school antics</code>.</p>
+
+		       <p><code>replace("crawl, crawling, crawled", "(crawl|fly)(*positive_lookahead:ing)", "leap")</code>
+		       produces <code>crawl, leaping, crawled</code>.</p>
+
+		       <p><code>replace("very high towers sway", "high(?=\s+towers)", "tall")</code></p>
+		       <p>This results in <code>very tall towers sway</code>. The pattern matches “high” when it is
+		       followed by one or more space characters and then the word “towers”, but
+		       only “high” is replaced because the remainder, which part of the lookahead
+		       assertion, is not captured.</p>
+
+		       <p><code>replace("very high buildings sway", "high(?=\s+towers)", "tall")</code></p>
+		       <p>This will not replace anything because although <code>high</code> is matched,
+		       the lookahead portion does not match, which means that the pattern as a whole fails to match.</p>
+
+		       <p><code>replace("very high towers", "high\s(?=(a-z][a-z+)", "tall $1-$1-")</code></p>
+		       <p>This results in <code>very tall t-t-towers</code>, replacing only the word <code>high</code> and
+		       the following space.</p>
+
+		       <note><p>Note: a positive lookahead assertion can be thought of as a gernalization of <code>$</code> at
+		       the end of a regular expression: assertions can occur anywhere,
+		       not only at the end, and can match arbitrary content.</p></note>
+		    </div5>
+		    <div5 id="regex-positive-lookbehind">
+		       <head>Positive lookbehind Assertions</head>
+		       <p><code>a(&lt;=b)</code></p>
+		       <p><code>a(*positive_lookbehind:b)</code></p>
+
+		       <p>Lookbehind assertions provide a way to match expressions only if they appear in the input immediately
+		       after a match of the expression inside the assertion. As with other assertions, positive lookbehind assertions
+		       do not affect <emph>what</emph> is matched, but only <emph>whether</emph> it matches.</p>
+
+		       <p>Example:</p>
+
+		       <p><code>replace("are we lower", "(?&lt;=\s)we", "they")</code></p>
+		       <p>This results in <code>are they lower</code>,
+		       replacing only the occurence of “we” preceded by a space,
+		       but not the characters “we” inside the word “lower”.
+		       The space character within the assertion is not consumed or captured.</p>
+
+		       <note><p>The <code>&lt;</code> must be escaped in XML-based host languages such as XSLT,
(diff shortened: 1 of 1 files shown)


ndw:
@liamquin , there appears to be a markup error in the PR that prevents it from building. Could you fix that, please?

liamquin:
Build locally succeeded overall but i see
> **Task :fo_html**
specdoc is FO40
Error: cannot resolve id-arithmetic-expressions in XP40 at id=op.numeric

I also see that fo:merge succeeded locally.
Parse succeeded (0.698) with no errors and no warnings

@ndw any ideas where to look? 

It’s true i get
> Task :binary_generate_tests
Cannot generate tests from specification; no Saxon license available.
Looks like it downloads Saxon-EE-12.2.jar, wonder if Saxon-EE 9 will work.

liamquin:
Fixed. Sorry for comment flood, i was too tired last night - it helps if i commit the change!

michaelhkay:
Generally this looks reasonable. I'll comment on the technical substance rather than editorial issues (there are many spellling mistakes, etc).

I think we can stick to one syntax, probably the cryptic one which has the virtue that regex afficionadoes know it well.

I think we can do without \K.

I would prefer to be completely prescriptive about what is allowed in a lookbehind, and to make the rules restrictive, certainly limiting it to a fixed width.

The proposed specification defines new constructs such as lookahead assertions, but as far as I can see it doesn't integrate these into the grammar, it doesn't say where they can appear or how they can be combined with other constructs (such as occurrence indicators). Compare with other XPath extensions to the XSD regex grammar which explicitly modify the XSD-defined grammar productions.

The specification seems to be a bit fuzzy in places. For example, consider this:

> Lookbehind assertions provide a way to match expressions only if they appear in the input immediately after a match of the expression inside the assertion. As with other assertions, positive lookbehind assertions do not affect what is matched, but only whether it matches.

The first sentence asserts that the construct "provides a way" to do something, but it leaves the reader to guess at exactly how it does so. That's a bit like saying that "-" provides a way to subtract numbers without actually saying that it subtracts the RH operand from the LH operand. The second sentence again skirts around a precise statement of the semantics. What's missing is definitive statements such as "a substring S matches `X(?=Y)` if S matches X and S is immediately followed by a substring T that matches Y`.

I'd like to be confident that someone is going to write comprehensive tests...

liamquin:
Thank you for a detailed consideration! Yes, for sure i’m happy to continue editing, fix typos etc.

I included two syntaxes because they are both widely supported (along with a third one that i think we can ignore), and because the “human-readable” versions cater much better to people maintaining code, when the people don’t use this feature often. But would like to hear from others about that too, we can drop the long forms. There are long forms for some other features where the XPath expressions only have the short variant, so there’s certainly precedent for omitting them.

On fixed width, yes, it helps implementors; how do you feel about an alternation of literals?

```
(*negative_lookbehind:boy’s|girl’s)\s+shoes
```
?

(libpcre and i think perl now both support that, and it was proposed for Swift too, and i think JS does but would have to check)

Adding the extensions to the XSD grammar is fine, along with making the proposal more precise. That, along with tests, is the kind of thing i wasn’t going to do until there seemed a possibility the proposal might be accepted.

On \K, i don’t have strong feelings, and am happy to withdraw it.

Thanks!

michaelhkay:
>On fixed width, yes, it helps implementors; how do you feel about an alternation of literals?

Yes, I think that's fine, so long as the rules are simple.

Happy to take input from others about the long forms. I hadn't come across them before.

Note, in #1848 I point out that we've been rather cavalier in extending `charClass` to include `$` and `^` which do not actually denote a set of characters as the definition of `charClass` implies. We may need to do some rearrangement of existing constructs in order to define a simple rule for what it means for a regex to be fixed-length.

specifications/xpath-functions-40/src/xpath-functions.xml
```diff
@@ -3351,6 +3351,227 @@ It is recommended that implementers consult <bibref ref="UNICODE-TR18"/> for inf
...
+		    <p><code>\&gt;</code> matches between a word character (<code>\w</code>) and a non-word character (<code>\W</code>);
+		    conceptually it matches at the end of a word. <code>\&gt;</code> also matches at the end of the input,
+		    if the last input character matches <code>\w</code>: that is, it matches at the end of the last
+		    human-language word in the input.</p>
+
+		    <p><code>\b</code> matches between two characters when there is a word character on
+		    one side and a non-word character on the other, in either order.</p>
+
+		    <note><p><code>\b</code> can be rewritten to an equivalent form in terms of lookbehind and lookahead assertions:</p>
+		    <p><code> ( (*positive_lookbehind:\w)(*positvie_lookahead:\W) ) |
```
ChristianGruen, line 3543:
```suggestion
		    <p><code> ( (*positive_lookbehind:\w)(*positive_lookahead:\W) ) |
```

ChristianGruen (commented):
Thanks, Liam, for the PR. I was happy to see that it also includes `\b`.

In my review, I have focused on minor typos. Other observations:
* With regard to `(*positive_lookbehind:`: Is this syntax used in any other language? If no, I would suggest using `(?lt=` as alternative, as `<` and `lt` are often used interchangeably (even if their semantics may differ). An alternative could be `(?positive-lookbehind:`, which keeps `?` and uses the more common hyphen syntax.
* As `<` causes a challenge in XSLT, U+FF1C (＜) was introduced as syntactical alternative (https://qt4cg.org/specifications/xquery-40/xpath-40.html#id-lt-and-gt-characters). It could also be allowed here.
* I could also live without `\K`. The lookahead/-behind syntax is maybe not the most intuitive one, but it has become very common, and it may be better to reduce the number of features that do similar things.

liamquin:
Thanks for reviewing!

I included \b \< \> \B here rather than in a separate PR because they are also zero-width assertions, and use \w and \W which are already in XSD.

The short syntax is in most modern regular expression variants of course, including Java, JavaScript, .net, etc., strting with Perl 5 in 1994.

The long syntax i think was introduced by Perl, and is also in pcre, but i see it also in Swift, and some other places. I didn’t invent any syntax in this proposal. I don’t have strong feelings, but when it was discussed before, there was talk of inventing something that did not use `<` so i thought it better to use something already in use elsewhere. It’s also more amenable to Web searches.

On constant strings for lookbehind, i see regular-expressions.info says, 

> While Perl requires alternatives inside lookbehind to have the same length, PCRE allows alternatives of variable length. [PHP](https://www.regular-expressions.info/php.html), [Delphi](https://www.regular-expressions.info/delphi.html), [R](https://www.regular-expressions.info/rlanguage.html), and [Ruby](https://www.regular-expressions.info/ruby.html) also allow this.

(however, i think recently Perl also dropped that restriction)

so allowing an alternation of fixed-length patterns that can be differing lengths seems OK

I’ll go through your review in detail along with addressing Mike Kay’s concerns, but probably not for a couple of days. Thanks!

ChristianGruen:
> The long syntax i think was introduced by Perl, and is also in pcre, but i see it also in Swift, and some other places.

A stupid question: Do you have any web reference? I didn't find any example for the `(*positive_lookbehind` syntax online.

liamquin:
> A stupid question: Do you have any web reference? I didn't find any example for the `(*positive_lookbehind` syntax online.

https://man7.org/linux/man-pages/man3/pcre2syntax.3.html

https://perldoc.perl.org/perlre#Extended-Patterns (some way down)

The proposal that got them added to the Swift language (in 5.7 apparently)
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0355-regex-syntax-run-time-construction.md

specifications/xpath-functions-40/src/xpath-functions.xml
```diff
@@ -3351,6 +3351,227 @@ It is recommended that implementers consult <bibref ref="UNICODE-TR18"/> for inf
...
+		   The input matches by an assertion is not included in the match, and for this reason in some systems
+		   these are considered examples of zero-width assertions.</p>
+
+		   <p>It is conventional in regular expression languages to provide multiple forms of asserrtions.
+		   One form uses an abbreviated syntax, and another a longer more readable version.  Some languages
+		   provide additional forms not included here. The short forms, although most widely supported and described
+		   elsewhere, use the <code>&lt;</code> character, which can be inconvenient in XML-based
+		   host languages such as XML where those characters must be escaped. In those cases, the longer forms are
+		   likely to be more convenient; in all cases the abbreviated form and the long form are identical in function
+		   and entirely interchangable.</p>
```
Nemris, line 3373:
```suggestion
		   and entirely interchangeable.</p>
```

specifications/xpath-functions-40/src/xpath-functions.xml
```diff
@@ -3351,6 +3351,227 @@ It is recommended that implementers consult <bibref ref="UNICODE-TR18"/> for inf
...
+		       <p><code>replace("high school antics", "high(?=\s*school)", "low")</code> results
+		       in <code>low school antics</code>.</p>
+
+		       <p><code>replace("crawl, crawling, crawled", "(crawl|fly)(*positive_lookahead:ing)", "leap")</code>
+		       produces <code>crawl, leaping, crawled</code>.</p>
+
+		       <p><code>replace("very high towers sway", "high(?=\s+towers)", "tall")</code></p>
+		       <p>This results in <code>very tall towers sway</code>. The pattern matches “high” when it is
+		       followed by one or more space characters and then the word “towers”, but
+		       only “high” is replaced because the remainder, which part of the lookahead
```
Nemris, line 3416:
```suggestion
		       only “high” is replaced because the remainder, which is part of the lookahead
```

specifications/xpath-functions-40/src/xpath-functions.xml
```diff
@@ -3351,6 +3351,227 @@ It is recommended that implementers consult <bibref ref="UNICODE-TR18"/> for inf
...
+		    characters or both non-word characters. In other words, it matches a non-boundary.</p>
+
+		    <p>The combination of <code>\b</code> or <code>\B</code>, immediately followed by <code>{</code>,
+		    is a syntax error, <errorref class="RX" code="0002"/>.
+		    This is for compatibility with other systems that assign a meaning to <code>\b{...}</code>.</p>
+
+		    <!--* This is so that people who copy/paste complex expressions don't get different
+		        * semantics - should we say that more explicitly?
+			* E.g. in Perl (but not libpcre yet) \p{gcb} matches a Unicode grapheme cluster boundary.
+			* This has been requested as a feature, but a lot of underlying mechanism would be needed to
```
Nemris, line 3557:
```suggestion
			* This has been requested as a feature, but a lot of underlying mechanisms would be needed to
```

michaelhkay:
Some thoughts on integrating this.

I would be inclined to change the definition of `atom` to

`atom | ::= | NormalChar | charClass | assertion | ( '(' regExp ')' )`

and to take `$` and `^` out of `charClass` and move them into `assertion`, where the new zero-width assertions also belong.

I think we've got to the point where it would be useful for F&O to include a recap of the entire extended regex grammar.

michaelhkay:
Liam, would it be OK for me to take over at this point and do the integration?

liamquin:
If you’re willing, for sure; it’s also on my list for this week now i seem to have some health back.

liamquin:
Oh - a agree with you about ^ and $ as assertions; some other regexp libraries group these together too.

I have two other possible change proposals in this area (but got stopped by illness, sorry) - one is to propose \u \U \l \L \E for replace() (i wanted to see if \< and \b in terms of \w were accepted first, and seems it's OK), and the other is just to ask if we want \X to match an eXtended grapheme cluster (as per pcre and perl, but i think some others now too). But none of these would affect a reorganization and recap.

I’ll be out of the office on Monday for medical tests, but should be back later on Tuesday & can start looking at all this if it'd be helpful. I don't want to dump a ton of work on other people. I can also write a bunch of tests.

michaelhkay:
Superseded by #1856.

PR #1838, opened 2025-02-24
1820 Attempt to add change markup in collapsed ToC

### LINKED ISSUE #1820: Delta markers in collapsed TOC
When the TOC is shown in collapsed mode, it would be nice to promote the Δ change markers to the level where they become visible.

ndw:
Fix #1820 

This PR updates the styling so that a small "Δ" is added to the expand arrow when there are changes or additions in the concealed subsections. It's smaller and not blue. I could argue that this is on purpose so that the marking is different and perhaps more subtle. But the truth is, it was just easier to add the Δ without any markup that would make it larger or blue.

I've opted to conceal the Δ when the ToC is "open" on the grounds that you can see what is or isn't marked new on the revealed subsctions.

### DIFF
--- a/specifications/js/toc.js
+++ b/specifications/js/toc.js
@@ -1,4 +1,38 @@
 (function() {
+  const updateDelta = function(target) {
+    if (target.classList.contains("expanded")) {
+      return;
+    }
+
+    let anchor = target;
+    while (anchor && anchor.tagName !== "A") {
+      anchor = anchor.previousSibling;
+    }
+    if (anchor && (anchor.querySelector("span.toc-chg")
+                   || anchor.querySelector("span.toc-new"))) {
+      return;
+    }
+
+    let delta = null;
+    let details = target;
+    while (details && details.tagName !== "DETAILS") {
+      details = details.parentNode;
+    }
+    if (details) {
+      delta = (details.querySelector("span.toc-chg")
+               || details.querySelector("span.toc-new"))
+    }
+
+    // We're assuming innerHTML is a single text node...
+    let inner = `${target.innerHTML}`;
+    inner = inner.substring(inner.length - 1)
+    if (delta) {
+      inner = "★" + inner;
+    }
+
+    target.innerHTML = inner;
+  }
+
   const updateToc = function(open) {
     document.querySelectorAll(".toc details").forEach(details => {
       details.open = open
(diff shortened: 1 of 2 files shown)


ndw:
Happily, the PR build for this works. Take a look and see what you think.

I've also fixed another bug where nested sections sometimes don't get the right kind of arrow. That's not in this PR build because it's part of the stylesheet. So if testing this causes you to notice that, just ignore it until you see if merging this fixes it. (Assuming we merge this, of course.)

ChristianGruen:
I like it! 

The excessive use of triangles could be a bit confusing. Now that we have ➕ for new features, what about using one of the following symbols for updated features, instead of the ▵ variants?

* ★ (U+2605)
* ✨ (U+2728)
* ♻ (U+267B)
* 🆙 (U+1F199)

Could ➕ be used in the collapsed representation if at least one subitem is new and the alternative symbol otherwise?

ChristianGruen:
Maybe like this? [js-css.zip](https://github.com/user-attachments/files/18967383/js-css.zip)

ChristianGruen:
One more thing that I noticed (and didn’t fix so far): The collapse/expand icons are only changed if you click on them directly. They keep unchanged when clicking somewhere to the right of the text.

ChristianGruen:
And a last comment (promised): Maybe we can completely abandon horizontal indentation for subitems. The numbering already indicates on which level we are. I agree that has nothing to do anymore with the original intent of this PR ;·)

ndw:
> One more thing that I noticed (and didn’t fix so far): The collapse/expand icons are only changed if you click on them directly. They keep unchanged when clicking somewhere to the right of the text.

I think this is the other bug fixed in my PR that you can't see in the preview build.

ndw:
> And a last comment (promised): Maybe we can completely abandon horizontal indentation for subitems. The numbering already indicates on which level we are. I agree that has nothing to do anymore with the original intent of this PR ;·)

See #1849

ndw:
No, you're absolutely right about the clicking problem. Will look into it.

The biggest problem I have with ✨ is that it's yellow, and appears always to be yellow, which I find almost impossible to see on white and very hard to see on the pale gray ToC background.

michaelhkay:
>I liked 🆙; 

I didn't. I find it quite hard to read and it's not obvious what it means.

>or we could use ★ (currently black)

I like that, it's very distinctive. I agree that the delta and triangular arrow characters are too similar.

ndw:
1. I fixed the problem where clicking on the summary (but not the graphic) expanded the ToC without changing the graphic
2. I switched to ★ for the icon next to the graphic

There's a bug in Firefox where the star and the arrow get reduced to 80% twice. That doesn't happen in Chrome or Safari so I think it's a FF bug.

ndw:
To me, that feels like trying to squeeze more detail into the iconography than it really supports. I think it's sufficient to distinguish between "move along, there's nothing new under here" and "you might want to look inside this one." But I don't feel strongly about it, if consensus is we should try to do more than that.

ndw:
At [meeting 112](https://qt4cg.org/meeting/minutes/2025/03-04.html), the CG agreed to merge this PR without discussion.

PR #1839, opened 2025-02-24
Relax the return type of the Invisible XML parsing function

### LINKED ISSUE #1796: Allow fn:invisible-xml to return a function that returns an item()
Our current `fn:invisible-xml` function returns a document node. That makes perfect sense when held up against the Invisible XML specification. But I wonder if we should leave the door open to some extensibility. I can imagine, for example, an implementation of an Invisible XML processor that has the ability to return a map or even a CSV structure instead of XML. (XML is the required, standard result in 1.0 but implementors have been known to offer user options to produce other serializations and one area of potential change in the future is other serialization formats.)

Pro: more extensible. Con: less type information about the result.

ndw:
Fix #1796 

This change does not appear to change any test results. (In other words, none of our tests checked that the return type was explicitly a document node.)

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -33715,7 +33715,7 @@ let $scan-right := function(
     
    <fos:function name="invisible-xml" prefix="fn">
       <fos:signatures>
-         <fos:proto name="invisible-xml" return-type="fn(xs:string) as document-node()">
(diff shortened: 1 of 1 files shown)


ChristianGruen:
Maybe even `item()*`?

ndw:
Practically speaking, we're talking about the output from a parse: an XML document, a text document, a map, or an array (maybe).

I considered `item()*` and couldn't think of a use case for returning a sequence and  I don't think the function should be allowed to return an empty sequence (what could that possibly mean?).

PR #1840, opened 2025-02-24
GH action remove-label-on-reopen.yml

pdaengeli:
In response to the mailing list post by @ndw: https://lists.w3.org/Archives/Public/public-xslt-40/2025Feb/0024.html

This is untested, but it might at least serve as an inspiration how to avoid the unwanted tag in an automated manner.


ndw:
Yes, that works. Thank you again!

Issue #1747, opened 2025-01-31
Function finder is broken


ndw:
The function finder in F&O (and elsewhere) is broken. I believe that it uses the ToC to find the link target and now that the ToC structure has changed, it's failing.

ChristianGruen:
I noticed that the function finder hides parts of the text on mobile devices when the page is scrolled down. Maybe this could be fixed as well.

<img src="https://github.com/user-attachments/assets/236a4ebe-a888-47a2-b15e-0a5c2367ebbd" width="200"/>

ndw:
I'll take a look. That's the sort of thing I was alluding to in the previous issue where I said we needed to investigate the accessibility of the specs before we try to publish final ones.

ChristianGruen:
Because it is so much easier to demand than to deliver: With the new layout, the available space is not well spent. Would it be possible to reduce the horizontal indentation?

![Image](https://github.com/user-attachments/assets/5b4fcf1c-1d70-4cbe-8376-9b274ca19963)

ndw:
CSS is not my strongest suit. The current formatting works by making the number a fixed width. It's fairly wide in order to account for things like `17.4.9` but, as you observe, that's not *actually* wide enough. 😦 

Rather than attempt to change the existing CSS, I'll take a stab at starting from scratch. But probably not today.

ChristianGruen:
I think we can close those issue; the current rendering is great.

PR #1849, opened 2025-02-26
Reduce the indentation in the ToC

ndw:


michaelhkay:
I'm not sure I prefer it. What about trying a very small indent?

ndw:
Taking @michaelhkay 's suggestion, I've reworked this a little bit. Instead of having no indentation, there's a small amount of indentation. I also arranged hanging indents so that when the titles wrap, they are further indented.

It's impractical[*] to make the indentation work such that the wrapped lines are aligned to the right of the section number. But at least they're indented some.

[*] For the specific definition of "impractical" that means impractical for someone with my level of CSS skill to do it in the amount of time I was willing to devote to task.

ChristianGruen:
Fixes #1747 (in particular, https://github.com/qt4cg/qtspecs/issues/1747#issuecomment-2626917560).

ndw:
At [meeting 112](https://qt4cg.org/meeting/minutes/2025/03-04.html), the CG agreed to merge this PR without discussion.

Issue #322, opened 2023-01-26
Map construction in XSLT: xsl:record instruction


michaelhkay:
Constructing maps in XSLT often involves code rather like this:

```
            <xsl:map>
               <xsl:map-entry key="'author'" select="string(AUTHOR)"/>
               <xsl:map-entry key="'title'" select="string(TITLE)"/>
               <xsl:map-entry key="'price'" select="xs:decimal(PRICE)"/>
               <xsl:map-entry key="'publisher'" select="string(../@name)"/>
           </xsl:map>
```
The alternative using XPath is also rather ugly:

```
<xsl:sequence select="map{'author': string(AUTHOR),
                                                 'title':string(TITLE), 
                                                 'price': xs:decimal(PRICE), 
                                                 'publisher':string(../@name)}"/>
```
(the fact  that it is creating a map doesn't stand out; the `xsl:sequence` is a distraction because there's no sequence involved; and many users dislike long multi-line XPath expressions because of formatting problems in their editing tools)

I propose a new instruction xsl:record which allows:

```
            <xsl:record author="string(AUTHOR)"
                                title="string(TITLE)" 
                                price="xs:decimal(PRICE)" 
                                publisher="string(../@name)"/>
```
This is rather like literal result elements in that the attributes are user-defined rather than system-defined. Unlike LREs, the values are general expressions rather than AVTs, because the values are not necessarily strings. The instruction can only be used where the keys (field names) take the form of NCNames.

If variable entries are required, or entries whose keys are not NCNames, they can appear as child instructions:

```
            <xsl:record author="string(AUTHOR)"
                                title="string(TITLE)" 
                                price="xs:decimal(PRICE)" 
                                publisher="string(../@name)">
              <xsl:if test="@private">
                  <xsl:map-entry name="'private entry'" select="true()"/>
              </xsl:if>
          </xsl:record>
```

Follow the tradition of LREs, duplicates are resolved as "last one wins".

If "standard attributes" such as `[xsl:]version` are required, they must be in the XSLT namespace, as with LREs.

ndw:
I like it. I don't personally find `xsl:map`/`xsl:map-entry` especially difficult to read or use, but I can see the appeal of something less verbose. Allowing `xsl:map-entry` as a child is a nice workaround for conditional values and values that don't have keys that are NCNames.

[...]

PR #1858, opened 2025-03-04
Initial xsl:record

johnlumley:
An initial draft of the xsl:record instruction, for discussion


johnlumley:
The proposed signature uses `@xsl:on-duplicates` but duplicate handling within the `xsl:map` should now use `@duplicates` so the signature needs to be altered.

michaelhkay:
I suggest putting xsl:record under a separate section heading from xsl:map, it makes it easier to find the rules applying to each instruction.

There's a need to change the rules for standard attributes such as [xsl:]version -- we can no longer simply say that if the element is in the XSLT namespace, the attribute isn't, and vice versa.

Otherwise looks good.

michaelhkay:
Needs extensions to the schema for XSLT 4.0.

ndw:
I can see the motivation for `xsl:as` and `xsl:duplicates`, but that's so different from the other instructions that I fear it will cause confusion. I think users are at least as likely to write `<xsl:record as="..."` to define the record type as they are to write `<xsl:record as="..."` meaning they want to create an entry with a key of `as`.

I think I'd prefer it if we just reserved the names `as` and `duplicates` for ourselves and said you have to use a sequence constructor to make those entries, if you need them.

michaelhkay:
>so different from the other instructions

Yes, I share this concern. The design is very logical, but also rather unfamiliar. However, I'm not comfortable with reserving attribute names either.

Perhaps an alternative is to put the attributes on a separate element, xsl:record-options, appearing as a child of xsl:record?

johnlumley:
> Yes, I share this concern. The design is very logical, but also rather unfamiliar. However, I'm not comfortable with reserving attribute names either.

I agree - we have a bit of a quandary, with no obvious single-element + options attributes + entry attributes solution possible  .

>Perhaps an alternative is to put the attributes on a separate element, xsl:record-options, appearing as a child of xsl:record?

This might be a possibility if it _only_ applies to the `as|duplicates` attributes, which is what I think you're suggesting, but if _all_ entry-generating attributes have to go on an `xsl:record-options` child (which I suspect you're not advocating) it sort of minimises the conciseness goal. And now we have a special instruction where the options declaration is in a non-intuitive and anomalous position.

Interestingly, in the absence of a contained sequence constructor (which probably will be the vast majority of use cases) we don't need the `@duplicates` option as there can be no key duplication, so it's only needed where there _is_ a sequence constructor, which already reduces the conciseness.

ndw:
#HEADDESK, of course. Apologies for the noise, I was thinking of a map constructor.

michaelhkay:
Fix #322

PR #1888, opened 2025-03-20
366 xsl:package-location

Arithmeticus:
First draft, for initial feedback.

Notes:
- Because the CG has little energy/resources to develop the EXPath Zip module, I have situated the question of archive (compressed or not) in the URI scheme itself. There are dozens of archives, dozens of URI schemes. The only case where I have found overlap is in the `jar:` scheme/archive. Yes, I've seen `zip:` used as an alias for `jar:`, but it's not an official IANA URI scheme. This may need discussion.
- I have opted to bind `@priority` to a non-zero integer. This is the first time the constraint for the union of positive and negative integers has been placed on an XSLT attribute, so I may not have correctly set up `element-catalog.xml`.
- I have opted to not make attribute values `format`, `name`, and `version` as criteria for the priority package location (new term), so that developers can be warned when the package is at odds with the declaration. To make them criteria would mean that inconsistencies between the declaration and the referenced packages would remain undetected.
- I adopted the terms "URL" and "entry" based upon the IANA nomenclature for the jar: scheme.
- I may have overthought the distinction between archive and non-archive URIs. Feedback is appreciated.
- Error code `3000` has been broken up into different possible errors.
- Suggestions on the type and number of tests that need to be written for the test suite are welcome.

### DIFF
--- a/specifications/xpath-functions-40/src/xpath-functions.xml
+++ b/specifications/xpath-functions-40/src/xpath-functions.xml
@@ -14188,7 +14188,7 @@ ISBN 0 521 77752 6.</bibl>
               existing key, in the case where the existing key and the new key differ (for example,
               if they have different type annotations), it is no longer guaranteed that the new
               entry includes the new key rather than the existing key.</p>
-
+           </item> 
            <item>
               <p>In regular expressions, the assertions <code>^</code> and <code>$</code> can no longer be
               followed by a quantifier. This is because (a) a quantifier that allows zero occurrences means
--- a/specifications/xslt-40/src/element-catalog.xml
+++ b/specifications/xslt-40/src/element-catalog.xml
@@ -161,7 +161,8 @@
             <e:data-type name="uri"/>
         </e:attribute>
         <e:attribute name="priority" required="yes">
-            <e:data-type name="integer"/>
+            <e:data-type name="positiveInteger"/>
+            <e:data-type name="negativeInteger"/>
         </e:attribute>
         <e:attribute name="format" required="no">
             <e:data-type name="string"/>
--- a/specifications/xslt-40/src/element-catalog.xml
+++ b/specifications/xslt-40/src/element-catalog.xml
@@ -77,6 +77,7 @@
          <e:data-type name="string"/>
       </e:attribute>
       <e:choice repeat="zero-or-more">
+         <e:element name="package-location"/>
          <e:element name="accept"/>
          <e:element name="override"/>
       </e:choice>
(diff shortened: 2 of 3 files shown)


michaelhkay:
3.5.3

> If two or more [xsl:package-location]declarations have the same priority value, a static error is raised.

* Specify **sibling** xsl:package-location declarations. Ditto in ERR XTSE3002. Technically, this isn't a declaration as defined in 3.7, but I think that's OK so long as we don't hyperlink it). 

> For a package location to be accepted as a priority package location

Additional condition: the value of @format must either be "xslt", or absent, or a value recognised by the implementation. i.e declarations with an unrecognized value of @format are ignored.

> file

The word "file" is widely used throughout the new text. I think more common W3C practice is to use "resource". And a URI doesn't "point to" a resource, it is dereferenced to deliver a resource. I think in the general case you should identify that the resource is an archive by means of the media type in the HTTP response, not by the URI scheme used to fetch it. In any case, if the package is in a local JAR or ZIP file, it's going to use the "file" URI scheme, and you're going to use either the file extension or peeking at the initial octets to identify it as an archive.

We need a note somewhere that the URI in the href attribute, after resolving as described in 5.9, may be dereferenced via an indirection mechanism such as a user-supplied URI resolver or catalog.

I'm wondering also whether we should make explicit provision for resolving using the combination of a repository URI (e.g "maven central") and an artefact ID within that repository.

michaelhkay:
>Suggestions on the type and number of tests that need to be written for the test suite are welcome.

I think the main thing is to exercise all the syntactic options. Tests in this kind of area tend to be difficult for a number of reasons: (a) some of the detail tends to be implementation-defined, (b) it's easy to end up testing the test driver rather than testing the product, (c) we don't have a reliable way of accessing HTTP resources from tests. Just do the best you can!

michaelhkay:
Given that JAR file URIs are not really standardized, I wonder if they too should be handled by a pair of attributes, one identifying the container, and the other the specific artefact within the container.

Arithmeticus:
> We need a note somewhere that the URI in the href attribute, after resolving as described in 5.9, may be dereferenced via an indirection mechanism such as a user-supplied URI resolver or catalog.

Isn't that editorial task for 5.9, not here?

> I'm wondering also whether we should make explicit provision for resolving using the combination of a repository URI (e.g "maven central") and an artefact ID within that repository. Perhaps just add an @artefact attribute...

Accommodating repositories is a good idea. But are they in parity with each other (compare Maven to NuGet, for which "art[ie]fact" is not a term of art)? Would it not be better if these were part of the URL itself as parameters for an API call, and so incorporated into `@href`? Or are we envisioning pointing to a base url and providing the parameters as an option map?

Arithmeticus:
Recently updated files take into account most of the feedback, but leaves open the question about how to negotiate with a software package management system. I fear that such systems will have asymmetric methods for accessing packages, and there may not be a clean solution that is universally satisfactory. (See long list of [SPMS](https://en.wikipedia.org/wiki/List_of_software_package_management_systems)s here.)

ndw:
At [meeting 125](https://qt4cg.org/meeting/minutes/2025/06-17.html), the CG agreed to merge this PR.

PR #1890, opened 2025-03-23
More examples added to fn:chain

dnovatchev:
Added 6 more examples and tests
All are correctly executed.


ndw:
This PR would be easier to review, I think, if you rebased it off of `master`. I'm seeing lots of spurious diffs.

michaelhkay:
There are two differences between fn:compose (as proposed) and fn:chain.

1. fn:compose delivers a function, fn:chain applies the function. I would argue that the first option is better, because in most practical use cases I have seen, you want to apply the function more than once, to different inputs. Most often, you want to use it in a context where a function is required, for example as a callback to fn:filter.
2. fn:chain is able to include functions in the chain that have an arity other than one. However, if this is the case, then the immediately preceding function has to deliver the multiple arguments in the correct format, which means there is a strong dependency between the two functions: to build your chain of functions, you have to know a lot about the functions you are stringing together. I think that explains why, in all the examples here that use functions with arity >1, the function you are calling is actually statically known. I'm sure it's possible to construct examples where that's not the case, for example where the second function might be any binary operator on two numerics. But in those cases, I think you can achieve the required effect using fn:apply:

For example, the example `chain(2, ($range, $double-all, op("*")))` can be expressed as 2 => compose($range, $double-all, array:append([], ?), apply(op("*"), ?))`. That's a little bit longer, but I would argue that it's actually clearer because it makes the steps explicit: in particular there is no magic conversion between sequences and arrays.

It's this "magic conversion" that I dislike most about the fn:chain specification. Specifically, the rule:

```
if (function-arity($f) eq 1) then [ $x ]
    else if ($x instance of array(*)) then $x 
    else array { $x }
```
breaks the principle of substitutability. If a function can handle any sequence of items as input, then it shouldn't behave differently if that sequence of items happens to be a singleton array item.

dnovatchev:
> This PR would be easier to review, I think, if you rebased it off of `master`. I'm seeing lots of spurious diffs.

Norm, what you are actually seeing is a whole new chapter appearing in the spec (and no, I didn't write it...) that is now hidden in the official FO 4 spec. due to someone's error.

The most likely suspect is probably PR 1856, which was accepted on our last week's meeting and then merged.

At present the official spec is missing completely chapter **_6. Regular expressions_** . Please, have a look - screenshot taken on March 24th at 8:03AM PDT:

image

and here is what was hidden but should have been there - now seen as "part" of the current PR:

image


During my work on the current PR gradle signaled an error - unclosed `<item>` tag  in content that is not touched by this PR.
I had to add this missing closing tag in order to make the build work, and this is how the whole chapter on  Regular Expressions magically appeared in its right place.

And I expected you to say: Thank you ...  😄 

Thus all these "spurious diffs" are the renumerations of all following sections, and they should have been caused by PR1856 if it didn't omit the end-tag.

To summarize: nothing is wrong in correcting the content to what it was supposed to be.

Thanks,
Dimitre

ndw:
I think we're struggling with some issues related to a broken PR getting merged last week.

But yes, @dnovatchev, please remove the PDF.

dnovatchev:
The new PR is: https://github.com/qt4cg/qtspecs/pull/1894

PR #1894, opened 2025-03-24
Additional examples to fn:chain - in a new branch

dnovatchev:
Re-submitted the same as PR 1890. 
Added some new examples to fn:chain.


michaelhkay:
I wrote in #1890:

For example, the example `chain(2, ($range, $double-all, op("*")))` can be expressed as `2 => compose($range, $double-all, array:append([], ?), apply(op("*"), ?))`. 

Correction:

The example `chain(2, ($range, $double-all, op("*")))` can be expressed as `2 => compose($range, $double-all, array:build#1, apply(op("*"), ?))`.

In fact, all the examples that rely on the feature of fn:chain whereby a function can be applied to a supplied sequence of N items could be handled with the aid of a helper function in the pipeline:

```
let $apply-to-sequence := fn($f as function(*), $args as item()*) {apply($f, array:build($args))}
2 => compose($range, $double-all, $apply-to-seq(op('*'))
```

PR #1940, opened 2025-04-21
1939 XQDY0153 (from try/finally) should be a type error

### LINKED ISSUE #1939: XQDY0153 (from try/finally) should be a type error
The `finally` clause is required to return an empty sequence. If not, it raises XQDY0153. This should be a type error rather than a dynamic error, so that it can be raised statically when appropriate.

Also noted in passing: "the the" in item 4 of the "changes" section of §4.20).

ChristianGruen:
Closes #1939


ndw:
The QT4CG agreed to merge this PR without discussion at [meeting 118](https://qt4cg.org/meeting/minutes/2025/04-22.html).

PR #1942, opened 2025-04-21
37 Support sequence, array, and map destructuring declarations

### LINKED ISSUE #37: Support sequence, array, and map destructuring declarations
Given a function that returns a sequence, array, or map of a fixed length or structure, it would be useful to extract those values in a destructuring declaration like can be done in other languages (such as JavaScript, Kotlin, C++, and Python). For example:

    let $(sin, cos) := sincos(math:pi()) (: sequence :)
    let $[x, y, z] := camera-angle() (: array :)
    let ${r, i} := complex(1, 2) (: map :)

These would be equivalent to:

    let $ret := sincos(math:pi()), $sin := $ret[1], $cos := $ret[2] (: sequence :)
    let $ret := camera-angle(), $x := $ret?(1), $y := $ret?(2), $z := $ret?(3) (: array :)
    let $ret := complex(1 ,2), $r := $ret?r, $i := $ret?i (: map :)

It should be possible to define the type of a component and/or the whole construct:

    let $(sin as xs:float, cos) as xs:float* := sincos(math:pi()) (: sequence :)

For maps, it would also be useful to rename the components, such as:

    let ${re := r, im as xs:double := i} := complex(1, 2) (: map :)

It should also be possible to capture any left-over items in the sequence/array/map, for example:

    let $(headings, rows) := load-csv("test.csv")

A destructuring declaration should be usable anywhere a variable binding can be defined.

It should not be an error to use the same variable name twice. This supports conventions such as using `_` for unused values. For example:

    let $[_, y, _] := camera-angle()

rhdunn:
Closes #37.

This currently only supports XPath. I'm working on the wording for XQuery.

### DIFF
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -1009,14 +1009,14 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
 
   <g:production name="LetBinding">
      <g:choice>
-        <g:ref name="LetItemBinding"/>
+        <g:ref name="LetValueBinding"/>
         <g:ref name="LetSequenceBinding" if="xpath40"/>
         <g:ref name="LetArrayBinding" if="xpath40"/>
         <g:ref name="LetMapBinding" if="xpath40"/>
      </g:choice>
   </g:production>
 
-  <g:production name="LetItemBinding">
+  <g:production name="LetValueBinding">
     <g:choice>
       <g:sequence>
         <g:ref name="VarNameAndType"/>
@@ -1028,7 +1028,8 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
   </g:production>
 
   <g:production name="LetSequenceBinding" if="xpath40">
-     <g:string>$(</g:string>
+     <g:string>$</g:string>
+     <g:string>(</g:string>
      <g:oneOrMore separator=",">
         <g:ref name="VarNameAndType"/>
      </g:oneOrMore>
@@ -1036,10 +1037,13 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
      <g:optional>
         <g:ref name="TypeDeclaration"/>
      </g:optional>
+     <g:string>:=</g:string>
+     <g:ref name="ExprSingle"/>
   </g:production>
 
    <g:production name="LetArrayBinding" if="xpath40">
-      <g:string>$[</g:string>
+      <g:string>$</g:string>
+      <g:string>[</g:string>
       <g:oneOrMore separator=",">
          <g:ref name="VarNameAndType"/>
       </g:oneOrMore>
@@ -1047,10 +1051,13 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
       <g:optional>
          <g:ref name="TypeDeclaration"/>
       </g:optional>
+      <g:string>:=</g:string>
+      <g:ref name="ExprSingle"/>
    </g:production>
 
    <g:production name="LetMapBinding" if="xpath40">
-      <g:string>${</g:string>
+      <g:string>$</g:string>
+      <g:string>{</g:string>
       <g:oneOrMore separator=",">
          <g:ref name="VarNameAndType"/>
       </g:oneOrMore>
@@ -1058,6 +1065,8 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
       <g:optional>
          <g:ref name="TypeDeclaration"/>
       </g:optional>
+      <g:string>:=</g:string>
+      <g:ref name="ExprSingle"/>
    </g:production>
 
   <g:production name="WindowClause" if="xquery40">
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -1008,6 +1008,15 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
   </g:production>
 
   <g:production name="LetBinding">
+     <g:choice>
+        <g:ref name="LetItemBinding"/>
+        <g:ref name="LetSequenceBinding" if="xpath40"/>
+        <g:ref name="LetArrayBinding" if="xpath40"/>
+        <g:ref name="LetMapBinding" if="xpath40"/>
+     </g:choice>
+  </g:production>
+
+  <g:production name="LetItemBinding">
     <g:choice>
       <g:sequence>
         <g:ref name="VarNameAndType"/>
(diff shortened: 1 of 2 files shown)


ChristianGruen (commented):
Thanks a lot, Reece. Minor observations:

1. I guess the expanded version of `let $b := ...` lacks the type: `let $b as xs:integer := ...`.
2. `let $local:x` may be confusing for maps, for example when different variables names are used: `let ${$a, $xml:a} := { 'a': 1 }`. In addition, it may restrict ourselves if we decide to also support QName bindings in the future (`let $xml:a := { xs:QName('xml:a'): 1 }`).
3. For the sake of symmetry, `map:get($temp, 'a')` could be used in the expanded map variant. Alternatively, the lookup operator could also be used for arrays (the 4.0 variant of which returns an empty sequence instead of an error when the position is out of bounds).

ChristianGruen:
One tricky thing about the presentation is the (example-only) `$temp` variable, which might shadow another variable of the same name in a more complex FLWOR expression in practice.

A pragmatic solution in an implementation would be to bind the input to the last variable name…

```xquery
let $($first, $last) := (4, 5)
return $first + $last

(: expanded :)
let $last := (4, 5)
let $first := item-at($last, 1)
let $last := item-at($last, 2)
return $first + $last
```

…unless the same variable name is not allowed multiple times in the destructruring let clause.

michaelhkay:
0. Need to add to the Changes list for Let Expressions
1. The term `LetItemBinding` for the traditional `let $x := (1,2,3)` seems inappropriate because it can bind any value, not just a single item. Perhaps `LetValueBinding`?
2. In the grammar, it seems the `:= ExprSingle` is allowed to appear only for a (traditional) binding, not for the new destructured bindings.
3. In`LetSequenceBinding`, `LetArrayBinding` etc allowing a type declaration outside the brackets as well as inside seems rather confusing: `let($x as xs:decimal, $y as s:decimal) as xs:integer* := EXPR`. I suggest disallowing the type declaration outside the parens.
4. Sadly, it's `fn:items-at` not `fn:item-at`
5. In the expansion of `let ${ $a, $b as xs:integer, $local:c } := (2, 4, 6)` the `as xs:integer` appears to have been lost. This happens also in the array and map examples.
6. Apparrent copy/paste error in `is the value of the destructuring array for at the given destructured index in the array.`
7. The phrase `In a map or record destructuring let expression` suggests that these are different things, which I don't think is the case. It would be safer to use the production name, "In a LetMapBinding". (Also, a Let Expression can include more than one kind of binding, e.g. let $[$a] in $A, ${$b} in $B`.)
8. In rule 6, the sentence "The result of the let expression is obtained by evaluating the return expression with a dynamic context in which the range variables are bound to the binding sequence." needs adjusting: the term "binding sequence" no longer applies.
9. Nothing to do with this proposal, but the existing text "The scope of a variable bound in a let expression is the return expression. The scope does not include the expression to which the variable is bound. " is clearly wrong; as the example demonstrates, the scope also includes the initialization expressions of subsequent variables bound in the let expression.
10. I think there is a need for some non-normative explanation (notes and/or examples) of how the feature works and how it might be used, More needs to be said about cases where the number of variables bound is smaller or greater than the length of the sequence/array, and about how type declarations work. (For example if there is no value to be bound, then the declared type must allow an empty sequence). (Also discuss maps with keys that are not NCNames.)
11. "The destructured key is the NCName of the destructured variable." I think this is imprecise shorthand for "The destructured key is the local part of the name of the destructured variable (the namespace part is ignored)."
12. The sentence "The binding sequence of the range variable is the value of the destructuring map for the given destructured key." doesn't work. Perhaps "the value associated with the destructured key within the destructuring map".
13. Perhaps the text would be more digestible if you made more use of variables rather than defined terms.

michaelhkay:
Editorial note: when we include meta-variables in a construct such as `let $( $A1, $A2, $A3, ..., $An ) := EXPR` I have been adopting the practice of putting A1, A2, EXPR etc within `<var>` tags. This renders them in italic within the code section, indicating that they are place-markers within the code; and a var such as A1 automatically acquired a subscript (for An you can get the subscript by writing `<var>A/n</var>`).

michaelhkay:
I would suggest making `$(` two tokens rather than one (and similarly for `$[` and `${`). The reason is that it's currently legal to write

`let $(:comment:) name := 17`

and if `$(` is a single token then this example would no longer parse correctly (under the longest token rule).

rhdunn:
@ChristianGruen and @michaelhkay Thanks for the reviews!

Christian: Re: point 2, my understanding w.r.t. `local:c` and `c` from the discussion on the issue was that we wanted to use the localname. I can change this, but we'll need to define semantics for QName and URIQualifiedName EQNames.

Michael:
- Re 1: I was mirroring `ForItemBinding` that is used for the generic (XPath 3.1) `ForBinding`. I've changed it to `LetValueBinding`.
- Re 3: This can be useful e.g. to specify that the let expression is a given record or sequence/array restriction.
- Re 8: This does still apply -- part (c) of each of the new constructs defines how the range variables and bounding sequences are constructed. I've made an attempt to restructure and clarify the logic, so hopefully that is easier to read/follow.
- Re 9: I've updated the wording to try to make this clearer.

Some additional observations:
- `<code><var>T</var></code>` adds an additional space at the end of the code block;
- The EBNF for `ForClause` and `LetClause` look wrong -- they only permit a `,` as a delimiter -- I thought, and from the wording, that `let`/`for` keywords were also permitted now.

michaelhkay:
>The EBNF for ForClause and LetClause look wrong -- they only permit a , as a delimiter -- I thought, and from the wording, that let/for keywords were also permitted now.

No, that's handled at the next level up: a `FLWORExpr` can include more than one `ForClause`

>`<code><var>T</var></code>` adds an additional space at the end of the code block;

It shouldn't.

rhdunn:
> > The EBNF for ForClause and LetClause look wrong -- they only permit a , as a delimiter -- I thought, and from the wording, that let/for keywords were also permitted now.
> 
> No, that's handled at the next level up: a `FLWORExpr` can include more than one `ForClause`

That only applies to XQuery. So the issue is that the ForExpr and LetExpr in XPath are missing a `+` after the ForClause/LetClause.

michaelhkay:
>So the issue is that the ForExpr and LetExpr in XPath are missing a + after the ForClause/LetClause.

It's handled through recursion

```
ForExpr  ::=  ForClause ForLetReturn
ForLetReturn  ::=  ForExpr |  LetExpr | ("return"  ExprSingle)
```

ChristianGruen:
> Christian: Re: point 2, my understanding w.r.t. `local:c` and `c` from the discussion on the issue was that we wanted to use the localname. I can change this, but we'll need to define semantics for QName and URIQualifiedName EQNames.

True, my comments were a bit brief. My initial thought was that it is easier to disallow bindings for prefixed/namespaced variable names (which would give us the option to define bindings for QName keys later on). I have no strong opinion on that: Maybe it is simpler indeed to only consider the local name.

rhdunn:
@michaelhkay point 9 is actually correct, so I'll revert/restore that wording:

```
let $x := doc('a.xml')/*, $y := $x//*
return $y[@value gt $x/@min]
```
expands via the multi-variable rule to:
```
let $x := doc('a.xml')/* let $y := $x//*
return $y[@value gt $x/@min]
```
which, as you noted above, is evaluated recursively to:
```
LetExpr
  LetClause (1) :: let $x := doc('a.xml')/*
  ForLetReturn (2)
    LetExpr
       LetClause (3) :: let $y := $x//*
       ForLetReturn (4) :: return $y[@value gt $x/@min]
```

Here, `$x` is not available in (1) but is available in (2). Likewise, `$y` is not available in (3) but is available in (4).

ndw:
Resolved by #2055

PR #1990, opened 2025-05-08
Update schema-for-xslt40.xsd

walmsleyph:
Fixed invalid syntax xs:simpleType/@ref (moved to @memberTypes) in simpleType named method

Based the type fixed-namespaces-type-default on xs:token instead of xs:string to allow for whitespace normalization

Changed collation attribute on xsl:merge-key to be an avt (according to spec)

Changed attributes that were previously of type "xsl:char-optionally-expanded" to just xs:string since the spec says they can be any string.  I couldn't think of a reason they should be limited to one character optionally followed by a colon and more characters, so I assumed this was some sort of artifact from the past.

Changed xsl:next-iteration and xsl:evaluate to not allow mixed content

Corrected _split_when to _split-when

Added missing shadow attributes (in two places) for allow-duplicate-names, build-tree, json-lines and json-node-output-method

Added missing shadow attribute for select on perform-sort

Added assertions for required attributes:
- xsl:use-package - name
- xsl:expose - names, component and visibility

Made the errors attribute a list of tokens instead of just xs:token.  Doesn't affect validation but I think it is more clear.

Changed default value of per-mille from a tilde to ‰

Changed the default value of the on-no-match attribute from shallow-skip to text-only-copy, per the spec

Gave xsl:exclude-result-prefixes the same type as no-namespace exclude-result-prefixes, to allow for #all and #default

Gave xsl:extension-element-prefixes the same type as no-namespace extension-element-prefixes, to allow for #default

Removed the xsl:prefixes and xsl:char-optionally-expanded types since they were no longer used after the above changes

Changed the type of the visibility attribute on xsl:attribute-set, xsl:function, xsl:template and xsl:variable to xsl:visibility-not-hidden-type to exclude "hidden" per the spec

Changed the keyword value of the fixed-namespaces attribute from #default to #standard (and adjusted type names)

### DIFF
--- a/specifications/xslt-40/src/schema-for-xslt40.xsd
+++ b/specifications/xslt-40/src/schema-for-xslt40.xsd
@@ -722,14 +722,14 @@ of problems processing the schema using various tools
       <xs:complexContent>
         <xs:extension base="xsl:element-only-versioned-element-type">
           <xs:attribute name="name" type="xsl:EQName"/>
-          <xs:attribute name="decimal-separator" type="xs:string" default="."/>
-          <xs:attribute name="grouping-separator" type="xs:string" default=","/>
+          <xs:attribute name="decimal-separator" type="xsl:char-optionally-expanded" default="."/>
+          <xs:attribute name="grouping-separator" type="xsl:char-optionally-expanded" default=","/>
           <xs:attribute name="infinity" type="xs:string" default="Infinity"/>
           <xs:attribute name="minus-sign" type="xs:string" default="-"/>
-          <xs:attribute name="exponent-separator" type="xs:string" default="e"/>
+          <xs:attribute name="exponent-separator" type="xsl:char-optionally-expanded" default="e"/>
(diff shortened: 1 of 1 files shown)


michaelhkay:
Thanks for this - a lot of detail here! I have reviewed the changes and have only one comment.

>Changed attributes that were previously of type "xsl:char-optionally-expanded" to just xs:string

The spec prose says: In each case the value must either be a single character, or a string in the form m:r where m is a single character (the marker) used to represent the property in the picture string, and r (the rendition) is an arbitrary string used to represent the property in the formatted result.

walmsleyph:
Oops, I looked for an explanation of that and I don't know how I missed it.  I will revert that change.

Arithmeticus:
@ndw Note that the encoding for the production of the specs in HTML is wonky here, at ‰:

<xs:attribute name="per-mille" type="xsl:char-optionally-expanded" default="â€°"/>

PR #2013, opened 2025-05-20
748 Parse functions: consistency

### LINKED ISSUE #748: Parse functions: consistency
**Updated (31-01-2025):** Aligned with latest XQFO spec.

The functions for parsing input have been defined by different people, and the current state is quite inconsistent:

Function | Parameters
--- | ---
`fn:parse-xml` | `$value as xs:string?, $options as map(*)?`
`fn:doc` | `$source as xs:string?`
`fn:parse-json` | `$value as xs:string?, $options as map(*)?`
`fn:json-doc` | `$source as xs:string?, $options as map(*)?`
`fn:parse-html` | `$html as (xs:string \| xs:hexBinary \| xs:base64Binary)?, $options as map(*)?`
`fn:parse-csv` | `$csv as xs:string?, $options as map(*)?`

I believe there’s some need to unify the functions. We should:

* add `fn:csv-doc($source, $options)` and `fn:html-doc($source, $options)`,
* rename the argument of all `fn:parse-...` functions to `$value`,
* enhance all `fn:parse-...` functions to also accept binaries, and
* always return an empty sequence if the source/input argument is empty.

See #1021 for `fn:doc`.

ChristianGruen:
Closes #748

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -19804,7 +19804,7 @@ return { "width": $int16-at($loc + 5),
    <fos:function name="parse-xml" prefix="fn">
       <fos:signatures>
          <fos:proto name="parse-xml" return-type="document-node(*)?">
-            <fos:arg name="value" type="xs:string?" example="'&lt;a/&gt;'"/>
+            <fos:arg name="value" type="(xs:string | xs:hexBinary | xs:base64Binary)?" example="'&lt;a/&gt;'"/>
             <fos:arg name="options" type="map(*)?" default="{}"/>
          </fos:proto>
       </fos:signatures>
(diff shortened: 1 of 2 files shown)


michaelhkay:
When the input to parse-xml or parse-xml-fragment is supplied as a binary value, I feel the rules for inferring an encoding should be the rules of XML specification, not the rules of the unparsed-text function. (I don't know if they are materially different.) 

Similarly for parse-html().

Using the unparsed-text() rules seems fine for parse-json() and parse-csv().

Arithmeticus:
I don't understand the motivation for allowing `parse-xml()` to allow binary input, aside from the attempt to provide symmetry with other parse functions. 

To me, the allowance doesn't make sense. If you're fetching an XML file, `doc()` or `document()` should suffice. If you have to resort to `unparsed-text()` or heavens the new `unparsed-binary()` something's terribly wrong, and you shouldn't try to parse it before you fix the problems. One of those problems might be encoding, resolution of which shouldn't be in `parse-xml()`'s to-do list. What am I missing?

michaelhkay:
>I don't understand the motivation

I'm inclined to agree. I can't quite see where you would get XML as a byte stream other than from fetching a URI. I can't prove non-existence of a use case, but it would be nice to see a real one. Same goes for parse-html, incidentally.

ndw:
At [meeting 126](htps://qt4cg.org/meeting/minutes/2025/06-24.html), the CG agreed to merge this PR.

PR #2077, opened 2025-07-02
2016 File Module: Incorporate changes

### LINKED ISSUE #2016: File Module: Incorporate changes
The EXPath File Module must be revised in several steps. First of all, several functions need to be incorporated that were added to the initial version (details: https://docs.basex.org/12/File_Functions).

ChristianGruen:
Includes general refactorings.

Closes #2016

### DIFF
--- a/specifications/EXPath/file/src/function-catalog.xml
+++ b/specifications/EXPath/file/src/function-catalog.xml
@@ -24,10 +24,22 @@
       <p>Tests if a file or directory exists.</p>
     </fos:summary>
     <fos:rules>
-      <p>Returns <code>true</code> if <code>$path</code> points to a file or directory,
+      <p>Returns <code>true</code> if <code>$path</code> points to a file or a directory,
         <code>false</code> otherwise.</p>
       <p>On a UNIX-based system, the root and the volume roots are considered directories.</p>
     </fos:rules>
+    <fos:examples>
+      <fos:example>
+        <fos:test>
+          <fos:expression>file:exists('.')</fos:expression>
+          <fos:result>true()</fos:result>
+        </fos:test>
+        <fos:test>
+          <fos:expression>file:exists('/')</fos:expression>
+          <fos:result>true()</fos:result>
+        </fos:test>
+      </fos:example>
+    </fos:examples>
   </fos:function>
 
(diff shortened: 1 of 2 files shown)


ChristianGruen:
> I suspect there are operating systems where a path can point to something other than a file, a directory, or a symbolic link, but perhaps such systems aren't important.

Yes, I ignored those cases; they depend too much on the underlying operating/file system. Of course this is a general challenge with this function set: Properties like file permissions are difficult to unify. However, I will be happy to try to embed some more functionality that may be useful.

> There are some changes in functionality that should perhaps be highlighted because they affect implementors, for example I notice file:read-binary can now supply a length without supplying an offset.

I can add the changes. The main reason for skipping them was that there was no previous W3 version of the spec.

> Why have you dropped the introductory paragraph of section 5? I found it useful. Is it not true?

It did not apply to e.g. `file:path-to-native`, and basically all functions of that section that operate on the file system.

> file:descendants is unusual in containing an example (hooray!), but the example uses file:children.

Oh ;) Yes, I should add more examples. I felt that the majority of the functions is self-explanatory, but that may also be laziness.

Thanks for the remaining observations, I will integrate them soon.

specifications/EXPath/file/src/function-catalog.xml
```diff
@@ -5,1133 +5,1845 @@
...
+        <fos:arg type="xs:string" name="path"/>
+      </fos:proto>
+    </fos:signatures>
+    <fos:properties>
+      <fos:property>nondeterministic</fos:property>
+      <fos:property>context-dependent</fos:property>
+      <fos:property>focus-independent</fos:property>
+    </fos:properties>
+    <fos:summary>
+      <p>Returns the parent directory of a given path.</p>
```
Arithmeticus, line 1406: "Returns the parent" => "Returns the path to the parent"

ChristianGruen, line 1406:
```suggestion
      <p>Returns the path to the parent directory of a given path.</p>
```

ChristianGruen:
I have finally added a whole bunch of new File Module tests: a65e6e454a0e3aef956f5807e40a0ca23f9c183f.

PR #2085, opened 2025-07-08
Fix markup errors in the EXPath file: specification

ndw:
The FOS schema does not allow an fos:example containing an fos:test to omit the fos:result element. I’ve inserted

```
<fos:result>���</fos:result>
```

as a placeholder to mark the problem. I also had to move some of the fos:errors sections to a different location.

@ChristianGruen apologies for just pushing these in. I wanted to get the specs building again.


ChristianGruen:
Thanks.

Issue #2051, opened 2025-06-14
XSLT group by cluster


Arithmeticus:
I propose an enhancement of `xsl:for-each-group` to support clustering.

To start off with a simple use case, suppose one has the following population, `<xsl:variable name="ages" as="xs:integer*" select="5, 24, 9, 5, 6, 8, 36, 38, 28"/>` and one wishes to cluster the figures like so, in four groups: `(5, 5, 6, 8, 9)`, `(24)`, `(28)`, `(36, 38)`.

One is tempted to create an `<xsl:for-each-group>` with `group-by="((. - 1) to (. + 1))"`. But this does not work. If `@composite` is absent or is `no`, eighteen groups are created. If `@composite` is `yes`, eight groups are created. In both cases, the results are not significantly close to the desired output.

I propose a new `@group-by-cluster`. The following code
```
<xsl:for-each-group select="$ages" group-by-cluster="((. - 1) to (. + 1))">
    <xsl:sort select="current-grouping-key()"/>
    <group key="{current-grouping-key()}" count="{count(current-group())}">
        <xsl:copy-of select="current-group()"/>
    </group>
</xsl:for-each-group> 
```

would produce this
```
   <group key="4 5 6 7 8 9 10" count="2">5 5 6 8 9</group>
   <group key="23 24 25" count="1">24</group>
   <group key="27 28 29" count="1">28</group>
   <group key="35 36 37 38 39" count="1">36 38</group>
```

There are numerous use cases for the proposed new feature. Here are a few:
* Clustering map or spatial coordinates
* Grouping disparate rectangles from OCR output
* Reconciling triples in linked open data (RDF) that use different IRIs synonymously
* Detecting typologies within in large corpora of documents that have periodically repetitive formulaic paragraphs.
* Discovering networks of connected things, e.g., networks of email correspondence or publication citations
 
Currently, the clustering I describe above is feasible in XSLT, but it requires creative strategies, usually a combination of preprocessing and the creation of specialized helper functions to recursively iterate over multiple grouping keys to create group numbers. These are challenging to write and debug, and one loses identity in a preprocessed copy of the original. 

By putting clustering into a `@group-by-cluster` construct, users benefit not only from convenience but also from performance, as a processor might bring novel strategies for clustering.

The `current-grouping-key()` for a group would consist of a sequence of all members' grouping keys, duplicates removed. No two groups would have any overlap in their grouping key sequences. (That's the definition of a cluster.)

[...]

PR #2123, opened 2025-07-24
2051: XSLT group by cluster

Arithmeticus:
Companion PR to #2051 .

I have opted for only two examples, hoping they catalyze the imagination of what is possible. Comments welcome.

### DIFF
--- a/specifications/xslt-40/src/xslt.xml
+++ b/specifications/xslt-40/src/xslt.xml
@@ -24697,11 +24697,11 @@ the same group, and the-->
                      groups that makes up the <termref def="dt-processing-order">processing order</termref>. 
                      That pairwise comparison of groups is iterated until it is the case that no pair of groups
                      have any shared items in their grouping keys.</p>
-                  <p>The result of <code>group-by-cluster</code> is the grouping of groups. As a result, an item
-                     in the population may be appended to zero or one groups. Because of the rules for grouping of
+                  <p>The result of <code>group-by-cluster</code> is the grouping of groups. As a result, each item
+                     in the population will be appended exactly one group or none at all. Because of the rules for grouping of
                      groups, an item will never appear in more than one group.</p>
-                  <p>The distinct nature of <code>group-by-cluster</code> relies upon items having multiple grouping keys,
-                     to be able to cluster groups. If <code>composite="yes"</code> is specified, the effect is the same
+                  <p>The distinctive feature of <code>group-by-cluster</code> depends upon items having multiple grouping keys,
+                     otherwise clustering cannot take place. If <code>composite="yes"</code> is specified, the effect is the same
                      as if the attribute <code>group-by</code> had been used.</p>
                </item>
                <item>
@@ -25251,10 +25251,11 @@ the same group, and the-->
             </example>
             <example>
                <head>Grouping Items by Proximity (Clustering)</head>
-               <p>Consider a sequence of atomic values that we wish to group according to proximity, which
-                  for the sake of illustration we define as two units or less.</p>
+               <p>Consider a sequence of atomic values that we wish to group according to proximity. For
+                  the sake of illustration we restrict ourselves to integers. An integer should join a cluster
+                  if it differs from another integer by two or less.</p>
                <eg xml:space="preserve" role="xslt-instruction">&lt;xsl:variable name="ages" as="xs:integer*" select="5, 24, 9, 5, 6, 8, 36, 38, 28"&gt;</eg>
-               <p>We wish to group by cluster, and return the following four groups:</p>
+               <p>We want to return the following four groups:</p>
                <eg xml:space="preserve" role="non-xml">map { 1 : (5, 5, 6, 8, 9), 2 : (24), 3 : (36, 38), 4 : (28)}</eg>
                <p>Solution:</p>
                <eg xml:space="preserve" role="xslt-declaration">&lt;xsl:map&gt;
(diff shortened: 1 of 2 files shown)


michaelhkay:
First:
1. I think I see what you're getting at (though it took some doing...)
2. I don't think I have ever had this requirement - but that doesn't mean no-one else has!

> For each existing group G, if there is some [grouping key] in G that is identical to some [grouping key] in J, then J is appended to that group, and all of J's grouping keys are merged into G's grouping keys, removing duplicates; otherwise a new group is created for items with all of J's grouping key values, and J becomes its first member.

You're using the concept here that a group has a set of grouping keys. That's a new concept, I think, and needs explaining. Until now a group only had one grouping key.

The word "identical" is wrong. Elsewhere we use "equal", and rely on the up-front statement "Atomic [grouping keys]are compared using the rules of the [distinct-values] function, using the relevant collation."

`preserving document order`. Seems wrong. xsl:for-each-group defines the order of groups and the order of items within a group in terms of population order, that is, the order of the input; document order is irrelevant. I think it should probably say `preserving population order`.

The proposal doesn't work well where the desired clustering is on a continuous range, e.g. doubles or timestamps. Would it not be better to have some kind of proximity function? I guess you could pick out specific values in the range, e.g. timestamps that are a multiple of 10 minutes, but it gets awkward. I can imagine that if you try to do this for spatial grouping in 2D space you're going to generate a rather large number of grouping keys for each item.

With a proximity function, I guess we might have `merge-groups-when="fn($key1, $ey2)"`. For your first example the proximity function would be `(abs($key1  - $key2) le 2)`, for the second it would be `$key1 = $key2` (relying on `=` which yields true if there is any item in common). For timestamp clustering it might be `($time1 - $time2) le xs:duration('PT10M')`.

With or without a proximity function, we could move from a procedural specification to a declarative one: within a group, if there is more than one item then every item X is "close to" at least one other item Y, and across groups, no item in group G is "close to" any item not in group G.

Or could this somehow be combined with split-when? Is there one mechanism that would satisfy both requirements?

What does current-grouping-key() return? All the grouping keys for the current group?

I'm wondering if this isn't more like a third value for "composite"? Two items go in the same group if their composite grouping keys are equal under the rules of the `=` operator, as distinct from the current rule ("Composite grouping keys are equal if they contain the same number of items and the items are pairwise equal") - but with the problem that the `=` operator isn't transitive so the multiple passes are needed.

Sorry this has gone off at various tangents.

michaelhkay:
I'm also wondering to what extent this problem is equivalent to doing a partial sort -- for which textbook algorithms exist and we don't have to invent our own?

ruv:
> case where one wishes to group overlapping rectangles

> We decide to group by the bounds of each rectangle: the two X and the two Y values. The first item becomes the first group, with grouping keys 0, 3, 0, 3. 

A simpler approach is to make the key of the item the _position_ of the item in the input sequence. 

Then the comparator/proximity function for this case will look as:
```xquery
fn ( $group_pos as xs:integer+, $item_pos as xs:integer ) as xs:boolean {
  some $ti in $input[ $item_pos ], $gi in $input[ $group_pos ]
  satisfies  my:is_rectangle_overlap( $ti , $gi )
}
```
So, the keys still can be atomic values.

For optimization, you could transform the representation of rectangles from an array of four points (pair of numbers) to an array of four numbers:
```xquery
let $input := $input !  . ? * -> array{ ()
  , . [ position() mod 2 eq 1 ] -> (min(.), max(.))
  , . [ position() mod 2 eq 0 ] -> (min(.), max(.))
}
```
(or to a map of four numbers, if you like)

-----

> That signature makes good sense, because it allows someone to specify different conditions for the cluster:

> * `function ($a, $b) { (count($a) div 2) le count($a[some $j in $b satisfies (abs(. - $j) le 2)]) }` (i.e., _most_ items satisfy the requirement)

Note that this makes it difficult to move from a procedural specification to a declarative one, since the order in which the items of the input population are processed is important (in the general case).

specifications/xslt-40/src/xslt.xml
```diff
@@ -24676,6 +24684,42 @@ the same group, and the-->
...
+                     group. Otherwise, for each existing group <var>G</var>, if there is some
+                        <termref def="dt-grouping-key">grouping key</termref> in <var>G</var> that
+                     is identical to some <termref def="dt-grouping-key">grouping key</termref> in
+                        <var>J</var>, then <var>J</var> is appended to that group, and all of
+                        <var>J</var>'s grouping keys are merged into <var>G</var>'s grouping keys,
+                     removing duplicates. Otherwise, a new group is created for items with all of
+                        <var>J</var>'s grouping key values, and <var>J</var> becomes its first
+                     member. </p>
+                  <p>After every item has been allocated to a group, all groups are
+                     evaluated pairwise. For any groups <var>G1</var> and <var>G2</var> (where group
```
ruv, line 24706: An alternative approach is to check each next item against each group and merge all the groups to which the item belongs. Then there is no need to evaluate groups pairwise after every input item is processed.

Arithmeticus: But sometimes the inclusion of an item in a group makes that group eligible to be included in a previous group that would not have included the item. 

For example, if we take the famous corn + chicken + fox story, where you cannot leave the corn with the chicken or the chicken with the fox, then given the input sequence ("chicken", "corn", "fox") the first two items each gets its own group. When "fox" is evaluated it is not allowed in the first group, but allowed in the second group. But now group two is eligible to be merged with group one. 

It is unknown if a merging of two groups will permit more merges, so groups need to be evaluated through the comparator function until it is the case that no pair of groups satisfy the merge condition. There are ways that evaluation can be optimized, but we leave that to the implementation.

Arithmeticus:
@ruv Yes, thanks! I like this idea of using the position as the grouping key very much. It allows the developer, when writing the comparator function, to focus on the items themselves and not a secondary set of keys.

And `@merge-when` is an excellent alternative choice for the attribute name.

specifications/xslt-40/src/xslt.xml
ruv: > But sometimes the inclusion of an item in a group makes that group eligible to be included in a previous group that would not have included the item.

Yes, but this only possible when a user-defined comparator that accepts the keys of two groups (or the contents of two groups) is used.

My initial comment was for the variant:
   - > if there is some _grouping key_ in `G` that is identical to some _grouping key_ in `J`, then `J`  is appended to that group, and all of `J` grouping keys are merged into `G`'s grouping keys, removing duplicates.

In this variant, the initial key of an item depends only on the item itself, and the resulting grouping key for any group is a sequence of some initial grouping keys. And effectively, the processor compares only keys of individual items.

specifications/xslt-40/src/xslt.xml
```diff
@@ -25023,6 +25033,81 @@ the same group, and the-->
...
+                  <p>The process described above
+                     is repeated until it is the case that there is no pair of groups for which the
+                        <termref def="dt-sequence-comparator">sequence comparator</termref>
+                     evaluates to <code>true</code>, or there is only one group.</p>
+                  <p><code>merge-when</code> recursively groups groups. As a result,
+                     each item in the population will be appended to exactly one group. </p>
+                  <p>When two groups are merged into a single group, the two sets of grouping keys,
+                     which identify the position of each item in the original population, are
+                     indicative of how the items themselves should be merged. </p>
+                  <p>Unlike other grouping methods, a group may have multiple grouping keys, and
```
ruv, line 25066: > a group may have multiple grouping keys,

"Multiple grouping keys" means that the key is an _arbitrary sequence_ of atomic items. Currently, such keys are allowed in any method, but only if the `composite` attribute is _explicitly_ set to "yes".

So if `composite` is not specified or is set to "no", providing multiple grouping keys looks like a conflict. A possible solution is to set the current grouping key to _absent_ if `composite` is not set to "yes".

specifications/xslt-40/src/xslt.xml
```diff
@@ -25023,6 +25033,81 @@ the same group, and the-->
                   </ulist>
                   
                </item>
+               <item>
+                  <p>If the <code>merge-when</code> attribute is present, then its value
+                        <rfc2119>must</rfc2119> be an expression that is a sequence comparator.
+                        <termdef id="dt-sequence-comparator" term=""> A <term>sequence
+                           comparator</term> is a function with the signature <code>fn(item()+,
+                           item()+) as xs:boolean</code>. </termdef>
```
ruv, line 25041: It's an inconsistent if in `split-when` we have to specify an expression that evaluates to `xs:boolean`, but in `merge-when` we have to specify an expression that evaluates to `function(*) as xs:boolean`.  They both should be similar.

It seems, the approach of `split-when` is better; if the user already has a comparator function, they supply a call to this function as  `merge-when=" my:compare-groups( $group, $next-group ) "`.  (concerning these variables see also #2131)

specifications/xslt-40/src/xslt.xml
```diff
@@ -25023,6 +25033,81 @@ the same group, and the-->
...
+                  </p>
+                  <p> For each item in the <termref def="dt-population">population</termref> a group
+                     is created, with that item's position as its grouping key, and the item as its
+                     only member.</p>
+                  <p>After the groups are created, they are evaluated pairwise against the <termref
+                        def="dt-sequence-comparator">sequence comparator</termref>. The pairwise
+                     comparison begins with any two groups. The items in the group that is earlier
```
ruv, line 25048: > The pairwise comparison begins with any two groups.

If the algorithm of choosing groups is not specified, then the resulting set of groups and their order will depend on implementation. Is it intended?  I think, the algorithm should determine both.

A possible way is to specify the algorithm in such a way that the next item/group is considered when the previous items/groups are fully merged only.  And when merging, the next group/item is added to the previous one (in the population order).

specifications/xslt-40/src/xslt.xml
Arithmeticus: yes, I agree, thanks.

specifications/xslt-40/src/xslt.xml
Arithmeticus: Also, I had thought my edits to the specs for `current-grouping-key()` handled this.

ndw:
At [meeting 136](https://qt4cg.org/meeting/minutes/2025/09-30.html), the CG agreed to merge this PR.

PR #2130, opened 2025-07-28
Proposed new operator keywords: precedes, follows

Arithmeticus:
The operators `<<` and `>>`, in my opinion, are poorly known, and challenging for developers working in XSLT. Many punctuation-based operators have aliases in ordinary-language quasi-equivalents, but `<<` and `>>` lack any ordinary verbal equivalents, and break this principle.

The attached proposal offers to make `precedes` a keyword equivalent to `<<` and `follows` a keyword equivalent to `>>`. This means that `//title[. &lt;&lt; following-sibling::isbn[1]]` can now be expressed as `//title[. precedes following-sibling::isbn[1]]`


michaelhkay:
Is it worth introducing `is-not` (or `isnt`) at the same time?

Arithmeticus:
OK, I've added `is-not` -- easy enough to roll back one commit if the CG doesn't like it.

michaelhkay:
Needs a "changes" entry at the start of §4.10.3

Arithmeticus:
thank you, done

PR #2164, opened 2025-08-13
Fix return type in `fn:parse-csv` signature

GuntherRademacher:
In f2e1f48, `fn:parse-csv` was changed to return an empty sequence, when its first argument is an empty sequence. This is however not reflected in the function's return type, which is here changed to `parsed-csv-structure-record?`.


ndw:
At [meeting 133](https://qt4cg.org/meeting/minutes/2025/08-19.html), the CG agreed to merge this PR without discussion

Issue #2221, opened 2025-10-02
fn:unparsed-text, bin:decode-string: Encoding, BOM handling


ChristianGruen:
When trying to write a PR for #2217, I noticed some general inconsistencies that we should clarify first.

### Unifications

`fn:unparsed-text` and `bin:decode-string` use different rules for handling the input encoding:

Rule | `fn:unparsed-text` | `bin:decode-string`
-- | -- | --
A. Require implementations to support… | `utf-8`, `utf-16` | `utf-8`, `utf-16`, `utf-16be`, `utf-16le`
B. Handling of initial `xFF` and `xFE` (before decoding) | ignored | interpret `utf-16` as `utf-16le`
C. Handling of initial `\uFEFF` (after decoding) | discarded | adopted (or interpreted, see PR for #2217)
D. Implementation-defined heuristics to detect the encoding… | allowed | not allowed

I think we should…

  A. require implementations to support `utf-8`, `utf-16`, `utf-16be`, and `utf-16le` in both functions;
  B. always interpret BOMs;
  C. adopt [ZWNBSP](https://www.compart.com/de/unicode/U+FEFF) as Unicode input (after all, it should hardly ever occur at this place);
  D. keep the current behavior unchanged (due to `bin:decode-string` being a more low-level operation)

### BOM handling

As insinuated in B., I would suggest always interpreting a BOM if it exists, and skipping the bytes. If an encoding is specified that conflicts with the inferred encoding, we could…

1. overwrite the inferred encoding,
2. ignore the encoding argument, or
3. raise an error.

Option 1 would be my preference, in adherence to the principle that user input should always be taken serious: If the input cannot be decoded with a specified encoding, this already results in an error (`err:FOUT1200`, most probably, although this needs to be clarified in the specs as well).

PR #2222, opened 2025-10-02
2217 bin:decode-string: Input encoding

### LINKED ISSUE #2217: `bin:decode-string`: Interpretation of `$encoding` argument
When the `$encoding` argument for [`bin:decode-string`](https://qt4cg.org/specifications/expath-binary-40/Overview.html#func-bin-decode-string) is omitted, the `utf-8` encoding is assumed.

I propose to change the function signature to…

```
bin:decode-string(
  $in        as (xs:hexBinary | xs:base64Binary)?,	
  $encoding  as xs:string?                         := ()   (: before: 'utf-8' :)
  $offset    as xs:integer?                        := 0,
  $size      as xs:integer?                        := ()
) as xs:string?
```

…try to interpret the first read bytes as byte order mark and, if no BOM is found, assume UTF-8.

As a result, both `bin:decode-string(xs:hexBinary('41'))` and `bin:decode-string(xs:hexBinary('EFBBBF41'))` would return `A`.

ChristianGruen:
Before discussing this PR, we should have a look at #2221.

Closes #2217

### DIFF
--- a/specifications/EXPath/binary/src/function-catalog.xml
+++ b/specifications/EXPath/binary/src/function-catalog.xml
@@ -841,7 +841,7 @@ $in =!> bin:to-octets() => bin:from-octets()
                   the value of <var>E</var>, or, if absent, <code>utf-8</code> is chosen;
                   otherwise</p></item>
                <item><p>it is chosen as described for <code>fn:unparsed-text</code> in
-                  <bibref ref="xpath-functions-40"/>. If a Unicode encoding is determined,
+                  <bibref ref="xpath-functions-40"/>. If a UTF encoding is determined,
                   a byte order mark at the beginning of the input is ignored.</p></item>
             </olist>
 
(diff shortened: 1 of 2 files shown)


ndw:
It feels like we're rubbing up against some unresolvable problems. Perhaps we should push some of this effort back on the user. After all, an offset of [1] when the string begins with a BOM is *either* exactly what the user wants to do *or* it's very, very likely an error, and I don't think we can tell which.

I think it was Juri who proposed an inspect function, let's say `bin:inspect-binary`. Off the top of my head:

```
bin:inspect-binary(
  $in as (xs:hexBinary | xs:base64Binary)?,
  $options as record(encoding as xs:string?, identify-characters as xs:boolean?, *)
) as record(has-bom as xs:boolean?, encoding as xs:string?, character-indexes as xs:integer*, *)
```

If you pass in an `encoding` in the options, then that's the encoding that's used and returned. If you don't, then the implementation uses heuristics to infer an encoding, which it returns. If the encoding is an appropriate UTF encoding, then the presence (or absence) of the BOM is returned in `has-bom`.

If `identify-characters` is explicitly `true()`, then `character-indexes` is a sequence (or array?) of offsets to the positions where characters start. (I think that's only useful for UTF encodings, but I could be wrong.)

It's a fairly expensive function to call, perhaps, but it means you, the user, can work out if you need to skip the BOM, or align a substring at a position where a character starts.

The `bin:decode-string` function can take an explicit encoding and it it uses that encoding and if the sequence of characters selected isn't usable in that encoding, an error is raised.

Is that at all sensible?

ChristianGruen:
> Is that at all sensible?

Possibly ;) The main reason why we got here was that we were looking for a function that decodes a binary-with-an-optional-BOM to a string. Do you think that `bin:inspect-binary` could help us here?

ndw:
The hard way would be, check if there's a BOM, if there is, decode starting at offset [3].

But I think we could special case the "decode this whole UTF-x string and discard the BOM" behavior if the encoding is UTF-x, begins with a BOM, and no starting offset was given.

ChristianGruen:
> But I think we could special case the "decode this whole UTF-x string and discard the BOM" behavior if the encoding is UTF-x, begins with a BOM, and no starting offset was given.

Yes, I agree. We can either raise an error if an offset without encoding is supplied, or assume `UTF-8`.

PR #2223, opened 2025-10-02
2193 fn:parse-xml, fn:doc: Drop security options

### LINKED ISSUE #2193: fn:parse-xml, fn:doc: Drop security options
Related: #2034, #1270.

I propose to drop the `allow-external-entities` and `entity-expansion-limit` options. They reflect only some of various existing limits that XML parsers provide.

Instead, it seems more reasonable to define such limits globally. It remains to be shown whether we find an implementation-oblivious approach for this.

ChristianGruen:
Closes #2193


ndw:
At [meeting 139](https://qt4cg.org/meeting/minutes/2025/10-21.html), the CG agreed to merge this PR.

PR #2247, opened 2025-10-12
716 Deferred Evaluation in XPath - the f:generator record

dnovatchev:
This PR introduces the notion of  deferred evaluation and presents the tool to achieve this - the built-in named record type `f:generator`.

- Two major use cases are presented.
- The fields of `f:generator` are defined.
- All system-provided methods  of `f:generator` are defined.
- Examples of using each of the methods  of `f:generator` are provided.
- The formal equivalents of each of the system-provided methods  of `f:generator` are provided in a separate appendix.

image


michaelhkay (commented):
Thanks: an immense amount of work has gone into this and the proposal is very comprehensive. As an implementor, I certainly have a concern that it will involve a commensurate amount of implementation effort. With 40 methods specified on generators, I would expect to see something in the order of 800 test cases, which is a massive undertaking (they don't only have to be written, they have to be reconciled with the results of an implementation). As a group, we need to make a decision about whether we really want XPath to be such a big and powerful language (and indeed, to make a realistic estimate of how much time it will take to iron out all the details). We also need to think about whether it might possible to deliver the benefits of the feature using something much simpler. I have to confess that I've developed a bit of an aversion to ambitious proposals, having spent many years of working group time sorting out the details of things like schema-awareness and streaming that turned out only to be needed by a small proportion of the user community.

liamquin:
@Arithmeticus yes, you could for example have a generator next function that goes out and calls some Web API to fetch data, and it would only be fetched when (if) the generator processor needed to use that value. An example might be fetching search results one by one until the "goodness of match" of the search results, that you computed on the fly, was too low to be interesting... e.g. "find me every wikipedia article containing the word sock”, then fetch the result articles one by one and do vector cosine similarity on them to some other document set, rather than fetching them all at once and using a ton of memory.

In a way it’s like xsl:iterate, except for XPath and XQuery too. E.g. for each entry in an 1812 biographical dictionary i put on the Web, i could search the Women Writers project and if it seems it’s the same person, search in the British Library 8th century short title catalogue for a list of their works, and append that to the biographical entry as i do today with the Early English Books Online project data.

So there can be an overlap with XProc, and with XSLT, but the basic idea of calling a user-supplied function to get the next item, only when it’s needed, has wide application.

Arithmeticus:
I'm not asking for a tutorial. I'm asking for consistency with other parts of the specs, in which a function is typically illustrated with a simple example to explain basic functionality, then a second or third more complex example (with different domain) to point to the wider range tasks.

michaelhkay:
Thanks for the excellent comments, they help me to focus my own thoughts about the proposal.

I have to confess I've been distracted by the detail. There are 40 or so methods defined, and I start by looking at one of them:

> make-generator | This method creates an initially empty generator and repeatedly calls $provider to get the next value to add into the generator being populated.The $provider function takes as argument the current number of values the generator, being constructed, already has. The $providerfunction returns a zero-argument function, which when called produces the wanted next value to add to the generator. When $provider returns something that is not a function, this means that all data items have already been produced, and thus the population of the new generator being created is complete. Type: fn($this as f:generator, $provider as function(xs:integer) as item()*) as f:generator

... and trying to read that repeatedly to understand what it means. 

- What exactly is a "generator" (the term doesn't seem to be defined) and what does it mean for it to be empty? I think "generator" is being used to mean "an instance of the f:generator record type", but I don't see how such an instance can be empty or how anything can be added to it, so I am clearly misreading the term. Perhaps the term is also being used to mean "a sequence of values generated by a generator", but the more I study the language, the more confused I become.

- Saying that the generator is "initially empty" and that the method "adds into" it implies that the state of a generator can change. But in that case clearly the "generator" is not the f:generator record instance, which is necessarily immutable.

- The method "repeatedly calls $provider", but what arguments does it supply, and what does it do with the result?

- The specification says what happens when $provider returns a zero argument function, and it says what happens when it returns something that is not a function, but that doesn't cover all possibilities. What happens, for example, when it returns a map, which doesn't fall into either category, but is allowed by the function signature?

- What are "data items"? What is the "population" of a generator? If the "population" of a "generator" is "complete", what does this mean? Elsewhere there is talk of the "contents" of a generator, of the "current member" (and "next member") of a generator, and of the "state" of a generator, and of the "values" that it "currently has".

This is an analysis of my problems with just one of the 40 methods being specified. If I had time, I could probe equally deeply into the other 39, and I would probably end up even more confused. The meta-problem is that I do not have time, and I'm not at all convinced that the CG has time.

There's the germ of a good idea here, but it's only viable if we trim it down to its essentials, and define the concepts very clearly and precisely. Establishing clear and precise definitions for the basic terminology (starting with "generator" itself) would be a good start.

michaelhkay:
That's certainly an improvement, but there's still a lot of vague language, like the reference to "a future call" - in a functional language there is no timeline and temporal language always raises a red flag.

I think I can now see what you intend the method to do, but I still don't understand what I might want to use it for. In particular, I don't understand how the `$provider` function can do anything useful if the only information it has available is an integer sequence number. Examples are needed here.

I would write the specification something like this (but I may have misunderstood your intent).

The method make-generator, when applied to an fn:generator instance G returns a map G' that is an instance of fn:generator and that also contains a field numDataItems [making it an instance of some extended record type which we ought to specify...]. The fields of the returned fn:generator are as follows:

* `initialized` is set to true() [?]
* `end-reached` is set to false() [?]
* `get-current` is set to a method that returns the result of invoking $provider, supplying an integer set to the value of ($G?numDataItems + 1 otherwise 1)
* `move-next` is set to [???]
* `numDataItems` is set to ($G?numDataItems + 1 otherwise 1)

But I may have got this completely wrong. The introduction says that the "user has to maintain the fields" `initialized`, `end-reached`, etc, so I may be quite wrong in thinking that in this case the system maintains them.

Whether I'm wrong or right, what is needed is (a) a simple description of what the method does, that is, how the output fn:generator record is derived from the input fn:generator record, and (b) a realistic example of how it might be used that illustrates why it is useful.

Then repeat this 40 times for all the other methods.

ChristianGruen:
Thanks, Dimitre, for the comprehensive and certainly time-consuming proposal.

I would like to reiterate the idea of defining the generator in a separate spec, similar to the Binary and File Modules:

* Generators would be suitable for this purpose; they are comprehensive and self-contained.
* We would have more freedom to finalize the specifications independently from each other.
* It would be eaiser to add other data structures in the future.

From a technical perspective, I would definitely prefer the functions of generators to be defined as classical built-in system functions. While it is technically feasible to represent them completely as records, working with them would be much easier if they were similar to with sequences, maps and arrays. It is an impressive demonstration of the 4.0 features that generators can be represented as self-contained records, but it comes at the expense of usability.

From an implementation perspective, this refactoring would also provide much more opportunities for optimization. No one would win if we introduce generators that process infinite data only theoretically.

Finally, I agree with others that we need to present at least one convincing and easily relatable use case, unless we want to present l’art pour l’art. My feeling is that the presented examples can all be realized with existing 4.0 functionality. What has been your motivation to propose the generator concept: Do you have any practical use case in which you have processed huge amount of data, and for which the current functionality came to its limits?

Finally, the `f:generator` constructor in the example section looks overly complex, and difficult to understand without reading the full prose. I would propose starting the examples with a simple self-explaining constructor call. What is the minimum sensible number of arguments for `f:generator`?

dnovatchev:
@ChristianGruen, Thank you for the good observations:

> From a technical perspective, I would definitely prefer the functions of generators to be defined as classical built-in system functions.

Yes, indeed. And the `=>` operator is much more pleasing than `=?>` ...

A new version of the `f:generator` record can have only  fields: `initialized`, `end-reached`, `get-current` and `move-next`.
All current system-provided methods will be system functions on generators.

> What is the minimum sensible number of arguments for `f:generator`?

Invoking the `empty-generator` method requires 0 arguments, so this could be the minimum. One certainly can provide something meaningful for the four fields mentioned above - or update the result of `empty-generator` using `map:put`

dnovatchev:
@arithmeticus, thank you for these questions.
Please, see the answers below.

> I find the order of groups of functions confusing. I would like to see 18.4.3 first (make a generator), then 18.4.1, then 18.4.2, then 18.4.4. I question whether we need some of the functions in 18.4.2 or 18.4.4, or if they should not be delayed to a second, later PR, so that the CG can focus on the core specs.

The CG can decide what should be the best ordering of the function groups.


> 
> I do not understand the discussion of `state-data` or `result-data` at 18.4.3. What are these properties of? Are they fields within the `fn:generator` record? If not, what? If so, why are these properties not part and parcel of 18.3's table defining `fn:generator`? 

No, `state-data` and `result-data` are the two members of the array that `$provider` returns.

The original text is precise, let us look at it again:

_A provider function is responsible for producing, on each invocation, both:_

1. _**state-data** - an array with information required for the provider to compute its next result, and_
3. _**result-data** - an array either containing a single yield-value or empty, signaling that the data for producing the yield has been exhausted._

And _**state-data**_ is the single argument that the caller should pass to the provider.

The semantics of _**state-data**_ and _**result-data**_ are fully described in the **Rules** section of of 18.4.3.2 gn:make-generator - the only of the four functions that works with a provider. Here is this original text again:

_**Rules**_
_The `$provider` function is called to get the first value of the yield of the result-generator. It accepts as argument an array, that contains data, referred to as "state-data" here, necessary to identify and return its next result. On the first call to the `$provider`, an empty array is passed as the state-data._

_The `$provider` returns an array with two members. The first member is the state-data to be passed to it on the next call. The second member is an array, containing the current value of the result-generator._

_To signal that there are no more yield-values to be returned the `$provider` returns an empty array as its result's second member. When this happens no more calls to the `$provider` are made and the `move-next` method of the current result-generator-instance is set to return an empty generator._

_If the `$provider` returns a non-empty array as its result's second member, the value of the single member of this non-empty array is set to be returned by the `get-current` method of the current result-generator instance and the `move-next` method of the current result-generator instance is specified in such a way that, when it is called, it would itself call the `$provider` with the latest returned state-data and based on the return of that call would construct correctly the next result-generator instance_.

>What does a `state-data` array look like? 

The power and generality of this definition is that:

1. We don't know, and don't have to know what is the meaning of the "thing" that is in that array - it can be anything that the `$provider` needs.
2. We don't care - this is up to the author of any provider function to decide - could be an index, a URL, a database table key, a disk seek-address, ..., anything that is appropriate to use in order to continue to provide data in a specific order from the actual data source



> 
> Why do we need the term `XPath collection`? The definition, "a sequence, an array or a map" (why "can"?) can be simplified to "a sequence," because a single array is a sequence and a single map is a sequence.

Sure, but these three are different datatypes. "Make things as simple as possible, but not simpler" ... , Einstein.

> 
> The term "external provider function" is in bold, as if it is a term of art, but it nowhere defined. What is the point of the modifier "external"? Does it behave differently than if it were an internal provider function (whatever that is)? 

"External" means exactly that - something about the inner workings of which we have absolutely no knowledge, and the only requirement to it is to have the specified signature. It is a function, that accepts an array containing a single member, and which returns an array containing two members, each of which is an array: its `state-data` to be passed to it on the next call, and an array containing a yield-value that the current generator instance method `get-current` should return when called.

>What is a "provider function"? How does it differ from functions generally? What is a "provider"? I think the specs here should be agnostic about whether the `move-next` function is designed to access something internal, something external, or a mix/alternate of both.

Again, the original text explains this:

_"Providers allow generators to expose yields that are too large to fit into memory, must be computed incrementally, or require access to external systems (such as files, databases, or web services)."_

And the text defines precisely the signature (arguments-types and return type) that any provider function must have, as we already discussed above.

There are specific provider-functions in the three examples in section 18.4.3.2 - one provider function provides the yield-values from 1 to 10, and the second provider-function provides as yield values all natural numbers.

The reader can see exactly what `state-data` and what `result-data` are being returned by each of these provider-functions.

> 
> I have other questions but these are the big ones that I encountered reading tonight.

Once again, thank you for reading, and for the questions.

liamquin:
> I find the order of groups of functions confusing.

Probably it could be improved. But, the CG has at least to have broad consensus to accept the proposal first :-)
It might be that a small example earlier would help.

> I question whether we need some of the functions in 18.4.2 or 18.4.4, 
There are a couple of functions that seem pointless until you see that they are there to avoid bugs in users’ code, which is a good reason to put them in the spec, i think.

>"compute its next result" implies that the state-data array contains not the data itself for the next result but rather a function that can be invoked to get that next round of data. 
Yes. This is where the name comes from :-) generating the sequence of values one item at a time.

> Why do we need the term `XPath collection`?
I am not sure it is needed.

I hope that we will post some use cases later this month, with some worked examples. But, consider a generator that uses some external metadata to return the next most likely TEI-encoded book to consider for a putative match on a misspelt author... the query needs only to look at enough of the books to have confidence two authors represent the same person, and fetching 60,000 books into memory would be a problem. So the generator has a function that stores n (the nth book) and fetched the n+1th book on request, and makes a new generator function with n one larger. And when the caller simply stops using the generator, the rest of the possible values are never generated.

specifications/xpath-functions-40/src/xpath-functions.xml
```diff
@@ -10512,6 +10521,393 @@ return <table>
           </div1>
       
 
+      <div1 id="lazy-evaluation">
+	      <head>Deferred evaluation</head>
+		  <p>Deferred-evaluation is achieved in a user-controlled way, using the system-defined <code>fn:generator</code> record.</p>
+		  <p>Deferred-evaluation is particularly useful when solving any problem, where data sources of unknown or unpredictable size 
+		    must be handled, and only one or a few values from these are needed at any time during the processing. </p>
```
xmljacquard:
```suggestion
		    must be handled and only one or a few values from these are needed at any time during the processing. </p>
```

specifications/xpath-functions-40/src/xpath-functions.xml
```diff
@@ -10539,6 +10548,393 @@ return <table>
...
+			
+			<note>
+			<p>Most of the examples in this section depend on the definition of the following variables:</p>
+			<example>			  
+			  <eg>
+$gen2ToInf := f:generator(initialized := true(), end-reached := false(), 
+                              get-current := fn($this as f:generator){$this?last +1},
+                              move-next := fn($this as f:generator)
+                              {
+                                if(not($this?initialized))
```
aljopainter: Given that the entry `initialized` is set to `true()`, do we really need the check for `$this?initialized` here? I understand that, in the general case, you cannot count on `initialized` always being `true()`, but in this specific example is it `true()`.

aljopainter (commented):
I have a few remarks but I'm quite afraid that I'm not helping much as I don't have any concrete suggestions for moving forward.

specifications/xpath-functions-40/src/xpath-functions.xml
```diff
@@ -10539,6 +10548,393 @@ return <table>
...
+                                  then map:put($this, "initialized", true())
+                                  else map:put($this, "last", $this?last + 1)
+                              }
+                             ) => map:put("last", 1),
+    $genN := $gen2ToInf => gn:for-each(fn($n) {$n - 1}),
```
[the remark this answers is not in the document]
dnovatchev: Again my fault. In Mathematics the set of Natural numbers has the widely accepted and used name **N**, sometimes stylistically written as **_`N|`_**. If people think that `$gen1ToInf` is a better name, we will rename it.

specifications/xpath-functions-40/src/xpath-functions.xml
dnovatchev: @aljopainter,
Can we resolve these comments for now, if you feel that the provided answers are satisfactory?

We will **definitely** return to these once the PR is ready to be merged.

ChristianGruen:
@dnovatchev I have added the `Revise` label: We should definitely look at the PR together when its presentation is aligned with the Binary and File modules.

Issue #1704, opened 2025-01-16
Ignore the byte order mark more completely/globally


ndw:
Following on a discussion with @line-o on the XML.com Slack, I took a peek at the way we deal with the byte order mark in Functions and Operators. We seem to be explicit about it in a couple of JSON functions but not elsewhere. I think we should assert that the byte order mark is explicitly ignored in all of the input functions (json-*, parse-*, unparsed-* etc.)

michaelhkay:
Yes I agree.

ndw:
I've tried to look at this at one more level of detail.

1. If the user specifies an encoding of UTF-8 and the first bytes are EF BB BF, that's the BOM, ignore them.
2. If the user specifies an encoding of UTF-16 and the first bytes are FE FF or FF FE, ignore them
3. If the user specifies an encoding of UTF-16LE and the first bytes are FE FF, ignore them
4. If the user specifies an encoding of UTF-16BE and the first bytes are FF FE, ignore them

If the user doesn't specify an encoding, we have some rules that run through to implementation defined heuristics followed by UTF-8. I guess that's okay. We should probably say that if the encoding selected is UTF-8 or one of the flavors of UTF-16, the rules above apply.

line-o:
The interesting discussion to have here is: Which one has precedence if BOM and encoding declaration are mismatching? Or is an error raised in that case?

ndw:
Well, XML says:

> It is a [fatal error](https://www.w3.org/TR/REC-xml/#dt-fatal) when an XML processor encounters an entity with an encoding that it is unable to process. It is a [fatal error](https://www.w3.org/TR/REC-xml/#dt-fatal) if an XML entity is determined (via default, encoding declaration, or higher-level protocol) to be in a certain encoding but contains byte sequences that are not legal in that encoding. Specifically, it is a fatal error if an entity encoded in UTF-8 contains any ill-formed code unit sequences, as defined in section 3.9 of Unicode [[Unicode]](https://www.w3.org/TR/REC-xml/#Unicode). Unless an encoding is determined by a higher-level protocol, it is also a [fatal error](https://www.w3.org/TR/REC-xml/#dt-fatal) if an XML entity contains no encoding declaration and its content is not legal UTF-8 or UTF-16.

[...]

PR #2249, opened 2025-10-14
2221 fn:unparsed-text: Encoding, BOM handling

### LINKED ISSUE #2221: fn:unparsed-text, bin:decode-string: Encoding, BOM handling

ChristianGruen:
Closes #2221

Note that this PR is also a revision of #1950, which I felt was a bit ambiguous, and did not cover all possible cases. I believe that the proposed changes are compatible with 3.1. In particular, it states more clearly what is going to happen if both an implicit encoding is determined (for example, via a BOM), and an explicit encoding is supplied.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -19278,49 +19278,43 @@ return tokenize(normalize-space($s), ' ')[. castable as xs:IDREF]</eg>
             to recognize are <code>utf-8</code>, <code>utf-16</code>, <code>utf-16le</code>,
             and <code>utf-16be</code>.</p>
 
-         <p>The implicit encoding <var>E</var> of the external resource is determined as follows:</p>
+         <p>The encoding candidate <var>E</var> is chosen as follows:</p>
          <olist>
             <item>
-               <p>External encoding information is chosen if available.</p>
+               <p>external encoding information if available; otherwise</p>
             </item>
             <item>
-               <p>Otherwise, if the media type of the resource is <code>text/xml</code> or
-                     <code>application/xml</code> (see <bibref
-                     ref="rfc2376"
-                     />), or if it matches
-                  the conventions <code>text/*+xml</code> or <code>application/*+xml</code> (see
-                     <bibref
-                     ref="rfc7303"
-                     /> and/or its successors), then the encoding is recognized
-                  as specified in <bibref
-                     ref="xml"/>.</p>
+               <p>the encoding recognized as specified in <bibref ref="xml"/> if the media type
+                  of the resource is <code>text/xml</code> or <code>application/xml</code>
+                  (see <bibref ref="rfc2376"/>), or if it matches the conventions
+                  <code>text/*+xml</code> or <code>application/*+xml</code>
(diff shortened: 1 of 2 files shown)


ChristianGruen:
I have updated the PR; it includes considerations from #1704 that the bytes of a BOM may be part of the input.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -19272,55 +19272,49 @@ return tokenize(normalize-space($s), ' ')[. castable as xs:IDREF]</eg>
...
-            <phrase diff="add" at="2022-12-19"> and non-empty</phrase>, is the name of an encoding. The values
-            for this option follow the same rules as for the <code>encoding</code> attribute in
-            an XML declaration. The only values which every
-            implementation is <rfc2119>required</rfc2119> to recognize are
-               <code>utf-8</code> and <code>utf-16</code>.</p>
-         <p>The encoding of the external resource is determined by applying the following rules, in order:</p>
+            <phrase diff="add" at="2022-12-19"> and non-empty</phrase>, is the name of an encoding.
+            The values for this option follow the same rules as for the <code>encoding</code> attribute
+            in an XML declaration. The values which an implementation is <rfc2119>required</rfc2119>
+            to recognize are <code>utf-8</code>, <code>utf-16</code>, <code>utf-16le</code>,
```
Arithmeticus: According to the rule 4 for effective encoding below, there will never be a case where `utf-16` is an option. It's always big endian or little endian. Not sure how/if you want to resolve that very minor discrepancy.

Arithmeticus (commented):
The Oxford/serial comma is needed for clarity in a couple of places. Other very minor suggestions.

Issue #938, opened 2024-01-10
Canonical serialization


Arithmeticus:
This issue picks up suggestions from #779 regarding canonical serialization, and solicits from the community group input on if such a function is desirable, and what such a function might look like. 

In the context of #779, the idea was that two XML documents with different physical representations, but semantically equivalent, could be serialized to a canonical form, with a hash value applied to each confirming identity. Of course, with canonical operation, a simple string comparison would be sufficient, absent any hashing.

XML Signature was suggested as one approach, with some hesitation. I would like to suggest, instead, that we look to implement [Canonical XML Version 1.1](https://www.w3.org/TR/xml-c14n/) (herein **CX1.1**), perhaps with map options that calibrate how CX1.1 is implemented. I have no experience using CX1.1, so user input is welcome.

Another point of discussion is whether this merits a new function, e.g., `fn:canonical-serialize`, or should be built upon `fn:serialize`. A problem with the latter option, is that such an approach makes no sense without the `method` option specified as `xml`. Another approach would be to go deeper, into the serialization spec, and expand the `xml` method to ensure a canonical option.

I believe that this function would be extremely useful. When preparing test suites, output could be saved as secondary documents as canonical XML, and any subsequent regression tests could adjust comparanda to canonical XML, and very precise node-wise comparisons could be made.

I look forward to everyone's input.

michaelhkay:
(a) Yes I think it would be useful.

(b) I think my preference would be to use `fn:serialize` with `method="canonical-xml"`. An alternative is to use `method="xml" canonical="yes"`, but this has the disadvantage that there are many interactions with other serialization options, e.g. `indent`, `cdata-section-elements`, and `omit-xml-declaration`.

Note, if you want to experiment, Saxon already offers `method="xml" saxon:canonical="yes": see https://www.saxonica.com/documentation12/index.html#!extensions/output-extras/serialization-parameters. I tested this against the canonicalizer offered by XOM.

ChristianGruen:
I also think this would be useful.

> (b) I think my preference would be to use `fn:serialize` with `method="canonical-xml"`. An alternative is to use `method="xml" canonical="yes"`, but this has the disadvantage that there are many interactions with other serialization options, e.g. `indent`, `cdata-section-elements`, and `omit-xml-declaration`.

+1 for adding a custom method, and we should raise an error if input is not a single node.

Arithmeticus:
[...]

PR #2259, opened 2025-10-20
938 Canonical serialization

Arithmeticus:
This represents a first draft to resolve issue #938 (and thereby slightly reduce our workload).

Because both JSON and XML have canonical specifications, and there is nothing barring canonicalization being imposed upon other methods, I have opted to introduce the serialization parameter `canonical` as a boolean, rather than extend the number of method options. 

I had to make certain decisions in this first draft, worth discussing.
* The CX1.1 specs permit a parameter on whether to keep comments or not -- I think the user should deal with that before trying to serialize.
* I opted not to specify what happens for escape-solidus, use-character-maps, normalization-form and some other options, where user control may be desired.
* When a JSON object has an XNode, and when canonical is yes, I think the current serialization parameters should propagate down to the sub-serialization process.
* I think the normalization step I added to 4 Serialization may not be necessary, or incorrectly conceived. 
* I opted to avoid specifying exactly what changes should be made to the JSON or XML, deferring to the external specifications. In line with that, I attempted to treat canonicalization as an encapsulated postprocess that could be cleanly disentangled from all the other options involved in serialization.

I've tried to make patches everywhere relevant, but I'm guessing there are places I've overlooked. Comments welcome.

### DIFF
--- a/specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
+++ b/specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
@@ -1764,10 +1764,6 @@ by ignoring the request to output a document type declaration or
                 <rfc2119>MUST</rfc2119> raise an error on documents containing relative namespace
               URIs, or on octet streams that do not contain a well-formed XML document.</p>
       </note>
-      <note>
-        <p>If <code>use-character-maps</code> is not empty, character expansion may be altered
-          by the later canonicalization process specified at <specref ref="serphases"/>.</p>
-      </note>
       <p>A serialization error <errorref code="0024" class="RE"/> results if any error is raised during the 
         canonicalization process.</p>
     </div3>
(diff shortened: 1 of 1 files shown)


Arithmeticus:
Excellent feedback, much appreciated. I'll work on this in the coming week.

Arithmeticus:
In thinking through remarks 7 and 8 I made decisions that are subject to discussion at the next CG meeting.

michaelhkay:
There is a note: _"If use-character-maps is set to true, character expansion may be altered by the later canonicalization process specified at [4 Phases of Serialization]_". But use-character-maps is not a boolean.

ndw:
At [meeting 143](https://qt4cg.org//meeting/minutes/2025/11-25.html), the CG agreed to merge this PR.

michaelhkay:
I have added a few token tests just to establish that the feature is present, but detailed testing (to show that the output actually conforms to canonical XML) is still needed.

PR #2286, opened 2025-11-13
2279 fn:string-length#1, fn:normalize-space#1: accept xs:anyAtomicType

### LINKED ISSUE #2279: `fn:string-join#1`: accept `xs:anyAtomicType`
`123 ! string-length()` is interpreted as `123 ! string-length(string(.))`, while `string-length(123)` raises an error. This is confusing.

Maybe we should allow `fn:string-length#1` to accept input of type `xs:anyAtomicType`; this would also make it symmetric to `fn:string-join`.

ChristianGruen:
Closes #2279


michaelhkay:
string-join#1 in the title should be string-length#1

ChristianGruen:
> If an argument is supplied, it would generate fewer errors, but it should not change results.

…and I am starting to doubt if that is really the case.

I should do some training in typed XML data. It leads to scenarios I could never have imagined. ;)

PR #2293, opened 2025-11-14
Updated RELAX NG grammar for XSLT 4.0 stylesheets

ndw:
This is mostly a copy of #2290, translated into the RELAX NG grammar.

I think the grammar still needs review, but I’m trying to keep it up-to-date with the XSD in the meantime.

### DIFF
--- a/specifications/xslt-40/src/schema-for-xslt40.rnc
+++ b/specifications/xslt-40/src/schema-for-xslt40.rnc
@@ -36,7 +36,7 @@ start =
   | package.element  
   | literal-result-element-as-stylesheet
   
-sequence-constructor.model = (instruction.category | literal-result-element | text)*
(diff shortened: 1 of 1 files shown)


davidcarlisle:
You could also consider supplying a wrapper xslt.rnc that (I think?) I got from you sometime around xslt 2 being released.
It's a really useful trick that I've used in other contexts as well, but isn't as well known as it could be:

This allows the validator (and in particular emacs editing help) to automatically switch the grammars after the version attribute is set.

## xslt.rnc

```
xslt10 = external "xslt10.rnc"
xslt20 = external "xslt20.rnc"
xslt30 = external "schema-for-xslt30.rnc"
xslt40 = external "schema-for-xslt40.rnc"
start = xslt10 | xslt20 | xslt30 | xslt40
```

PR #2309, opened 2025-11-20
2299 Allow SimpleMapExpr after ArrowExpr

### LINKED ISSUE #2299: Make the pipe-like operators left associative and have the same precedence
Is it possible to make the operators `/`, `//`, `!`, `?`, `=>`, `->`  **left associative** (at least) and have the same precedence? This would avoid (very annoying) extra parenthesis.

Currently, we cannot use `/`, `!`, `?` after `=>` . For example, the following expression (intentionally simplified) does not work:
```xquery
 (1,2,3) => exists() ! count(.)
```
It only works with parenthesis:
```xquery
 ( (1,2,3) => exists() ) ! count(.)
```

Is there any compelling reason for this?

ruv:
Fix #2299

Allow the form `A => B() ! C` and make it equivalent to `( A => B() ) ! C`.  Also, allow all operators from `PostfixExpr` in the same manner.

### DIFF
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -1812,13 +1812,14 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
   <g:production name="SimpleMapExpr" condition="&gt; 1" if="xpath40 xquery40  xslt40-patterns">
     <g:ref name="PathExpr"/>
     <g:zeroOrMore>
-      <g:ref name="SimpleMapTarget"/>
+      <g:string>!</g:string>
+      <g:ref name="PathExpr"/>
     </g:zeroOrMore>
   </g:production>
 
   <g:production name="SimpleMapTarget" if="xpath40 xquery40  xslt40-patterns">
     <g:string>!</g:string>
-    <g:ref name="PathExpr"/>
+    <g:ref name="StepExpr"/>
   </g:production>
 
   <!-- ] end OrExpr etc -->
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -23471,7 +23471,7 @@ return string-join($chopped, '; ')
          </scrap>
 
          <scrap>
-            <prodrecap ref="ArrowExpr"/>
+            <prodrecap ref="SimpleMapTarget"/>
          </scrap>
 
          <p>


ruv:
> It seems odd that this allows
> 
> `$string => tokenize() ! upper-case(.)`
> 
> without also allowing
> 
> `$node => siblings() / *`

Yes, it's my omission. I was planning to allow the later (and other) as well. I will do so.

ruv:
> There's definitely a need to add some prose (semantics, examples) alongside the grammar change.

I added some examples and `RelativePathTarget` (for `/` and `//`). Concerning the semantics, I don't see any changes for `!`:
  - > A mapping expression `S!E` evaluates the expression `E` once for every item in the sequence obtained by evaluating `S`.

For example, in  `a => b() ! c`,    the expression `a => b()` is `S`, and `c` is `E`. 

Overall, this approach seems to be working. However, the solution is quite lengthy, as it requires special handling of each operator in `PostfixExr` (not yet implemented).  Furthermore, it complicates understanding of the formal grammar.

I think, a better approach is to make the unary plus/minus operators more narrow. This allows to easily formalize piping of all pipe-like operators.

> it would be nice if you introduced yourself.

Yes, thank you. I would like to attend some of the following meetings.

michaelhkay:
Isn't there an ambiguity here with the parsing of

`a => b() ! c / d / e`

The ArrowTarget could be followed by `! c / d / e` as a SimpleMapTarget, or it could be a SimpleMapTarget `! c` followed by two RelativePathTargets `/ d` and `/ e`.

My instinct, though I haven't worked out the detail, would be instead to build on the PostfixExpr construct.

Something on the lines of:

```
PostfixExpr ::= PrimaryExpr Suffix*
Suffix := Predicate | ArgumentList | Lookup | ...
ArrowExpr ::= UnaryExpr ArrowTarget Suffix*
```

But I suspect it's a lot more complicated than that (and it's going to be rather disruptive change, and one where it's very hard to verify non-ambiguity and backwards compatibility).

ruv:
> Isn't there an ambiguity here with the parsing of
> `a => b() ! c / d / e`

Yes, I also tested that — you are absolutely right!
I have fixed this by replacing `PathExpr` with `StepExpr` in `SimpleMapTarget`.

> My instinct, though I haven't worked out the detail, would be instead to build on the `PostfixExpr` construct.
>
> Something on the lines of:
>
> ```abnf
> PostfixExpr ::= PrimaryExpr Suffix*
> Suffix := Predicate | ArgumentList | Lookup | ...
> ArrowExpr ::= UnaryExpr ArrowTarget Suffix*
> ```

Yes, I also looked in a similar direction, when I wrote that "it requires special handling of each operator in PostfixExr" — that is, every postfix operator production will be changed.

One problem is that in a straightforward approach we will have an ambiguity with `Predicate`.
For example, in `a => b() ! c [d]`,   the part `c [d]` can be parsed as:
  - `StepExpr{ AxisStep{ c[d] } }`
  - `StepExpr{ AxisStep{ c } }  Suffix{ Predicate{ [d] } }`

It would be much easier to add `ArrowTarget` to `StepExpr`:
```abnf
StepExpr ::= ( PostfixExpr | AxisStep ) (ArrowTarget+ Suffix+)*
```
But a problem is that:
   - while the expression `- a => b() ` is parsed as ` (- a) => b()`,
   - the expression ` - a => b() ! c `  will be parsed as   `- ( a => b() ! c )`.

PR #2313, opened 2025-11-24
2298 XQFO rules: definition of default values

### LINKED ISSUE #2298: XQFO rules: definition of default values
The existing XQFO rules are pretty heterogeneous when it comes to presenting the handling of default values.

Sometimes, they repeat what the signature says…

#### https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-name

> ```
> fn:name(
>   $node  as node()?  := .
> ) as xs:string
> ```
> If the argument is omitted, it defaults to the context value (`.`).

Sometimes, it is further elaborated…

#### https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-string

> ```
> fn:string(
>   $value  as item()?  := .
> ) as xs:string
> ```
> In the zero-argument version of the function, `$value` defaults to the context value. That is, calling `fn:string()` is equivalent to calling `fn:string(.)`.

…sometimes, it is ignored:

#### https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-upper-case

> ```
> fn:upper-case(
>   $value  as xs:string?	
> ) as xs:string
> ```
> If `$value` is the empty sequence, the zero-length string is returned. […]

I would propose to revise the rules, and to consistently remove phrases on choosing default values. I believe this will make the rules better readable (once you have understood what default values are, there should be no need explaining it again and again). If we want to keep the information, though, we should add it everywhere.

I can make an attempt if no one objects.

ChristianGruen:
@michaelhkay I have decided to create this PR, even if we are still discussing its necessity. I hope it clarifies where I see the need for unifications in the current presentation. In general, it follows the proposal that I developed in my thoughts in #2298. In particular, it fixes various small inconsistencies in the prose.

I have stumbled upon related editorial issues that I haven’t worked at yet: [1.6 Function calls](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#id-function-calls) is outdated, for example the `fn:function-name` (which could be confused with the actually existing function with the same name) should have an optional parameter. Next, the text  talks about “two signatures of `fn:string`”; and it is not clear why the “zero or one values” and “zero or more values” cases are presented in more detail, whereas “one or more” is missing. The reason is probably simple: It seems to stem back to version 3.1.

Closes #2298

### DIFF
--- a/specifications/xpath-functions-40/src/xpath-functions.xml
+++ b/specifications/xpath-functions-40/src/xpath-functions.xml
@@ -632,7 +632,7 @@ Michael Sperberg-McQueen (1954–2024).</p>
                   value is used. If the parameter is marked <code>(:default-on-empty:)</code>,
                   the default value is also used when the supplied argument is the empty sequence.
                   Otherwise, the empty sequence is treated according to the function-specific rules.</p>
-                  <p>The marker is only permitted on parameters that have default values.</p>
+                  <p>The marker is only used on parameters that have default values.</p>
                </note>
 
                <p>If there are two or more parameter declarations, they are separated by a comma.</p> 


ChristianGruen:
Thanks for the discussion. I will keep the PR open, and I will incorporate the changes in the new presentation, once we have decided how it will look like.

Arithmeticus:
Oops, `string-join` wasn't the best example, and the contrast wasn't clear.

If the empty sequence will be changed to the default value, keep the signature as it is:
```
fn:string-join(
  $values     as xs:anyAtomicType*,
  $separator  as xs:string?         := ""
) as xs:string
```

Add to the signature only if the empty sequence will not be changed to the default value:
```
fn:adjust-dateTime-to-timezone(
   $value     as xs:dateTime?,
   $timezone  as xs:dayTimeDuration?  := fn:implicit-timezone() unless ()
) as xs:dateTime?
```

ndw:
Here's my attempt from this morning. It wasn't too difficult to get the note and the link. (Though there was some fun with the schema, but that's a story for another day.)  You click on the link and you go to the corresponding paragraph. I can't think of any way to mark the link target such that it's clear it's the note (short of rewriting the rules in a more substantial way). 

---

Screenshot 2026-01-14 at 09-48-53 XPath and XQuery Functions and Operators 4 0

---

The link target in `fn:string-join` is the first paragraph in the rules.

---

Screenshot 2026-01-14 at 09-49-28 XPath and XQuery Functions and Operators 4 0

---

The link target in `fn:adjust-dateTime-to-timezone` is the fifth paragraph in the rules.

It's obvious when you click on the note, but I don't know if that's enough.

ndw:
No, I don't have any questions. I confess, however, that I've lost track of what the next steps are.

ndw:
The implementation in that example, by the way is this:

1. If there's a `@note`, annotate with `(: Note :)` pointing to the note.
2. If there isn't a note, and the default is `()`, don't annotate it.
3. If there isn't a note and the default isn't `()`, annotate it with `(: Default :)` linked to the default in 1.5.3.

ndw:
I pushed [a branch](https://github.com/ndw/qtspecs-xslt4/tree/pr-2313-alt) with the changes:

```
        modified:   schema/xmlspec.dtd
        modified:   specifications/xpath-functions-40/src/fos.scm
        modified:   specifications/xpath-functions-40/src/fos.xsd
        modified:   specifications/xpath-functions-40/src/function-catalog.xml
        modified:   specifications/xpath-functions-40/src/xpath-functions.xml
        modified:   specifications/xpath-functions-40/style/merge-function-specs.xsl
        modified:   specifications/xpath-functions-40/style/xpath-functions.xsl
        modified:   style/xmlspec-2016.xsl
```

I haven't made it a PR on the qtspecs repo, but I can if you have trouble getting it from my branch...

ChristianGruen:
I have done minor revisions:
* `(:empty:)` is now `(:empty-on-default:)`
* The comment appears to the left of the default
* The 2nd and 3rd table column were merged.

Original version:



Intermediate version:



New version:

PR #2345, opened 2025-12-15
2299 Expand pipeline to allow arrow expression in path expression

### LINKED ISSUE #2299: Make the pipe-like operators left associative and have the same precedence

ruv:
Fix #2299 
Supersedes PR #2309 

Allow an arrow expression in a path expression (and, by implication, in a simple map expression):
```xquery
$x => B() / foo ! bar => C() ! baz / quz
```

This implementation does not change the relative precedence of `/` and `!`.

---

In this implementation we divide `UnaryExpr` to `SignedUnaryExpr` (always with a leading unary sign) and `UnsignedUnaryExpr` (always without a leading unary sign). And the new feature is supported only in the latter case, since in the former case we cannot avoid ambiguity:
  - an expression `- aaa => B() / ccc => D()` can be parsed both:
      - as `(- aaa ) => B() / ccc => D()`
      - and as  `- ( aaa => B() / ccc ) => D()`
       - because `aaa => B() ` is the left-hand side operand of `/`; see also a [comment](https://github.com/qt4cg/qtspecs/issues/2299#issuecomment-3544885512).

### DIFF
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -1674,17 +1674,21 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
   </g:production>
 
   <g:production name="UnsignedUnaryExpr">
-    <g:ref name="ValueExpr"/>
+    <g:ref name="ValueExprU"/>
   </g:production>
 
   <g:production name="ValueExpr">
+    <g:ref name="SimpleMapExpr"/>
+  </g:production>
+
+  <g:production name="ValueExprU">
     <g:choice>
       <g:ref name="ValidateExpr" if="xquery40"/>
       <g:ref name="ExtensionExpr" if="xquery40"/>
       <g:ref name="SimpleMapExpr"/>
     </g:choice>
   </g:production>
-  
+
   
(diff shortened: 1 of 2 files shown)


michaelhkay:
I can't see where the `=>` and `=!>` operators fit into the proposed grammar.

In any case, it still seems to have unbounded lookahead.

ndw:
@ruv, are you planning to do any more work on this PR?

ndw:
Progress on this PR seems to have stalled. I propose that we close it. It can be reopened if further development occurs.

ndw:
At [meeting 162](https://qt4cg.org/meeting/minutes/2026/04-28.html), the CG agreed to close this PR until further development occurs.

Issue #2315, opened 2025-11-26
Optimization of the workflow action checkout


ruv:
https://github.com/qt4cg/qtspecs/blob/11f17414d317121b333aa2b56553c7f6fea91fc2/.github/workflows/build-pr.yml#L15-L19

@ndw, is it really necessary to [fetch](https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4) the entire history for all branches and tags?  This step is quite long ([example](https://github.com/qt4cg/qtspecs/actions/runs/19700321991/job/56435838465)) since the full repo is huge.

If only the latest version is needed, the line `fetch-depth: 0` can be changed to `fetch-depth: 1` or removed (in all occurrences).

ndw:
As I recall (but perhaps its worth checking again), the problem is that if you only check out the most recent commit, then you can't get the merge base branch. The merge base branch is needed in order to get the diffs. I think I did the whole repository on purpose, even though it's slow.

Suggestsions for improvements, most welcome.

ndw:
At [meeting 172](https://qt4cg.org/meeting/minutes/2026/07-14.html), the CG agreed to close this issue without further action.

PR #2383, opened 2026-01-14
Attempt to resolve action QT4CG-148-01

ndw:
Per #2315:

1. Added ‘at-risk’ changes to the fn:insert-separator, array:members, and array:of-members
2. Added ‘at-risk’ changes to the XPath/XQuery section on map and array filtering
3. Added a note about what ‘at risk’ means to the status sections


ndw:
I did wonder, ... update pushed.

PR #2467, opened 2026-02-17
Harmonize the fn: and file: functions that read text

### LINKED ISSUE #2460: file:read-text and invalid XML characters
`file:read-text()` should be aligned with `fn:unparsed-text()` to remove the restriction regarding valid XML characters.

Incidentally, `bin:decode-string()` has never had such a restriction even in the EXPath 1.0 version.

ndw:
Close #2460 

This PR harmonizes the functions fn:unparsed-text, fn:unparsed-text-lines, file:read-text, and file:read-text-lines with respect to handling non-permitted characters. Each function has an options parameter and that parameter may contain a fallback function to remap non-permitted characters.

This leaves unresolved the question of what to do about permitted characters not allowed in XML, but that’s orthogonal. Strings containing such characters might arise from any of these functions, but equally, might arise from other operations.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -19489,6 +19489,20 @@ return tokenize(normalize-space($s), ' ')[. castable as xs:IDREF]</eg>
                   </fos:value>
                </fos:values>
             </fos:option>
+            <fos:option key="fallback">
+               <fos:meaning>
+                  Provides a function which is called when the input contains a character
+                  that is not a <termref def="dt-permitted-character"/>. It is called once for
+                  each non-permitted character, passing in the Unicode codepoint of the
+                  character. For backwards compatibility,
+                  this option has no default. If no fallback function is
+                  provided, a dynamic error is raised <errorref class="UT" code="1190"/>
+                  if the input contains a non-permitted character.
+                  If the function returns the empty
+                  sequence, the character is simply removed from the input.
+               </fos:meaning>
+               <fos:type>(fn(xs:integer) as xs:anyAtomicType?)?</fos:type>
+            </fos:option>
          </fos:options>
                 
          <p>The way in which an absolute URI is dereferenced to obtain an external resource is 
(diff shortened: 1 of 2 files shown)


ndw:
I constructed a document that ends with an unpaired high surrogate to use in testing and discoverd that `java.nio` throws a `MalformedInputException` when it's read. I may or may not pursue that any further. I will test, but expect the same outcome, from a document that contains 0xDEAD.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -19511,6 +19489,20 @@ return tokenize(normalize-space($s), ' ')[. castable as xs:IDREF]</eg>
                   </fos:value>
                </fos:values>
             </fos:option>
+            <fos:option key="fallback">
+               <fos:meaning>
+                  Provides a function which is called when the input contains a character
+                  that is not a <termref def="dt-permitted-character"/>. It is called once for
+                  each non-permitted character, passing in the Unicode codepoint of the
+                  character. For backwards compatibility,
```
Arithmeticus:
```suggestion
                  character as an integer. For backwards compatibility,
```

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -19711,8 +19703,21 @@ return tokenize(normalize-space($s), ' ')[. castable as xs:IDREF]</eg>
                <fos:type>xs:string?</fos:type>
                <fos:default>()</fos:default>
             </fos:option>
+            <fos:option key="fallback">
+               <fos:meaning>
+                  Provides a function which is called when the input contains a character
+                  that is not a <termref def="dt-permitted-character"/>. It is called once for
+                  each non-permitted character, passing in the Unicode codepoint of the
+                  character. For backwards compatibility,
```
Arithmeticus:
```suggestion
                  character as an integer. For backwards compatibility,
```

Arithmeticus (commented):
I prefer the fallback feature because it allows me to intercept the faulty codepoints directly as integers. It would be a hassle to push the results to a binary and find the problems in `bin:decode-string` then cast back to a string.

PR #2510, opened 2026-03-09
Harmonize functions that read text (supersedes #2467)

ndw:
Just make the fallback option a boolean.

I propose this as an alternative to #2467


Arithmeticus (commented):
Two minor points, and a point of confusion.

Issue #2573, opened 2026-04-02
Generated test case for codepoints-to-string(0)


michaelhkay:
This test case is marked as having a "narrative" result, but the narrative result allows a dynamic error, and the generated test case doesn't allow for this.

michaelhkay:
Fixed by #2579

PR #2579, opened 2026-04-10
Accept narrative errors in generated tests

ndw:
If an `fos:result` contains `errorref` elements, allow those errors to be accepted as a pass.


michaelhkay:
Fix #2573

Fix looks good. A more general approach would also allow `xerrorref` returning errors outside the FOXX9999 family.

ndw:
Oh, yes. I didn't see an example of that, but you're absolutely right.

PR #2594, opened 2026-04-27
2389 Adaptive Serialization: more freedom

### LINKED ISSUE #2389: Adaptive Serialization: more freedom?
The adaptive serialization method was introduced “for the purposes of debugging query results”. For our processor, it has turned out pretty soon that it does not satisfy the requirements of our users, which is why we have introduced a custom debugging method.

I wonder what others think: Shouldn’t we relax several of the rules and let the implementation decide what to output? We haven’t defined either how the output of `fn:trace` needs to look like.

Some examples:
* The output of doubles often causes confusion. If parsed JSON is output, small integers will be output in exponential notation. For example, `parse-json('{ "A" : 20 }')` needs to be output as `{ "A": 2.0e1 }`.
* `xs:date("2001-01-01")` is output as `xs:date("2001-01-01")`, while `xs:token('x')` is output as `"x"`.
* `fn() { 1 }` is output as `(anonymous-function)#0`, whereas an implementation could prefer to use the output of `fn:function-identity` (see #2388), or output the original query string (if available), reproduce a string representation of the function body, etc.

I will be glad to create a PR.

ChristianGruen:
Closes #2389

### DIFF
--- a/specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
+++ b/specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
@@ -4061,7 +4061,7 @@ according to the <bibref ref="JSON-LINES"/> format, as described in <specref ref
     <change issue="2388" PR="2485" date="2026-02-20">
       The adaptive serialization of anonymous functions is now implementation-dependent.
     </change>
-    <change issue="2389" PR="TODO" date="2026-04-27">
+    <change issue="2389" PR="2594" date="2026-04-27">
       The rules for serializing atomic items have been revised and relaxed.
     </change>
   </changes>
@@ -4154,6 +4154,9 @@ as a URI-qualified name (that is, in the form <code>Q{uri}local</code>).</p>
       <code>http://www.w3.org/2001/XMLSchema</code>, using the conventional prefix
       <code>fn</code>, <code>math</code>, <code>map</code>, <code>array</code>, or <code>xs</code>
       as appropriate; or</p></item>
+    <item><p>an <termref def="impdef">implementation-defined</termref> conventional prefix and the
+      local name (<code>prefix:local</code>) if the name is in some other namespace for which the
+      implementation defines such a prefix; or</p></item>
     <item><p>the URI-qualified name otherwise (<code>Q{uri}local</code>).</p></item>
   </ulist>
 </item>
--- a/specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
+++ b/specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
@@ -4061,6 +4061,9 @@ according to the <bibref ref="JSON-LINES"/> format, as described in <specref ref
     <change issue="2388" PR="2485" date="2026-02-20">
       The adaptive serialization of anonymous functions is now implementation-dependent.
     </change>
+    <change issue="2389" PR="TODO" date="2026-04-27">
+      The rules for serializing atomic items have been revised and relaxed.
+    </change>
   </changes>
 
   <p>The Adaptive output method serializes the <termref diff="chg" at="2023-11-01" def="dt-input-tree"/> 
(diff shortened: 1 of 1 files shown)


michaelhkay:
- I don't see any need to escape ampersands within string literals as `&amp;`.

- In serializing QNames, "the prefix and the local name if the namespace is statically known, using the syntax prefix:local". That begs the question of what prefixes are "statically known" to the serializer. I suggest using the conventional prefix for namespaces that are pre-declared in XQuery/XSLT (and only those prefixes). I don't think we want any kind of context dependency in the serializer.

ChristianGruen:
> - I don't see any need to escape ampersands within string literals as `&amp;`.


With ampersands escaped, the output could be copied and pasted back into a query. Without, the syntax seems unclear to me: Why are quotes duplicated, and why is the string enclosed with quotes at all?

michaelhkay:
>With ampersands escaped, the output could be copied and pasted back into a query

But not into an XPath expression, unless it's to be hosted in XML. It seems inappropriate to use an XML escaping convention in a non-XML format.

ChristianGruen:
> But not into an XPath expression, unless it's to be hosted in XML. It seems inappropriate to use an XML escaping convention in a non-XML format.

Oh, I see. So if the output is only appropriate for XPath, but not for XQuery, maybe we should make it implementation-dependent as well.

ChristianGruen:
Revised:
* Explicit URIs have been added as suggested.
* Implementations are now allowed to omit the quotes on top level, or to output ampersands either literally or escaped.
* Implementation-dependent → -defined choices

As the prose tells, adaptive serialization was introduced for “purposes of debugging query results”. The rationale behind these changes is that I believe we need to give processors more freedom on what seems a sensible choice for the given environment. This is in analogy with the output of `fn:trace`, which completely depends on the implementation you use.

michaelhkay:
The proposal now says:

> Occurrences of the quote character within the value are doubled. It is [implementation-defined] whether ampersand characters are emitted literally or encoded as &amp;, so that the output can be pasted back into its host language.

Why are `<` characters not treated the same way as ampersands?

ChristianGruen:
> Why are `<` characters not treated the same way as ampersands?

In XQuery, only ampersands need to be escaped to get a valid string (`<` is valid).
In principle, we could extend the rule to arbitrary characters.

specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
```diff
@@ -4117,78 +4120,49 @@ item</xtermref> is serialized as follows:</p>
...
+<termref def="impdef">implementation-defined</termref> representation.</p></item>
 <item><p>An instance of <code>xs:NOTATION</code> is serialized 
 as a URI-qualified name (that is, in the form <code>Q{uri}local</code>).</p>
 </item>
 <item><p>An instance of <code>xs:QName</code> is serialized with a <code>#</code> character,
   followed by:</p>
   <ulist>
-    <item><p>the local name if the name is in no namespace, or</p></item>
+    <item><p>the local name if the name is in no namespace;</p></item>
+    <item><p>the conventional prefix and the local name (<code>prefix:local</code>) if the name is in one of the namespaces
```
Arithmeticus: This sentence seems to be two sentences squashed into one. And has the colon been forgotten? Perhaps some examples too? Further, the whole list of items would be better understood if the conditions appeared at the beginning of each item-sentence rather than at the end.

I propose:

...followed by:

- if the name is in no namespace, only the local name (e.g., `#example`);
- if the name is in one of the conventional namespaces A, B, C, D, E, the conventional prefix `fn`, `math`, `map`, `array`, or `xs`, respectively, followed by a colon and the local name (e.g., `#math:sin`);
- if the name is in an implementation-defined prefix, the predefined prefix, followed by a colon and the local name (e.g., `#my:example`);
- otherwise as a URI-qualified name (e.g., `#Q{http://example.com}example`).

Arithmeticus (commented):
Unclear prose; suggestion below.

ndw:

PR #2595, opened 2026-04-28
DM edits on XML Schema data types

Arithmeticus:
Edits here were motivated by several points that caused me confusion tonight:

- 4.1.2 and 4.1.5 had duplicate text, an indication that they belong next to each other.
- In 4.1.2 there were no pointers to the definitions of the datatypes discussed.
- The fifth bullet point in 4.1.2 is not a schema type (which is the topic of the unordered list). The material belongs outside the list.
- The fifth bullet point claims that XDM defines two data types, but it doesn't. And it doesn't mention another data type discussed in predefined data types.
- There is a long discussion of `xs:error` in the XPath specs, but the reader of this section has no clue on where `xs:error` is defined or explained.

Etc. I suspect my draft here still has problems, but I hope it's an improvement.

### DIFF
--- a/specifications/xpath-datamodel-40/src/xpath-datamodel.xml
+++ b/specifications/xpath-datamodel-40/src/xpath-datamodel.xml
@@ -702,39 +702,37 @@ a reference to a type definition in the Schema Component Model.
 <ulist>
 <item>
 <p>The 19 <phrase diff="add" at="2022-11-05">primitive atomic</phrase> types defined in
-<xspecref spec="XS2" ref="built-in-primitive-datatypes"/>
-of <bibref ref="xmlschema11-2"/>.</p>
+<xspecref spec="XS2" ref="built-in-primitive-datatypes"/> of <bibref ref="xmlschema-2"/>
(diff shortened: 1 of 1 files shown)


michaelhkay:
>The datatype xs:untyped is used as the type annotation of an element node that has not been validated, or that has been validated in skip mode. It is a classified as a complex type. The properties of xs:untyped are the same as the properties of xs:anyType except for the base type and name. The base type of xs:untyped is xs:anyType. No predefined types are derived from xs:untypedand no such derivations are allowed.

I suggest avoiding the term "datatype". Although it's used in XSD part 2, I think we established that it's a near-synonym of "type", and by convention is used only for simple types (whereas this is a complex type).

"It is a classified" -> "It is classified".

Arithmeticus:
Thanks, looks like those two issues are inherited from the previous version. Patch forthcoming.

ndw:
At [meeting 163](https://qt4cg.org/meeting/minutes/2026/05-05.html), the CG agreed to merge this PR.

Issue #2568, opened 2026-04-01
Error codes for fn:build-dateTime


martian-a:
Is `err:FODT0006` really necessary?  Why does it need to be distinct from the existing `err:FORG0001`, _Invalid value for cast/constructor_?  Other error codes make it possible to distinguish between a date/time formatting error (`err:FOFD1340`, `err:FOFD1350`), an invalid combination of date/time components (`err:FODT005`) and other date-related errors.  If the purpose of `err:FODT0006` is to be more targeted, dedicated to an invalid value for a component of a date/time, then shouldn't it be general to the same circumstances in all date/time constructor functions rather than specific to `fn:build-dateTime`?

Similarly, if the invalid value is for the timezone component, then I'd like to propose that `err:FODT0003` _invalid timezone value_  should be re-used instead of the more generic `err:FODT0006`.

michaelhkay:
Unfortunately we are not at all consistent in the granularity of error codes. Traditionally XSLT has been finer-grained than XQuery. I'm reluctant to spend time trying to improve this unless we can define some general principles; it's a matter of taste and we could go on polishing it for ever without really making the spec any better.

I'm a bit reluctant to use FORG0001 which is specifically about casts and constructor functions. I don't particularly want to change the description of FORG0001 to "Invalid value for casts, constructors, and a few other similar things".

PR #2599, opened 2026-04-29
2568 Error code changes for dateTime functions

martian-a:
* Deleting `err:FODT0006` 
* Updating the information about errors thrown by `fn:build-dateTime`
* No change to the text for `fn:parts-of-dateTime`, as it doesn't currently include a section on errors.  This is consistent with  the current expectation that neither `fn:parts-of-dateTime` nor the implicit constructor for `fn:dateTime-record` will check that the values supplied represent a valid date/time datatype.

As outlined in issue #2568, this represents a proposal to drop `err:FODT0006`, which is currently only expected to be thrown by `fn:build-dateTime`.  

Like the [functions for extracting components of dates and times](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#component-extraction-dateTime), `fn:build-dateTime` essentially orchestrates casting from one date/time-related datatype to another date/time-related datatype; determining which standard date/time constructor needs to be used and supplying the relevant arguments.  The errors thrown by date/time constructors are already defined: most commonly `err:FORG0001` but also `err:FODT0003` for an out-of-range timezone value.  My suggestion is that, as with the functions for extracting components of dates and times, there is no need for an error code specific to `fn:build-dateTime`.

`err:FODT0005` would remain unchanged and still be thrown by `fn:build-dateTime` if the combination of arguments supplied can't be mapped to a date/time datatype.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -11486,12 +11486,10 @@ return ($value - unix-dateTime()) div seconds(0.001)</eg>
             <item><p>day, and optional timezone (delivering an <code>xs:gDay</code>)</p></item>
             <item><p>hours, minutes, seconds, and optional timezone (delivering an <code>xs:time</code>)</p></item>
          </ulist>
-         <p>A dynamic error is raised <errorref class="DT" code="0006"/> if any of the
-            components in <code>$value</code> is outside the permitted range
-            (for example if the value of the <code>minutes</code> component is less than zero or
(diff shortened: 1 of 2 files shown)


ChristianGruen (approved):

michaelhkay:
Fix #2568

michaelhkay (approved):

martian-a:
I've just found a use-case for this error code :(  It's when the hour value is "24".  It's still valid to supply that value to a date/time constructor but not valid per the rules of `fn:build-dateTime`.  Assuming that inconsistency is to remain, I propose that this error code remain for this particular situation.

martian-a:
I'll raise another PR taking the 24 hour use-case into account.

PR #2601, opened 2026-04-29
2568 Error code changes for dateTime functions

### LINKED ISSUE #2568: Error codes for fn:build-dateTime

martian-a:
Replaces https://github.com/qt4cg/qtspecs/pull/2599 to fix #2568

Error code `FODT0006` is retained but reserved only for reporting when the value of `hours` is 24.  This is the only rule that applies to a valid combination of components, that diverges from the existing rules for the date/time constructors.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -11485,10 +11485,12 @@ return ($value - unix-dateTime()) div seconds(0.001)</eg>
             <item><p>day, and optional timezone (delivering an <code>xs:gDay</code>)</p></item>
             <item><p>hours, minutes, seconds, and optional timezone (delivering an <code>xs:time</code>)</p></item>
          </ulist>
-         <p><A dynamic error is raised <errorref class="RG" code="0001" /> (or <errorref class="DT" code="0003" /> in relation to a timezone value) if any of the components in <code>$value</code> is outside the permitted range
-            specified by the target date/time datatype.  For example if the value of the <code>minutes</code> component is less than zero or
-            greater than 59, or if the timezone is outside the range <code>-PT14H</code> to <code>PT14H</code>).  See <xspecref 
-               spec="DM31" ref="dates-and-times"/>.</p>
+      	<p>A dynamic error is raised <errorref class="DT" code="0003" /> if the timezone component of <code>$value</code> is outside the range <code>-PT14H</code> to <code>PT14H</code></p>
+      	<p>A dynamic error is raised <errorref class="DT" code="0006" /> if the hours component of <code>$value</code> is <code>24</code></p>
+         <p><A dynamic error is raised <errorref class="RG" code="0001" /> if any other component of <code>$value</code> is outside the permitted range
+            specified by the target date/time datatype, for example: if <code>minutes</code> is less than zero or
+            greater than 59; or if <code>day</code> is 30 and <code>month</code> is 2.  See <xspecref 
+               spec="DM31" ref="dates-and-times".</p>
       </fos:errors>
       <fos:notes>
          <p>Components that are present in <code>$value</code> but with an empty value are treated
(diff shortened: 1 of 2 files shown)


ndw:
There are a couple of markup errors in `function-catalog.xml`:
``` diff
$ git diff
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: specifications/xpath-functions-40/src/function-catalog.xml
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ specifications/xpath-functions-40/src/function-catalog.xml:11491 @ return ($value - unix-dateTime()) div seconds(0.001)</eg>
         </ulist>
        <p>A dynamic error is raised <errorref class="DT" code="0003" /> if the timezone component of <code>$value</code> is outside the range <code>-PT14H</code> to <code>PT14H</code></p>
        <p>A dynamic error is raised <errorref class="DT" code="0006" /> if the hours component of <code>$value</code> is <code>24</code></p>
         <p><A dynamic error is raised <errorref class="RG" code="0001" /> if any other component of <code>$value</code> is outside the permitted range
         ***^***
         <p>A dynamic error is raised <errorref class="RG" code="0001" /> if any other component of <code>$value</code> is outside the permitted range
            specified by the target date/time datatype, for example: if <code>minutes</code> is less than zero or
            greater than 59; or if <code>day</code> is 30 and <code>month</code> is 2.  See <xspecref
               spec="DM31" ref="dates-and-times".</p>
               spec="DM31" ref="dates-and-times"/>.</p>
                                             ***^^***
      </fos:errors>
      <fos:notes>
         <p>Components that are present in <code>$value</code> but with an empty value are treated
```
You should be able to test the build before you push a commit with `./gradlew`.

martian-a:
Sorry.  Fixed now.

martian-a:
Response to discussion during QT4 CT Meeting 163 (2026-05-05): https://qt4cg.org/meeting/minutes/2026/05-05.html#pr-2601

I proposed FORG0001 because that's the code that Saxon currently uses when constructing a date/time and one or more the of the components is outside of the permitted range.  The description for that error code when it was first introduced in XPath F&O 2.0 was simply:

> Invalid value for cast/constructor

There are some tests in prod/CastExpr.xml from 2007 that suggest that this code was originally intended for use when a date component is out-of-range as well as when it's uninterpretable as a number, for example: https://github.com/qt4cg/qt4tests/blob/master/prod/CastExpr.xml#L9924:~:text=K%2DSeqExprCast%2D295

In XPath F&O 3.1 a fuller description was added that implies that this error code is only relevant when casting, although the original description was preserved as a title/reason.

> **err:FORG0001, Invalid value for cast/constructor.**
> A general-purpose error raised when casting, if a cast between two datatypes is allowed in principle, but the supplied value cannot be converted: for example when attempting to cast the string "nine" to an integer.

But the error code wasn't changed to something else for those existing cases that test out-of-range.  Possibly because strictly speaking the constructor is also casting from a string.  Has it been possible to construct date/time datatypes from anything but a string before now?

However, in the case of `24` being an invalid value for the hours component supplied to `fn:build-dateTime`, this is a deviation from the norm, as it's still a valid value for the same component when supplied to the other date/time constructors.  It struck me as an unusual circumstance that merits a little extra explanation to the user, who may be confused that the same value is accepted when they use an implicit constructor (or `fn:dateTime`) but not when they use `fn:build-dateTime`.

martian-a:
I've updated https://github.com/qt4cg/qt4tests/pull/371 to match the decision to use `FORG0001` in the case of `24` hours.  It already included the change for the other instances of `FODT0006`.

martian-a:
I've updated this PR to reflect the decision during QT4 CT Meeting 163 (2026-05-05) to replace all uses of `FODT0006` with `FORG0001`.

martian-a:
Was it the committee's intention that `FODT0003` should also be replaced by `FORG0001` in this context?

ndw:
To simplify review, could you please rebase it off of the current `master` branch and force-push the update? Thanks...

PR #2606, opened 2026-04-30
2605 file:copy, file:move: define behavior when target is inside source dir

### LINKED ISSUE #2605: `file:copy`, `file:move`: define behavior when target is inside source directory
The current spec for `file:copy` and `file:move` does not address the case where `$source` is a directory and `$target` is located inside it. A naive implementation of the recursive rules can enter an infinite loop or even corrupt the file system.

#### Example:

```xquery
file:copy("/data/foo", "/data/foo/bar")
file:move("/data/foo", "/data/foo/bar")
```

ChristianGruen:
Closes #2605


ndw:
At [meeting 163](https://qt4cg.org/meeting/minutes/2026/05-05.html), the CG agreed to merge this PR.

PR #2607, opened 2026-04-30
2314 fn:error: return type xs:error

### LINKED ISSUE #2314: `fn:error`: return type `xs:error` ?
The current return type of `fn:error` is `item()*`. I assume we should change it to `xs:error`?

This would also align with what the spec says in [3.2.10. The type xs:error](https://qt4cg.org/specifications/xquery-40/xquery-40.html#id-xs-error):

> The practical uses of `xs:error` as a sequence type are limited, but they do exist. For instance, an error-handling function that always raises a dynamic error never returns a value, so `xs:error` is a good choice for the return type of the function.

ChristianGruen:
Closes #2314

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -1345,9 +1345,6 @@
          <fos:change issue="898 1161" PR="1265" date="2024-07-02">
             <p>The constraints on the result of the function have been relaxed.</p>
          </fos:change>
-         <fos:change issue="2314" PR="2607" date="2026-04-30">
-            <p>The return type has been changed to <code>xs:error</code>.</p>
-         </fos:change>
       </fos:changes>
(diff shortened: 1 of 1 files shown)


michaelhkay:
The Notes for `fn:error` still say "Since the function never returns a value, the declared return type of item()* is a convenient fiction." This note needs to change.

ChristianGruen:
> The Notes for `fn:error` still say "Since the function never returns a value, the declared return type of item()* is a convenient fiction." This note needs to change.

True, I overlooked this one. Nice wording… Maybe I’ll simply drop this section in the notes?

ChristianGruen:
…resolved; I think the PR can be fixed.

ndw:
At [meeting 163](https://qt4cg.org/meeting/minutes/2026/05-05.html), the CG agreed to merge this PR without discussion.

PR #2608, opened 2026-04-30
2312 Removal of Map/Array Filter (somewhat sigh)

### LINKED ISSUE #2312: Map/Array Filter: too much syntactic sugar?
Adopted from https://github.com/qt4cg/qtspecs/issues/2195#issuecomment-3426412436, and raised because of new user feedback:

Now that we have dropped `map:pairs`, the map filter expression is the only remaining that represents map entries as `record(key, value)`.

For the sake of simplicity and usability, we could question if the alternative writings for both filtering maps and arrays are not sufficient:

```xquery
let $map := { 0: 1, 1: 2, 2: 4, 3: 8 }
let $filter := fn($k, $v) { $k = $v - 1 }
return (
  (: existing ways of filtering map entries :)
  map:filter($map, $filter),
  { for key $k value $v in $map where $filter($k, $v) return { $k: $v } },
  (: redundant? :)
  $map?[$filter(?key, ?value)]
)

let $array := [ (), 1, (2, 3), (4, 5, 6) ]
let $filter := fn($m) { count($m) > 1 }
return (
  (: existing ways of filtering array members :)
  array:filter($array, $filter),
  array:join(for member $m in $array where $filter($m) return [ $m ]),
  (: redundant? :)
  $array?[$filter(.)]
)
```

ChristianGruen:
Closes #2312

### DIFF
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -2005,7 +2005,6 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
       <g:ref name="DynamicFunctionCall"/>     
       <g:ref name="LookupExpr"/>
       <g:ref name="MethodCall"/>
-      <g:ref name="FilterExprAM"/>
     </g:choice> 
   </g:production>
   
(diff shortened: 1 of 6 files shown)


ndw:
At [meeting 163](https://qt4cg.org/meeting/minutes/2026/05-05.html), the CG agreed to merge this PR without discussion.

Issue #2632, opened 2026-05-15
XQuery 4.0: automated review


ChristianGruen:
I have reviewed  and revised the current XQuery 4.0 draft with the help of AI. To simplify the intellectual review of the PRs, I have split the fixes into multiple steps, for each of which I will create a PR:

1. Critical bugs
2. Broken examples
3. Typos, grammar
4. Logic, semantics
5. Stale 4.0 content
6. Cross-cutting consistency

Arithmeticus:
Great idea @ChristianGruen

ChristianGruen:
I would like us to have a short (5 min) discussion on whether we want to keep or drop the 6 PRs that reference the issue.

ChristianGruen:
Let’s close it for now; I think it’s not worth to resolve the conflicts of the PRs.

PR #2635, opened 2026-05-15
2632-3: typos and grammar (automatically generated & reviewed)

ChristianGruen:
expressions.xml:
* "Universal Resource Identifier" -> "Uniform" (per RFC 3986)
* "an unprefixed name used appearing" -> drop "used"
* "ten character" -> "ten characters" (plural)
* "These means" -> "This means"
* "partial applying" -> "partially applying"
* "is `$Z` arguments, is obtained" -> "is `$Z`, is obtained"
* "attribute(*, xs:integer)selects" -> add missing space
* "type type tests" -> drop doubled "type"
* "(//a)[1] selects the a element" -> add "first"
* "language design for querying" -> "designed"
* "Because the fn:jtree creates" -> "fn:jtree function creates"
* "wheras" -> "whereas"
* "an sequence" -> "a sequence"
* "This is the output of the above function :" -> drop space before colon (3x)
* "because it is has no namespace binding" -> drop "is"
* "an named item type" -> "a named"
* constructor exclusion list: drop the second "and" and align order with the matching list in the cast section
* "is it is interpreted" -> drop doubled "is it"
* "summarizes the differences between these two operators" -> add colon
* "static and dynamic functions calls" -> "function calls"
* "arithemetic" -> "arithmetic"
* "be the present" -> "be present"
* "may also occur XQuery" -> "may also occur in XQuery"
* "is is a static error" -> "is a"
* "following following expressions" -> drop doubled "following"
* "typed defined" -> "types defined"
* "union typed declared" -> "union type declared"
* "noticably" -> "noticeably"
* "its datum not within" -> "its datum is not within"
* `xs:positive-integer` -> `xs:positiveInteger` (example head and body)
* ChoiceItemType: "it if matches" -> "if it matches"
* "an enumerated set of xs:strings" -> "an enumerated set of xs:string values"
* "or if unprefixed, the / is interpreted" -> "or, if unprefixed, the name is interpreted"
* "A is an singleton enumeration type" -> "a singleton" (term starts with /s/)
* missing commas in function list and missing space in "import schemain"
* "(see FunctionDeclns.)" -> "(see FunctionDeclns)."
* "of the language. without" -> "of the language, without"
* stray '"' after </code> in IDREFS example

ebnf.xml:
* head "Lexical structure" -> "Lexical Structure"
* "does not quality" -> "qualify"

errors.xml:
* "(as defined by the eq operator.)" -> "operator)." (twice)

query-prolog.xml:
* "Implementions" -> "Implementations"
* "variable or functions names" -> "function names"
* "an unprefixed name used appearing" -> drop "used"
* "another variable in in-scope variables" -> "in the in-scope variables"
* "refered" -> "referred"
* "in which altitude entry may be absent" -> "the altitude entry"
* "and and all options" -> drop doubled "and"
* `"fixed",` -> `"fixed"` (drop stray comma, twice)
* "It is an static error" -> "a static error"
* "Static Base URI property" -> "the Static Base URI property"
* "(for example, schema-attribute(xml:space)," -> close the parenthesis
* "can not import" -> "cannot import" (three occurrences)
* "in / in this section" -> drop the doubled fragment
* "may provide mechanism" -> "may provide a mechanism"

back-matter.xml:
* "( (which adds constructor functions" -> single opening paren
* "can returned to" -> "can be returned to"
* "Comparisons ... lacking a timezone uses" -> "use"
* "in the following sections" -> add trailing period
* "if any.)" -> "if any)." (period outside parenthesis)

conformance.xml:
* "attributes node types" -> "attribute node types"
* "a XQuery expression" -> "an XQuery expression"

introduction.xml:
* "Javascript" -> "JavaScript"

Issue: #2632

### DIFF
--- a/specifications/xquery-40/src/back-matter.xml
+++ b/specifications/xquery-40/src/back-matter.xml
@@ -289,7 +289,7 @@ the value of the component for their subexpressions.</p></item>
  <specref ref="FunctionDeclns"/>, 
 <specref ref="id-module-import"/>,
 and
-<specref ref="id-schema-import"/> ( (which adds constructor functions for user-defined types)</td>
+<specref ref="id-schema-import"/> (which adds constructor functions for user-defined types)</td>
   <td>no</td>
   <td>Must be consistent with statically known function signatures</td>
 </tr>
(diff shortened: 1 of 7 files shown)


ChristianGruen:
I think it’s not worth to resolve the conflicts of these PRs.
Let’s use these PRs as a depository for future editorial cleanups.

PR #2649, opened 2026-05-20
2647 file:descendants: recursion, filtering

### LINKED ISSUE #2647: file:descendants: recursion, filtering
It is advisable for an implementation (even if this is not addressed in the spec) to evaluate `file:descendants` fully before returning results. Otherwiese, a file walker may become corrupt, for example when a directory is deleted while it is being traversed. On the other hand, this means that the filtering can only take place after the full result is available.

Other programming languages differentiate between filtering the directories to traverse and the results to return, and I think we should do the same. My proposal:

```xquery
file:descendants(
  $path     as xs:string
  $options  as record(
    recurse  as (fn(xs:string, xs:integer) as xs:boolean?)? := true(),
    filter   as (fn(xs:string, xs:integer) as xs:boolean?)? := true()
  )? := {}
) as xs:string*
```

With `recurse`, the directories to be traversed can be selected; with `filter`, the actual results can be reduced. The `xs:integer` value indicates the current position; this way, traversal can for example be aborted after the first *n* directories.

ChristianGruen:
Closes #2647

### DIFF
--- a/specifications/EXPath/file/src/function-catalog.xml
+++ b/specifications/EXPath/file/src/function-catalog.xml
@@ -1568,7 +1568,9 @@ file:write(
         directory. The order of the items in the resulting sequence is not defined.
         The references to the specified directory and its parent directory
         (<code>.</code> and <code>..</code>) are not returned.</p>
-      
+      <p>Each returned string is composed of the directory argument joined with the
+        entry name using the implementation-defined directory separator; paths that
+        refer to directories carry a trailing separator.</p>
     </fos:rules>
     <fos:errors>
       <p>
@@ -1604,6 +1606,7 @@ file:write(
     <fos:signatures>
       <fos:proto name="descendants" return-type="xs:string*">
         <fos:arg type="xs:string" name="path"/>
+        <fos:arg name="options" type="map(*)?" default="{}"/>
       </fos:proto>
     </fos:signatures>
     <fos:properties>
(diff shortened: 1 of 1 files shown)


ChristianGruen:
Will be revised. Suggestions:
* Remove the position argument from the callback functions.
* Add a level/depth option.

ChristianGruen:
Revised:
* The position argument was removed from the callback functions.
* A depth option was added.

Feel free to discuss it and accept/reject it in Prague!

@line-o I decided against adding the glob pattern, it seemed too overloaded to me. Instead, you can do things like:

```xquery
file:descendants('.', { 'filter': file:is-dir#1 })
file:descendants('.', { 'filter': matches(?, '\.log$', 'i') })
```

line-o:
@ChristianGruen I understand. The callback approach offers some possibilities where a "simple" globbing pattern would fail. Your type filter is a good example.
Maybe there is an opportunity here to offer a `file:path-matches($globs as xs:string+)` callback?

ChristianGruen:
> @ChristianGruen I understand. The callback approach offers some possibilities where a "simple" globbing pattern would fail. Your type filter is a good example. Maybe there is an opportunity here to offer a `file:path-matches($globs as xs:string+)` callback?

@line-o Sorry, I overlooked your comment. – I feel a bit uneasy about providing this function. As Norm stated, there may be different expectations on what the function is going to do. In practice we only use XPath regex functions, as they offer much more freedom than glob patterns. However, I would certainly be interested in seeing a proposal for the function… ;)

ChristianGruen:
Revised; ready for review.

Arithmeticus:
The current text reads "Each returned string is composed of the directory argument followed by". So if I input `file:children('..')` I should expect the return values of `../parent-folder-file1.txt` `../parent-folder-file2.txt` etc., right? And not have the double dots resolved to an absolute URI? Just checking because I'm getting inconsistent results currently with different processors (BaseX keeps the dots but Saxon appears to resolve them). Also I do not see a test for this behavior in the qt4tests/file/file.xml.

ChristianGruen:
> The current text reads "Each returned string is composed of the directory argument followed by". So if I input `file:children('..')` I should expect the return values of `../parent-folder-file1.txt` `../parent-folder-file2.txt` etc., right? And not have the double dots resolved to an absolute URI? 

Exactly.

ndw:

PR #2662, opened 2026-05-22
2659 bin:count-ones, bin:rotate, bin:is-bit-set, bin:sets-bits

### LINKED ISSUE #2659: Binary Module: `bin:count-ones` for counting bits (population)
The Binary Module comes with helpful bitwise operators like `bin:and`, but no way to count set bits (`bin:length` counts octets). A population count, as offered by Java’s `Long.bitCount` or SQL’s `BIT_COUNT`, seems like a simple addition to me. For example, it is needed to compute the Hamming distance(`bin:count-ones(bin:xor($bin1, $bin2))`).

Proposed signature:

```xquery
bin:count-ones(
  $value  as (xs:hexBinary | xs:base64Binary)
) as xs:integer
```

An equivalent XPath solution is doomed to be slow and more verbose. – I’ll be happy to do the PR.

ChristianGruen:
Closes #2659

I have added three other trivial functions for operations that are cumbersome and inefficient to formulate in XPath: `bin:rotate`, `bin:is-bit-set`, `bin:sets-bits`.

### DIFF
--- a/specifications/EXPath/binary/src/binary-functions.xml
+++ b/specifications/EXPath/binary/src/binary-functions.xml
@@ -906,8 +906,8 @@ Michael Sperberg-McQueen (1954–2024).</p>
             <div2 id="func-bin-set-bits">
                 <head><?function bin:set-bits?></head>
             </div2>
-            <div2 id="func-bin-count-ones">
-                <head><?function bin:count-ones?></head>
+            <div2 id="func-bin-count-bits-set">
+                <head><?function bin:count-bits-set?></head>
             </div2>
         </div1>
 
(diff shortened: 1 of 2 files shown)


michaelhkay:
I tried to find a simpler expansion for `bin:rotate` that simply splits the value into two parts using `bin:part`, but realised bin:part uses octet numbering not bit numbering. Perhaps the fact that the expansions are messy just demonstrates that the functions are needed.

ChristianGruen:
`bin:count-ones` was renamed to `bin:count-bits-set` (@ndw the PR can be merged)

ndw:
At [meeting 166](https://qt4cg.org/meeting/minutes/2026/05-26.html), the CG agreed to merge this PR with minor changes. (Thank you, @ChristianGruen )

Issue #2641, opened 2026-05-18
[Feature] Comments and parse-csv()


djbpitt:
There is no standardization of comments in csv, but the inclusion of comment lines in csv documents is sufficiently widespread that Python pandas (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) and R (https://stat.ethz.ch/R-manual/R-devel/library/utils/html/read.table.html) cater for it. These are large communities that have substantial engagement over a long time with data in csv files.

Is it realistic to request a similar addition to the options parameter of XPath `parse-csv()`? The user can, to be sure, filter the lines before passing them to `parse-csv()`, but if it would not create an unacceptable burden for XPath 4.0 implementers to support it, wouldn't it be nicer to have it built in?

PR #2698, opened 2026-06-19
2641 Support comments in csv

djbpitt:
Modified `parse-csv()` and `csv-to-arrays()` to add support for comments.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -28863,13 +28863,12 @@ element-to-map-plan((<a><b/><b/></a>, <a><b/><c/></a>))
 
          <p>If <code>$value</code> is the empty sequence, the function returns the empty sequence.</p>
 
-         <p>The input supplied in <code>$value</code> is CSV data, as defined
-            in <bibref ref="rfc4180"/>. The function
-            first parses the input using <function>fn:csv-to-arrays</function>, and then
-            further processes the result. The initial parsing is exactly as defined for
-            <function>fn:csv-to-arrays</function>, and can be controlled using the same options.
-            Additional options are available to control the way in which header information and
-            column names are handled.</p>
+        <p>The input supplied in <code>$value</code> is CSV data, as defined in <bibref
+            ref="rfc4180"/>, extended to permit comments. The function first parses the input
+          using <function>fn:csv-to-arrays</function>, and then further processes the result. The
+          initial parsing is exactly as defined for <function>fn:csv-to-arrays</function>, and can
+          be controlled using the same options. Additional options are available to control the
+          way in which header information and column names are handled.</p>
 
          <p>If the input is the a zero-length string, the function
                returns a <code>parsed-csv-structure-record</code> whose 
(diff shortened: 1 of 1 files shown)


michaelhkay (commented):
The specification should avoid the term "line" and instead be defined in terms of row delimiters. (Perhaps "If the character following a row delimiter is a comment character, then all characters up to and including the following row delimiter (or the end of input if that occurs first) are ignored"). It needs to be clear that a comment character is not recognized if it occurs (following a row delimiter) within a quoted field. It needs to be stated that the comment character mustn't conflict with other special characters. It needs to be clear what happens if the file consists entirely of comments.

djbpitt:
Thank you, @michaelhkay , for the quick review. This is my first spec PR (ever), and I appreciate the guidance.

I had already mentioned that the comment character must not conflict with other special characters by incorporating it into the error conditions for the two functions (`parse-csv()` and `csv-to-arrays()`). I've now also updated the description of both of those functions to attempt to implement your suggestions, but I'd be grateful for a follow-up if it looks as if I've missed or misunderstood anything.

On a semi-related topic: The checks do not all pass. Should I be concerned about that?

michaelhkay:
>On a semi-related topic: The checks do not all pass. Should I be concerned about that?

No, it's happening a lot. Some kind of problem at GitHub. It usually works on a subsequent attempt.

So long as you run `./gradlew` to build the specs before pushing your commit, all should be well.

michaelhkay:
Fix #2641

michaelhkay:
> A file that consists entirely of comments is treated as if it consisted entirely of empty rows separated by row delimiters.

I think it should be equivalent to an empty file.

>A blank row is deemed to exist when a row begins with a specified comment-character or when row delimiter immediately follows either another row delimiter

Similarly: a comment row should be ignored completely, it shouldn't be treated the same as a blank row (which results in an empty array in the output). Hence my suggestion, which you incorporated, that everything up to AND INCLUDING the following row delimiter is ignored.

djbpitt:
@michaelhkay I'm revising in response to your comments, and before I update the PR, I'd be grateful for your guidance about the following.

>If the character immediately following a row delimiter is a comment character, then all characters up to and including the following row delimiter (or the end of input if that occurs first) are ignored.

This wording appears not το account for the first line of a file that begins with a comment character, that is, a file where the first row is what we've called a comment row. I was about to add that condition to the wording above when I hesitated because an initial comment row might require adjusting our wording about the optional header row. The description of `parse-csv()` describes the `header` option as:

> Determines whether the first row of the CSV should be treated as a list of column names, or whether column names are being supplied by the caller. 

I could change this from "first row" to "first parsed row", since we write elsewhere (or will write elsewhere once I implement your latest remark about that) that a comment row (row that begins with a comment character) is not parsed. Is that modification appropriate, or are there complications about the header row I've failed to notice? Similarly:

> The comment character is not recognized (that is, is treated as a regular textual character) if it occurs (following a row delimiter) within a quoted field.

I'm hesitating now over the "following a row delimiter", since not only could the first row (not following a row delimiter) be entirely a comment, but a comment might occur after data in a first row (whether it is a header row or not). I copied the parenthesized phrase in question from your suggestion, above, so I hesitate to remove it now in case I've misunderstood the original rationale.

If the thoughts above sound constructive, I'll be happy to update the PR accordingly. Thank you again for your review comments.

ChristianGruen:
@djbpitt I’m afraid the latest PRs have led to numerous conflicts, mostly because of #2746 (https://qt4cg.org/pr/2746/xpath-functions-40/autodiff.html). Would you be willing to revise the PR?

djbpitt:
@ChristianGruen I think it's simplest just to reject my PR. I can't rebase my PR branch on upstream/master without conflicts that need to be resolved manually, and redoing the PR from scratch is likely to be less prone to confusion and error. I regret that I overestimated my ability to prepare a useable PR; apologies for the faff.

michaelhkay:
I'll take this on board.

PR #2707, opened 2026-06-24
1962 fn:map-to-element

### LINKED ISSUE #1962: `fn:map-to-element`
As this feature request has been reported back to us more than once, I want to raise the question here if we want to introduce a function that inverts the result of `fn:map-to-element` back to an XML representation – provided that a conversion plan exists.

Many results would certainly be lossy, but as the plan is now available separately, it would be possible to roundtrip a lot of data with regular structures, without having to write custom conversion code.

ChristianGruen:
It repeatedly bothered me that we have no reverse counterpart for `fn:element-to-map`, so I finally pulled myself together and wrote a PR for `fn:map-to-element`.

All feedback is much appreciated. While I’m quite confident about the defined rules, I would mostly be interested in grammatical/stylistic advise.

Closes #1962

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -29061,7 +29061,7 @@ element-to-map-plan((<a><b/><b/></a>, <a><b/><c/></a>))
       <fos:see-also prefix="fn" name="element-to-map"/>
       <fos:see-also prefix="fn" name="element-to-map-plan"/>
       <fos:changes>
-         <fos:change issue="1692" PR="" date="2026-06-24"><p>New in 4.0.</p></fos:change>
+         <fos:change issue="1962" PR="2707" date="2026-06-24"><p>New in 4.0.</p></fos:change>
       </fos:changes>
    </fos:function>
 
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -28968,6 +28968,103 @@ element-to-map-plan((<a><b/><b/></a>, <a><b/><c/></a>))
       </fos:changes>
    </fos:function>
 
+   <fos:function name="map-to-element" prefix="fn">
+      <fos:signatures>
+         <fos:proto name="map-to-element" return-type="element()?">
+            <fos:arg name="map" type="map(*)?" usage="inspection"/>
+            <fos:arg name="options" type="map(*)?" usage="inspection" default="{}" note="default-on-empty"/>
+         </fos:proto>
+      </fos:signatures>
+      <fos:properties>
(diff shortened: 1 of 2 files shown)


michaelhkay (commented):
I think my main reservation here is about handling of maps that could not have been generated by element-to-map(). This requires an understanding of use cases. Ideally I think it would be better to provide a function that can handle any map (but that works best if it was generated by element-to-map).

If I understand correctly, an array of strings will be converted to an element with multiple adjacent text nodes. That doesn't feel right, even though it might be the closest we can achieve to round-tripping behaviour.

ChristianGruen:
Thanks for the first feedback. 

> I think my main reservation here is about handling of maps that could not have been generated by element-to-map(). This requires an understanding of use cases. 

In an existing use case we transfer client JSON data to a server and vice versa. The RESTXQ server uses our custom `json:parse` and `json:parse` functions to convert the data to an XML representation, and the XML data is validated and compactified using a lightweight XML model. With the proposed `fn:map-to-element` function, it seems straightforward to replace this 2-phase workflow by a single step.

As already stated, I would prefer a strict mode to be the default, in both directions. One reason is that it is always easier to start off with restrictions and get more liberal later on. My suggestion would thus be to reject any data that cannot be represented via a plan, and to introduce a `lax` option in a second step.

> If I understand correctly, an array of strings will be converted to an element with multiple adjacent text nodes. That doesn't feel right, even though it might be the closest we can achieve to round-tripping behaviour.

Good point; I wasn’t happy about it anyway. I will treat non-single-member arrays as an error, a commit is on the way.

michaelhkay:
Note that supplying the same attribute marker is not sufficient to ensure that attributes will round-trip as attributes; this will only happen if the attribute marker is sufficiently distinctive (and not, for example, set to "").

michaelhkay:
> "In the absence of a plan, the layout is inferred from the structure of the map value, as the inverse of the rules in [14.6.2 Creating a conversion plan]"

This seems very hand-waving. Running an algorithm backwards is not especially well defined.

In 14.6.12.1

>A value V appearing as element content is interpreted according to its kind:

What does "appearing as element content" mean here? Surely we're talking about values appearing in map content?

ChristianGruen:
I’ve revised the wording with my last commit.

ndw:
At [meeting 175](https://qt4cg.org/meeting/minutes/2026/08-11.html), the CG agreed to merge this PR.

ChristianGruen:
@johnlumley The answer to your question (why “dates”?) couldn’t have been more obvious. The string was simply outdated, it should have been “years”:

```xquery
map-to-element(
  { "years": [ "2023", "2023" ] },
  { "plan": { "years": { "layout": "list", "child": "year" } } }
)
```

It will be fixed with the next editorial batch update (#2827).

PR #2710, opened 2026-06-25
2702 Dynamic selectors: focus-dependency

### LINKED ISSUE #2702: Dynamic selectors: focus-dependency
…probably editorial, but I decided to give it a separate issue:

The current spec says:

> It is permitted, though not generally useful, for the enclosed expression to be focus-dependent, in which case its value will vary from one selected node to another. For example, `child::{@name}` selects those child elements whose local-name is equal to the value of their `@name` attribute.

This is not reflected by the equivalent code yet:

```xquery
let $A as (xs:QName | xs:string)* := E 
return AXIS::*[some $a in $A, $n in (node-name(.), local-name(.)) 
               satisfies atomic-equal($a, $n)]
```

ChristianGruen:
Closes #2702


michaelhkay:
Looks like we overlapped here.

ChristianGruen:
Closed; #2712 has better chances to survive.

PR #2713, opened 2026-06-25
2257 Record declarations without namespace

### LINKED ISSUE #2257: Record declarations without namespace
Thanks to #2200, user-defined functions can now be declared without namespaces:

```xquery
declare function yes() {};
yes()
```

We should extend this principle to record declarations:

```xquery
declare record coord(x, y);
coord(1, 3)
```

A simple adjustment might be to adapt the rules in [5.20.2 Constructor Functions for Named Record Types](https://qt4cg.org/specifications/xquery-40/xquery-40.html#named-records-as-functions), by changing…

> The function name is the QName of the named record declaration, expanded using the [default type namespace rule](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-default-type-namespace-rule). The resulting QName must be the same as the module namespace if the declaration appears in a library module, and in any event, it must be in some namespace.

…to “using the [default function namespace rule](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-default-function-namespace-rule)”, and dropping the last sentence.

Another solution would be include the default element namespace when searching for functions with unprefixed function names.

ChristianGruen:
This solution aims to provide a solution for the vast majority of use cases where the default function and element namespaces remain untouched, and where people write simple one-file scripts. I think I have found a good compromise:

* Unprefixed records are allowed: `declare record coord(x, y); coord(1, 3)`
* Name keeps the default-type-namespace rule; drops "must be in some namespace."
* No default element namespace → record and constructor in no namespace; unprefixed call resolves via no-namespace-first
* Default element namespace → call constructor prefixed

Closes #2257

### DIFF
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -4937,8 +4937,8 @@ types</term> (such as <code>xs:integer</code>) and function types
                   <p>Some consequences of these rules may not be immediately apparent.</p>
                   <p>Suppose that an XQuery query contains the declarations:</p>
                   <eg>
-declare type my:color := enum("red", "green", "orange");
-declare type my:fruit := enum("apple", "orange", "banana");
+declare type my:color as enum("red", "green", "orange");
+declare type my:fruit as enum("apple", "orange", "banana");
 declare variable $orange-color as my:color := "orange";
 declare variable $orange-fruit as my:fruit := "orange";
                   </eg>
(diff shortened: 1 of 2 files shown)


ChristianGruen:
The change also includes type declarations, which is why I have added an example (and fixed one that was wrong in the current `main` branch).

ndw:
At [meeting 172](https://qt4cg.org/meeting/minutes/2026/07-14.html), the CG agreed to merge this PR.

PR #2714, opened 2026-06-25
2219 Generalize method calls to sequences

### LINKED ISSUE #2219: Generalize method calls to sequences
Adopted from https://github.com/qt4cg/qtspecs/pull/2206#issuecomment-3333242827:

We should further simplify the equivalent code for `M =?> N(X, Y, ...)` to:

```xquery
for $map as map(*) in M
return $map?N($map, X, Y, ...)
```

This would align the syntax with the current generalization of dynamic function calls to sequences, and it would also make the semantics more similar to the sequence arrow operator, which now allows:

```xquery
123 => (count#1, sum#1)()
```

ChristianGruen:
Closes #2219


ndw:

PR #2717, opened 2026-06-25
2660 fn:matching-segments: named capture groups

### LINKED ISSUE #2660: `fn:matching-segments`: named capture groups
Once again I’m flooding the issue list: Named capture groups would be a great extension in the interplay with `fn:matching-segments`. Instead of `?groups?1?group`, we could write `?groups?city?group`.

I’ve marked the issue with `nice-to-have`. Let’s see if I can create a PR for it after Prague.

ChristianGruen:
If someone has time for a review: I definitely need feedback on the XSLT part, or someone who is willing to complete my edits.

Closes #2660

### DIFF
--- a/specifications/xslt-40/src/function-catalog.xml
+++ b/specifications/xslt-40/src/function-catalog.xml
@@ -837,6 +837,12 @@
                <p>The capturing subexpression exists, and matched a zero-length substring of
                   the input string.</p>
             </item>
+            <item>
+               <p>The capturing subexpression exists, but is a
+                  <xtermref spec="FO40" ref="dt-named-capturing-group">named capturing group</xtermref>,
+                  in which case it is accessed by its name, and supplying its number returns a
+                  zero-length string.</p>
+            </item>
          </olist>
 
          <p>The set of captured groups is a context variable with dynamic scope. It is initially
--- a/specifications/xslt-40/src/function-catalog.xml
+++ b/specifications/xslt-40/src/function-catalog.xml
@@ -700,7 +700,7 @@
    
    <fos:function name="regex-groups">
       <fos:signatures>
-         <fos:proto name="regex-groups" return-type="map(xs:integer, record(value as xs:string, position as xs:integer))"/>
+         <fos:proto name="regex-groups" return-type="map((xs:integer | xs:string), record(value as xs:string, position as xs:integer))"/>
       </fos:signatures>
       <fos:properties>
          <fos:property>deterministic</fos:property>
@@ -727,7 +727,12 @@
                is identified by a record containing both the captured string itself (as <code>value</code>)
             and the start position (1-based) of the
                captured group within the input string (as <code>position</code>).</p>
-         
+
+         <p>If a capturing subexpression is a
+            <xtermref spec="FO40" ref="dt-named-capturing-group">named capturing group</xtermref>,
+            its entry is keyed by the name of the group (as an <code>xs:string</code>) rather than
+            by its group number (as an <code>xs:integer</code>).</p>
+
                <note><p>This means that the string value of group zero is
             the same as the value of <code>.</code> (dot).</p></note>
          
@@ -743,6 +748,11 @@
             <item>
                <p>The <var>N</var>th capturing subexpression exists, but did not match any part of the input
                   string.</p>
+            </item>
+            <item>
+               <p>The <var>N</var>th capturing subexpression is a
+                  <xtermref spec="FO40" ref="dt-named-capturing-group">named capturing group</xtermref>,
+                  in which case its entry is keyed by name rather than by number.</p>
             </item>
          </olist>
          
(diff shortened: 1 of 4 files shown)


michaelhkay:
Excellent.

I think we also need to be able to refer to captured groups by name in `fn:replace()`. In a replacement string, I suggest the syntax `$<group-name>`. In a replacement function (which is new in 4.0), I suggest we make the second argument a map of groups rather than a sequence of groups.

I'm not convinced that where groups have a name, we still need to make the value available by number, but I guess it does no harm.

ChristianGruen:
> I'm not convinced that where groups have a name, we still need to make the value available by number

I’ll better drop it again; too much redundancy.

ChristianGruen:
Updated: The new version returns either the positional or the named group

Arithmeticus:
If we drop integer-based entries for named groups, users need to be warned that there may be gaps in the integer-based sequence of map entries. And I think some users will find such gaps an annoyance or nuisance.

Also, just checking: if a regex group is given the name of an integer, is there the potential for ambiguity when dereferencing it?

ChristianGruen:
> If we drop integer-based entries for named groups, users need to be warned that there may be gaps in the integer-based sequence of map entries. And I think some users will find such gaps an annoyance or nuisance.

A helpful observation; I have added a note. With regard to gaps, they also occur when a group is not part of the match (for example, an alternative branch that was not taken).

> Also, just checking: if a regex group is given the name of an integer, is there the potential for ambiguity when dereferencing it?

We are safe here. With the current grammar rules, the group name cannot be an integer. Even if we allowed numbers, we would end up with a string and an integer key.

ChristianGruen:
Feedback to be incorporated:
* Extend the character range of named groups (if common in other programming languages)
* Possibly: Include both positions and names in the result (dropped, could be readded)

ChristianGruen:
Revised:
* Positional access is already easily possibly by using predicates. I have added an example.
* I decided against extending the character range. As Michael insinuated, an extension would make it more difficult to reuse existing implementations. This includes the underscore `_`, which is not supported by Java by default, for example.

specifications/xslt-40/src/xslt.xml
```diff
@@ -26935,6 +26936,24 @@ the same group, and the-->
...
+    regex="(?<day>[0-9]{{1,2}})\s(?<month>[A-Z][a-z]+)\s(?<year>[0-9]{{4}})">
+    <xsl:matching-substring>
+        <xsl:number value="regex-group('year')" format="0001"/>
+        <xsl:text>-</xsl:text>
+        <xsl:number value="index-of($months, regex-group('month'))" format="01"/>
+        <xsl:text>-</xsl:text>
+        <xsl:number value="regex-group('day')" format="01"/>
+    </xsl:matching-substring>
+</xsl:analyze-string>]]></eg>
+               <p>A named capturing group is also numbered in the usual way, so the captured groups
```
Arithmeticus: @ChristianGruen I would expect that this paragraph is incorrect in light of revisions to XPath `[fn:matching-segment-record](https://qt4cg.org/pr/2717/xpath-functions-40/autodiff.html#matching-segment-record)`.

specifications/xslt-40/src/function-catalog.xml
```diff
@@ -727,7 +727,12 @@
                is identified by a record containing both the captured string itself (as <code>value</code>)
             and the start position (1-based) of the
                captured group within the input string (as <code>position</code>).</p>
-         
+
+         <p>If a capturing subexpression is a
```
Arithmeticus, line 731: This paragraph needs to be duplicated and added to the companion function, `fn:regex-group()`

Arithmeticus (commented):
Two comments attached.

specifications/xslt-40/src/xslt.xml
ChristianGruen: Thanks, I missed that one; fixed.

specifications/xslt-40/src/function-catalog.xml
ChristianGruen: XSLT is my weak part, but as I understand, `regex-group(G)` is defined as `regex-groups()?G?value otherwise ""`, so it should inherits the named-group behaviour automatically.

Arithmeticus: That is true, but `regex-group` is the older, more familiar function, and it will be the one most used of the 4.0 specs, so the logic for named groups needs some attention there, in my opinion.

ChristianGruen: Agreed; I have added a note.

michaelhkay:
§6.1.5 says: "For example, in the result of [fn:matching-segments], a captured group that has a name is made available both under its group number and under its name."

I think that is no longer true.

michaelhkay:
§6.3.5 says "the entries of the map retain the order of the capturing groups, so a group can still be selected by position. The following expression returns the second capturing group:" I'm not sure that's true in general. The spec says that "When, in the course of evaluating a regular expression, a particular [segment] of the input matches a capturing subexpression, that segment becomes available as a captured group." Given the regex `a(b)c|d(e)f` and the input string `abc`, I think this rule implies that there is no captured group #2. The only possible segment that could match `(e)` would have the string value "e", and there is no such segment.

This is why the spec for regex-groups() says: The result map contains no entry for a number N if there is no captured group with the number N. This can occur for a number of reasons:... (3) The Nth capturing subexpression exists, but did not match any part of the input string.

ChristianGruen:
Thanks; revised. I tried to shorten the prose in both cases.

ndw:

Please resolve the merge conflict and merge it at your convenience.

PR #2719, opened 2026-06-25
1234 Serialization Parameters: Indentation, Whitespace, Newlines

### LINKED ISSUE #1234: Seralization Parameters: Indentation, Whitespace, Newlines
Summary of #358 and #101:

Parameter | Description | Values
--- | --- | ---
`indent-unit` | Character sequence to use for indentation. | Pattern: `(\\t\| +)`<br/>Examples: (`"\t"`, `" "`, `"    "`)
`indent-attributes` | Indent multiple attributes<br/>(similar: [HTML Tidy’s configuration option](http://api.html-tidy.org/tidy/quickref_5.0.0.html#indent-attributes)).<br/>Should be used with spaces as indentation characters. | `"yes"`, `"no"`
`line-ending` | Newline character. | `"\r\n"`, `"\n"`, `"\r"`, …

If no parameter is specified, the default of the implementation is used.

ChristianGruen:
Closes #1234


specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
```diff
@@ -1981,6 +2011,28 @@ be based on whitespace stripped from either the source document or the
...
+
+<div3 id="XML_INDENT-UNIT" diff="add" at="2026-06-25">
+  <head>XML Output Method: the <code>indent-unit</code> and <code>indent-attributes</code> Parameters</head>
+
+  <p>The <code>indent-unit</code> and <code>indent-attributes</code> parameters refine the
+  indentation performed when the <code>indent</code> parameter has the value <code>true</code>;
+  they have no effect when <code>indent</code> is <code>false</code>.</p>
+
+  <p>If the <code>indent-unit</code> parameter is specified, the
+  <termref def="serializer">serializer</termref> <rfc2119>MUST</rfc2119> use its value as the string
```
Arithmeticus: As currently written the closest referent of "its" is "serializer", but "parameter" is intended. Better to recast with the passive voice: "If the indent-unit parameter is specified, its value MUST be used by the serializer as its value for a single level of indentation."

specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
```diff
@@ -777,6 +777,24 @@ This parameter <rfc2119>MAY</rfc2119> be absent.</td>
...
+  separate lines. It has no effect unless <code>indent</code> is <code>true</code>.</td>
+</tr>
+
+<tr diff="add" at="2026-06-25"><td rowspan="1" colspan="1" valign="top">
+  <code>indent-unit</code></td><td rowspan="1" colspan="1">A string of whitespace characters
+  matching the pattern <code>(\t+| +)</code>, that is, one or more tab characters, or one or more
+  space characters. It defines the string used for a single level of indentation when the
+  <code>indent</code> parameter has the value <code>true</code>. This parameter
+  <rfc2119>MAY</rfc2119> be absent, in which case the unit of indentation is
+  <termref def="impdef">implementation-defined</termref>.</td>
```
Arithmeticus: Do you need to specify here what error code is raised when the value does not conform to the regex?

specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
```diff
@@ -1981,6 +2011,28 @@ be based on whitespace stripped from either the source document or the
...
+  <p>If the <code>indent-unit</code> parameter is specified, the
+  <termref def="serializer">serializer</termref> <rfc2119>MUST</rfc2119> use its value as the string
+  of whitespace characters representing a single level of indentation. The whitespace added before an
+  indented item is the value of <code>indent-unit</code> repeated once for each level of
+  element nesting that encloses the item. If the parameter is absent, the unit of indentation is
+  <termref def="impdef">implementation-defined</termref>.</p>
+
+  <p>If the <code>indent-attributes</code> parameter has the value <code>true</code>, the
+  <termref def="serializer">serializer</termref> <rfc2119>MAY</rfc2119> output each attribute of an
+  element on a separate line, indented relative to the start tag of the element. If it has the value
```
Arithmeticus: Does "indented relative" mean "the same level of indentation as" or "one level of indentation greater than"? I think users would expect the latter?

Does the formulated definition provide for an implementation-dependent third option, one where an attribute is put on a new line only when it exceeds a user-supplied parameter to the implementation of the maximum width of an unconstrained line, i.e., in the context of a an element marked as `xs:untyped` or `xs:anyType`. This is the default Oxygen behavior, where attributes are grouped on the same line, and new lines for attributes are made reluctantly.

Arithmeticus (commented):
Excellent feature...I look forward to seeing it incorporated into the specs.

specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
ChristianGruen: Good point. I tend to simply replace `its` by `that`.

specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
ChristianGruen: I believe this is already achieved by the following global rule:

> It is a serialization error [err:SEPM0016] if a parameter value is invalid for the given parameter.

specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
ChristianGruen: > Does "indented relative" mean "the same level of indentation as" or "one level of indentation greater than"? I think users would expect the latter?

I will simplify the rule, and make the amount of indentiation implementation-defined.

> Does the formulated definition provide for an implementation-dependent third option

It doesn’t, but I would say the current wording already provides for it: the default (false) corresponds to Oxygen’s behavior: attributes SHOULD be kept on the same line, but a configured maximum width may be the reason to insert new lines.

We could also make the option non-binary; more suggestions are welcome.

ChristianGruen:
Feedback to be incorporated:
* Enable the parameters for other serialization methods
* (Possibly) allow special characters (`\t`, `\r`, `\r\n`, `\n`) instead of strings
* Add parameters for `fn:serialize`, XSLT, …

michaelhkay:
Any escaping with `\` or `&` is going to require double-escaping in some contexts, and is going to show as a blank canvas in others. My inclination is to go with S=space, T=tab, N=newline, R=carriage return, so `indent-unit="SSSS"` means indent with four spaces, `indent-unit="T"` means indent with one tab, `newline="RN"` means use CRLF for newlines.

michaelhkay:
I wonder whether a composite attribute with microsyntax might be appropriate: `indent="SSSS+A" means indent with four spaces as the indentation unit, and indent attributes.

ChristianGruen:
> I wonder whether a composite attribute with microsyntax might be appropriate: `indent="SSSS+A" means indent with four spaces as the indentation unit, and indent attributes.

I consistently fail to memorize the idiosyncratic `format-integer(65535, '16^xxxx')` syntax ;·) …which is why I would prefer standard strings.

ChristianGruen:
Revised:
* The new parameters were added to the other serialization methods (I had already implemented it, just forgotten to complete the spec).
* The parameters were incorporated in various other documents: XQFO, XQuery, XSLT, RNC (…let’s hope we won’t change too much again)
* `indent-unit`: Values must now conform to the regex `(\t|\\t)*| *` (@johnlumley this includes empty strings)
* `line-ending`: Values must now conform to `\r\n|\r|\n|\\r\\n|\\r|\\n`

The extended regex patterns allow you to specify either backslashed characters (as they are now allowed for `fn:char` or the regex functions) or literal characters (including escaped characters à la `&#x;`). This solution should be unambiguous and easy to use: If a `\t` character sequence is unescaped by an intermediate step, the resulting 0x09 character is still processable.

specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
```diff
@@ -1981,6 +2009,54 @@ be based on whitespace stripped from either the source document or the
...
+
+<div3 id="XML_INDENT-ATTRIBUTES" diff="add" at="2026-06-25">
+  <head>XML Output Method: the <code>indent-attributes</code> Parameter</head>
+
+  <p>The <code>indent-attributes</code> parameter refines the indentation performed when the
+  <code>indent</code> parameter has the value <code>true</code>; it has no effect when
+  <code>indent</code> is <code>false</code>.</p>
+
+  <p>If the <code>indent-attributes</code> parameter has the value <code>true</code>, the
+  <termref def="serializer">serializer</termref> <rfc2119>MAY</rfc2119> output each attribute of an
```
Arithmeticus:
```suggestion
  <termref def="serializer">serializer</termref> <rfc2119>MAY</rfc2119> output any attribute of an
```
To avoid the implication that if indentation is applied it must be applied to every single attribute.

specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
```diff
@@ -1981,6 +2009,54 @@ be based on whitespace stripped from either the source document or the
...
+  element on a separate line; the amount of indentation applied to such lines is
+  <termref def="impdef">implementation-defined</termref>. If it has the value
+  <code>false</code>, the <termref def="serializer">serializer</termref> <rfc2119>SHOULD</rfc2119>
+  output the attributes of an element on the same line as the element name.</p>
```
Arithmeticus, line 2045: In Oxygen, attribute indentation may also cause the `/>` at the end of an empty element to drop to a new line, on its own. Do we want to accommodate that in the specs?

Arithmeticus (commented):
Suggestion and question.

specifications/xslt-xquery-serialization-40/src/xslt-xquery-serialization.xml
ChristianGruen: Thanks; added.

michaelhkay:
I think that allowing spaces, tabs, and newlines as literal characters representing themselves is hopelessly error prone, partly because it's not distinguishable to the human reader, and partly because serialization parameter values can be written as XML attributes, which are subject to attribute value normalization. Backslashes are less error prone (though when the spec says that the value must match `\\t`, the spec gets confusing). I would suggest making it simpler: indent-unit="SSS" for three spaces, or indent-unit="T" for one tab, and use `newline="NL | CR | CRLF"`

ChristianGruen:
If someone wants two spaces as indentation, I believe that `"  "` is exactly what (s)he would type first. I would advocate against custom syntax: It’s so much easier to use syntax that is also used somewhere else.

We could additionally allow `"\s"`, but of course it’s not identical to a space character.

ChristianGruen:
If space characters are too invisible, they can also be written as `"&#32;"`. Or we could support both `" "` and `"S"`.

Allowing space characters is something that we already have for `item-separator`.

ChristianGruen:
Conflicts resolved in this PR (fortunately the last, hopefully correct).
I have adopted the new presentation style for headers.

michaelhkay:
The table of serialization parameters in section 3 is becoming a little incoherent: for most parameters we only say what the possible values are, but for some we also say what the purpose of the parameter is.

michaelhkay:
"line-ending":

>If the parameter is specified, every line ending that the serializer would otherwise emit, including line endings that occur in the content being serialized, must be replaced by this string.

I think it would be better to say:

>If the parameter is specified, every newline character (U+000A) that the serializer would otherwise emit, including newline characters that occur in the content being serialized, must be replaced by this string.

For example, if the line ending parameter takes the value "\n", this shouldn't affect the handling of a CR within the content.

What happens if indent-spaces or line-ending is set to a value such as "&" that causes ill-formed XML to be output?

If line-ending changes the content of elements or attributes, when does this happen in terms of the "phases of serialization" in section 4? How does it interact with, for example, character map expansion and unicode normalization? 

I'm really not at all sure that changing the representation of newlines within the content is a good idea, unless we constrain the representation to be one that an XML parser will normalize back to a regular newline.

I think it would be wise to restrict `indent-spaces` to be whitespace, and to restrict line-ending to a value that XML recognises as a line ending.

ChristianGruen:
> If the parameter is specified, every newline character (U+000A) that the serializer would otherwise emit, including newline characters that occur in the content being serialized, must be replaced by this string.

…revised.

> What happens if indent-spaces or line-ending is set to a value such as "&" that causes ill-formed XML to be output?

There’s currently no way to do so, unless an implementation provides illegal strings for `indent-unit` and `line-ending`.

> If line-ending changes the content of elements or attributes, when does this happen in terms of the "phases of serialization" in section 4? How does it interact with, for example, character map expansion and unicode normalization?

The substitution is the final step; it operates on the character stream the serializer emits.

> I'm really not at all sure that changing the representation of newlines within the content is a good idea, unless we constrain the representation to be one that an XML parser will normalize back to a regular newline.

That precondition is met: the value domain is exactly CR/LF/CRLF, all of which an XML parser normalizes to a single U+000A.

I have revised the text, I hope it clarifies the intent.

ndw:

ndw:
Not sure where the conflicts arose. Please fix them at your convenience and then merge this PR.

ChristianGruen:
I missed this one… Done.

PR #2736, opened 2026-06-29
2169 Reformulate `StringInterpolation` to remove the `` `{ `` and `` }` `` literal terminals

### LINKED ISSUE #2169: Longest-token rule incorrectly produces `StringInterpolation` delimiter
`StringInterpolation` currently defines a two-character token, curly right brace + backtick, to follow `Expr` as a terminator:

```
StringInterpolation ::= "`{" Expr? "}`"
```

On other occasions, `Expr` is followed by a single right curly brace:

```
EnclosedExpr ::= "{" Expr? "}"
``` 

The following applies to tokenization (the "longest-token" rule):

> If the current position is not the end of the input, then return the longest literal terminal or variable terminal that can be matched starting at the current position, regardless whether this terminal is valid at this point in the grammar.”

My concern is that input like

```xquery
<a>{42}`</a>
```

is going to be mis-tokenized under the longest-token rule: after `42`, the next (longest) token is the two-character `StringInterpolation` terminator, which however is not a valid terminator of the `EnclosedExpr` serving as `CommonContent` of the direct element constructor.

**Proposed fix**

My proposal is to replace the two-character tokens that introduce and terminate `StringInterpolation` with single backticks around an `EnclosedExpr` with no intervening whitespace:

```
StringInterpolation ::= "`" EnclosedExpr "`"   /* ws: explicit */
```

This replaces both of the two-character delimiters of `StringInterpolation`, while still describing the intended language, but without causing the longest-token rule to produce a token that cannot be handled afterwards.

GuntherRademacher:
Reformulates `StringInterpolation` as proposed and discussed in #2169.

Closes #2169

### DIFF
--- a/specifications/grammar-40/xpath-grammar.xml
+++ b/specifications/grammar-40/xpath-grammar.xml
@@ -2666,12 +2666,10 @@ VersionDecl ::= "xquery" (("encoding" StringLiteral) | ("version" StringLiteral
     </g:zeroOrMore>
   </g:production>
 
-  <g:production name="StringInterpolation" if="xquery40">
-    <g:string>`{</g:string>
-    <g:optional>
-      <g:ref name="Expr"/>
-    </g:optional>
-    <g:string>}`</g:string>
+  <g:production name="StringInterpolation" if="xquery40" whitespace-spec="explicit">
+    <g:string>`</g:string>
+    <g:ref name="EnclosedExpr"/>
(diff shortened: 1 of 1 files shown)


ndw:
At [meeting 170](https://qt4cg.org/meeting/minutes/2026/06-30.html), the CG agreed to merge this PR.

PR #2744, opened 2026-07-01
2741 Record coercion: keep empty entries

### LINKED ISSUE #2741: Record coercion, drop/keep empty entries?
#2566 says that optional fields in records are to be *dropped* (there is “no longer a distinction between absent values and empty values”). But the `TypedRecordType` coercion rule still adds an empty entry for every omitted field, so a coerced record keeps all its fields. The two seem to be mutually exclusive.

An example…

```xquery
let $x as record(a as xs:integer, b as xs:string?) := { "a": 1 }
return map:size($x)
```

Coercion rule: `2` (with `b` added as `()`); with #2566: `1`.

…and the conflicting text snippets:

#2566 change note, *Record Types*:

> Optional fields in records are dropped (in records, there is no longer a distinction between absent values and empty values).

Record-construction prose, same section:

> The rules for the coercion operation ensure that the resulting record has one entry for every field in the corresponding record definition …

Coercion rules, `TypedRecordType`, the "Otherwise" clause:

> Otherwise (if J contains no such entry), an entry whose key is the name of F and whose associated value is the empty sequence; a type error occurs if the empty sequence does not match the declared type of F.

ChristianGruen:
Closes #2741

### DIFF
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -6149,9 +6149,11 @@ name.</p>
                   </change>
                   <change issue="1979" PR="2566" date="2026-03-31">
                      Records now carry a type annotation identifying a record type, and their behavior
-                     as well as their content is constrained by that record type. Optional fields in records
-                     are dropped (in records, there is no longer a distinction between absent values 
-                     and empty values).
+                     as well as their content is constrained by that record type. A record always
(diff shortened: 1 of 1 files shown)


ndw:
At [meeting 171](https://qt4cg.org/meeting/minutes/2026/07-07.html), the CG agreed to merge this PR without discussion

PR #2746, opened 2026-07-01
2745 CSV parsing: delimiters

### LINKED ISSUE #2745: CSV parsing: delimiters
The current CSV spec provides a `field-delimiter` and a `row-delimiter` option. My suggestion would be to:

#### 1. Drop `row-delimiter`

* Line endings (CR, CRLF) are already normalized to `\n` before rows are split.
* No mainstream CSV dialect uses a non-newline record separator.
* Python, Java Apache Commons, Go, R, … don’t let you change it at all when reading. It seems to be overwhelmingly for serialization.
* Removing the option also simplifies the CV0002/CV0003 error conditions. It would also simplify #2698.

#### 2. Rename `field-delimiter` to `separator`:

* “Separator” is predominant across the spec (we have `item-separator`, `decimal-separator`, `grouping-separator`, …); there is no other `-delimiter`.
* With `row-delimiter` gone, the `field-` prefix is redundant and can be dropped.

ChristianGruen:
Closes #2745

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -28983,20 +28983,11 @@ element-to-map-plan((<a><b/><b/></a>, <a><b/><c/></a>))
          
          <p>The entries that may appear in the <code>$options</code> map are as follows:</p>
          <fos:options>
-            <fos:option key="field-delimiter">
-               <fos:meaning>The character used to delimit fields within a record. An instance of
+            <fos:option key="separator">
+               <fos:meaning>The character used to separate fields within a record. An instance of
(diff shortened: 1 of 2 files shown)


ChristianGruen:
One more reason for unfying the options, as I see it, is that it will streamline serialization (#2748): We don’t need competing options for newlines.

ndw:

ndw:
Please resolve the conflicts at your convenience and merge this PR.

ChristianGruen:
@michaelhkay Thanks in advance for updating the tests; I’ll skip this step.
For the remaining PRs that go back to me, I’ll certainly update the tests this month (the earlier the better).

PR #2747, opened 2026-07-01
2591 Grammar: make step?lookup valid

### LINKED ISSUE #2591: Grammar: step?lookup is currently invalid
According to the current grammar, I believe that `step?lookup` is not a valid expression. My reasoning (let’s see if I missed something):

1. `step` on its own is not a `PrimaryExpr`. A name can only occur as a `NameTest` inside `StepExpr`/`AxisStep`.
2. A lookup attaches only to a `PostfixExpr`, which starts with a `PrimaryExpr`.
4. Inside a `RelativePathExpr`, consecutive steps must be separated by `/` or `//`.

I think we can live with the restriction, but it may be a reason for avoiding wild mixtures of paths and lookups in practice.

Related is the last example in [`map:build`](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-map-build). It uses `./ancestor-or-self::jnode()?name`, and I think it needs to be `(./ancestor-or-self::jnode())?name`.

ChristianGruen:
Closes #2591

### DIFF
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -12553,7 +12553,7 @@ return if (every $r in $R satisfies $r instance of gnode())
 	               as well as XNodes (found in trees representing parsed XML).
 	            </change>           
                 <change issue="2624" PR="2627" date="2026-05-11">Error code XPTY0020 is replaced by XPTY0004.</change>
-                <change issue="2591" PR="2591" date="2026-07-01">An axis step may now be followed
+                <change issue="2591" PR="2747" date="2026-07-01">An axis step may now be followed
                    directly by a <nt def="Lookup">lookup</nt> (as well as by predicates), so that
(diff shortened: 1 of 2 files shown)


michaelhkay:
The metadata for the changes says PR2591 rather than PR2747.

specifications/xquery-40/src/expressions.xml
```diff
@@ -12553,6 +12553,9 @@ return if (every $r in $R satisfies $r instance of gnode())
 	               as well as XNodes (found in trees representing parsed XML).
 	            </change>           
                 <change issue="2624" PR="2627" date="2026-05-11">Error code XPTY0020 is replaced by XPTY0004.</change>
+                <change issue="2591" PR="2591" date="2026-07-01">An axis step may now be followed
```
ChristianGruen:
```suggestion
                <change issue="2591" PR="2747" date="2026-07-01">An axis step may now be followed
```

ndw:

PR #2752, opened 2026-07-03
1697 Add documentary names to callback function signatures

### LINKED ISSUE #1697: Add documentary names to callback function signatures
If PR #1696 is accepted we can add documentary names to the parameters of callback function signatures, for example `fn:filter` can become 

```
fn:filter(
  $input as item()*, |  
  $predicate as fn($item as item(), $position as xs:integer) as xs:boolean? |  
) as item()*

```
and we can (if we need to) use the parameter names in the prose

ChristianGruen:
Summary:
* All function parameters are now decorated with documentary variables.
* The variables are rendered in a grey/gray/grau color, to indicate that they are not assignable.
* In various places (certainly not everywhere) I have changed the function prose to reference the variables.
* For callback functions I have (finally) added a section to explain what the `$pos` parameters are about, and that they can be omitted.
* The variables in function parameters in record definitions now consistently use a single character (for example: `$s as xs:string, $replacement as ...` → `$s as xs:string, $r as ...`).
* Minor consistency tweaks (use `function() → fn()` everywhere) and fixes.

It seems that no changes were needed for the EXPath functions.
Feedback is welcome!
Closes #1697

### DIFF
--- a/specifications/css/xpath-functions-40.css
+++ b/specifications/css/xpath-functions-40.css
@@ -125,6 +125,12 @@ table.proto tr td:nth-child(3) code {
     white-space: nowrap;
 }
 
+/* Callback parameter names are documentary only (the caller supplies the function
+   and chooses its own names); mute them to set them apart from assignable arguments */
+.doc-param {
+    color: #8f7f8f;
+}
+
 /*
 table.record tr.arg td code.req {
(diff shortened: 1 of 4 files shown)


ndw:

Issue #1459, opened 2024-09-24
Function properties and arities


ChristianGruen:
I try to understand the semantics behind the **Properties** section in the XQFO spec. Here are some examples of existing functions:

1. `fn:format-integer`

> The two-argument form of this function is ·deterministic·, ·context-dependent·, and ·focus-independent·. It depends on default language.
> The three-argument form of this function is ·deterministic·, ·context-independent·, and ·focus-independent·.

2. `fn:format-time`

> This function is ·deterministic·, ·context-dependent·, and ·focus-independent·. It depends on default calendar, and default language, and default place, and implicit timezone.

3. `fn:index-of`

> The two-argument form of this function is ·deterministic·, ·context-dependent·, and ·focus-independent·. It depends on collations, and implicit timezone.
> The three-argument form of this function is ·deterministic·, ·context-dependent·, and ·focus-independent·. It depends on collations, and static base URI, and implicit timezone.

4. `fn:element-number`

> The zero-argument form of this function is ·deterministic·, ·context-dependent·, and ·focus-dependent·.
> The one-argument form of this function is ·deterministic·, ·context-independent·, and ·focus-independent·.
> The two-argument form of this function is ·deterministic·, ·context-independent·, and ·focus-independent·.
> The three-argument form of this function is ·deterministic·, ·context-independent·, and ·focus-independent·.

I am afraid I cannot give hints for improvement yet; I just noticed it is increasingly difficult to interpret. For the moment, I wonder…

1. Are all of these current definitions correct?
2. In many cases, if the user supplies an empty sequence as argument, the default value is used instead. This contrasts with the current property description that relies a lot on the function arities. For example, `fn:index-of($a, $b)` and `fn:index-of($a, $b, ())` should have identical properties, but the current wording implies that the three-argument of this function has different properties, e.g., it is said to rely in this case on the static base URI.
3. Some functions with default values have a single property description (see `fn:format-integer`), other have multiple descriptions (see `fn:format-time`). Is there a particular rule behind that?
3. Could we merge the descriptions for arities with identical properties?
4. Maybe it will be easier not to consider the arities at all anymore?

ndw:
I was looking at those sections recently as well. I think it would be nice if we had explicit markup for the properties, so that I could easily and reliably build a table of all the functions and their properites, for example. But I became discouraged by the variety (the two argument form..., the three argument form...) and the comments about dependencies (calendar, time zone, etc.)

michaelhkay:
[...]

PR #2753, opened 2026-07-03
2300 fn:doc: determinism rules

### LINKED ISSUE #2300: `fn:doc`: determinism rules
I think the rules for `fn:doc` must be stabilized with regard to determinism.

First, the function properties say:

> This function is ·deterministic·, ·context-dependent·, and ·focus-independent·. It depends on available documents, and executable base URI.

This is outdated (it does not apply if the `stable` option is enabled), but this can possibly be fixed along with #1459.

Next, the text says:

> By default, this function is ·deterministic·. Two calls on this function return the same document node if the same URI Reference (after resolution to an absolute URI Reference) is supplied to both calls.
> …
> Note: This equivalence applies only because the two calls on the `fn:doc` function have the same options and the same static and dynamic context, to the extent this is relevant. 

This implies that the function will be deterministic if the options are the same; but we should at least exclude the `stable: false` case. Next, it implies that `fn:doc($doc, { 'include': true() }) is fn:doc($doc, { 'include': true() })` **MUST** return `true()`, but we could be more clear.

Maybe it is easier to define the function as ·nondeterministic· by default, and describe the cases in which it will be deterministic?

ChristianGruen:
I have chosen a rather minimal approach, by making more explicit for this specific function when the function is (non)deterministic. The result could possibly change when we #1459, but for now this PR will make the rules correct.

Closes #2300

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -19458,6 +19458,10 @@ return tokenize(normalize-space($s), ' ')[. castable as xs:IDREF]</eg>
                   return different document nodes, and if the underlying resource changes between
                   the calls, the contents of these nodes may differ.</p>
             </item>
+            <item>
+               <p>If two calls supply different values for <code>stable</code>, the effect is the
+                  same as if both calls supplied <code>false</code>.</p>
+            </item>
          </ulist>
          <note diff="chg" at="issue2300"><p>If two calls on <function>fn:doc</function>
            have different dynamic contexts, then the mapping from URIs to document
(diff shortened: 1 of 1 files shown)


michaelhkay:
A sllightly pedantic comment, but we don't say what happens if one call on doc says stable=true and another says stable=false. We should say that it's the same as if both said stable=false.

michaelhkay:
It might also be a good idea to firm up the corresponding rules in doc-available.

ChristianGruen:
Thanks; I have considered the two comments in my revised version. I tried to avoid too much redundancy.

@michaelhkay Btw, it’s a bit of a pain to properly maintain the PR references in all the commits. Maybe the issue number is sufficient when only one PR is presented as a solution? – If we want the PR numbers nevertheless, I would recommend adding them later via a simple batch script, which looks up the missing PRs, based on the issue number, on GitHub.

michaelhkay:
A minor clarification, where we say "two calls on this function return the same document node" it might be useful to qualify this: "two calls on this function **within the same execution scope** return the same document node". (This affects the behaviour of fn:transform, for example).

[Re PR numbers: if it's too much of a pain, then sure, just include the issue number for now, and we can add the PR number later. I go through them from time to time. But it's always more work to identify the PR later, so I prefer to do it at the time.]

ndw:

PR #2754, opened 2026-07-03
2284 fn:csv-doc, fn:json-doc: Equivalent code, unifications

### LINKED ISSUE #2284: fn:csv-doc, fn:json-doc: Equivalent code, unifications
Now that we have dropped binary types from `fn:parse-csv`, we need to fix the rules of `fn:csv-doc`, specifically:

> The effect of the two-argument function call `fn:csv-doc($H, $M)` is equivalent to the function composition `fn:unparsed-binary($H) => fn:parse-csv($M)`.

Next, we should re-align the rules of `fn:csv-doc` and `fn:json-doc`. The detection of the encoding should be clarified, possibly by referring to the rules proposed in #2282, or the current version of `bin:decode-string`.

ChristianGruen:
Editorial, and overdue: Ensures that the `-doc` functions use the same rules.

Closes #2284

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -22264,7 +22264,8 @@ serialize(
          <p>Reads an external resource containing HTML, and returns the result of parsing the resource as HTML.</p>
       </fos:summary>
       <fos:rules>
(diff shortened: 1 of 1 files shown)


ndw:

PR #2755, opened 2026-07-03
2620 fn:unparsed-text: editorial issues

### LINKED ISSUE #2620: fn:unparsed-text: editorial issues
Adopted from #2408:

* [x] Reference `bin:infer-encoding`, drop redundant rules

ChristianGruen:
Cleanups (`fn:unparsed-text` was streamlined).
Mostly editorial, but we’ll need some more tests here (with or without this PR).

Closes #2620


michaelhkay:
I'm not sure why you've dropped the parenthetical in "The values which an implementation is required to recognize are UTF-8, UTF-16, UTF-16LE, and UTF-16BE (in any upper/lower case combination)." I'm not sure that the reference to the XML spec is sufficient to indicate that lower-case versions must also be accepted.

ChristianGruen:
I have readded this sentence. My (half-baked) reason for removing it that I believe we need to find one central place to define mandatory encodings. I have added an Editorial Note for it (https://github.com/qt4cg/qtspecs/issues/2772#issuecomment-4904756019).

ndw:

PR #2757, opened 2026-07-03
2569 Rules for unprefixed function calls with explicit default function ns

### LINKED ISSUE #2569: Inconsistent rules for unprefixed function calls with explicit default function namespace
The spec appears to give conflicting rules for resolving unprefixed function calls when an explicit default function namespace is declared.

Example:

```xquery
declare default function namespace "urn:test";
declare function f() { "namespace" };
declare function Q{}f() { "no-namespace" };
f()
```

Two readings seem possible:

- from [5.14.2 Default Function Namespace Declaration](https://qt4cg.org/specifications/xquery-40/xquery-40.html#id-default-function-namespace-declaration) and the [2.2.1 Static Context](https://qt4cg.org/specifications/xquery-40/xquery-40.html#static_context) component "default function namespace":
   `f()` refers to the function in "urn:test", yielding "namespace"
- from the [default function namespace rule](https://qt4cg.org/specifications/xquery-40/xquery-40.html#dt-default-function-namespace-rule):
   the no-namespace function is preferred, yielding "no-namespace"

ChristianGruen:
After further reflection, I have chosen to have three states:

* The no-namespace-first search exists for one simple reason: without any declaration, the default function namespace is the reserved `fn:` namespace, where users cannot declare functions.
* This means that the search is the only way to make unprefixed calls to user-defined functions work.
* Once a default function namespace is explicitly declared, that reason disappears: users can declare functions there.
* I believe the declaration should simply mean what it says: unprefixed calls go to the declared namespace and are not silently redirected by some no-namespace function that happens to exist.

The implementation is straightforward, and the proposed solution goes hand in hand with #2257.

Closes #2569

### DIFF
--- a/specifications/xquery-40/src/back-matter.xml
+++ b/specifications/xquery-40/src/back-matter.xml
@@ -104,7 +104,7 @@ per lexical scope.</td>
 
 <tr>
   <td>Default function namespace</td>
-  <td><code>fn</code></td>
+  <td><phrase diff="chg" at="issue2569">absent</phrase></td>
   <td>overwriteable (not recommended)</td>
   <td>overwriteable by <specref ref="id-default-namespace"/></td>
   <td>no</td>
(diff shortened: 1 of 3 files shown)


ndw:
At [meeting 172](https://qt4cg.org/meeting/minutes/2026/07-14.html), the CG agreed to merge this PR.

PR #2759, opened 2026-07-04
1506 Type declarations: Constructor functions

### LINKED ISSUE #1506: Type declarations: Constructor functions?
I’ve tagged this as a discussion issue:

Would it make sense to declare constructor functions for the new XQuery type declarations, similar to what we now have for records, or do we believe that’s over the top?

It might improve typing in complex code, and it would allow us to write things like:

```xquery
declare type world:continent as enum('Africa', 'America', 'Asia', 'Australia', 'Europe');
world:continent('Africa')
```

For the type above, an implicit constructor function would be created that would simply coerce its argument to the declared type:

```xquery
declare function world:continent($value) as world:continent { $value };
```

One current drawback is that this only works for prefixed types (see #657), whereas it’s possible to define types without prefix, and reference them in local type declarations:

```xquery
declare type continent as enum('Africa', 'America', 'Asia', 'Australia', 'Europe');
let $c as continent := 'Africa'
return $c
```

ChristianGruen:
[NEW] This PR originally introduced constructor functions for type declarations, restricted to enums and atomic types (see below). As it turned out, the feature already existed in the spec (which also explains the existing test case).

My confusion was caused by the fact that we closed #1506 without further notice, and that this was not mentioned in the *Item Type Declarations* section. Next, its note claims that item type names never need to be retained at execution time, which contradicts the constructor functions. The PR tries to complete the picture, without changing the scope of the feature:

* The section on item type declarations points to the existing constructor function rules.
* Annotations rules/arity overlap have been added, they seemed to be missing.
* A note has been corrected.

→ And we definitely need more test cases (indicated, for now, with the »Tests Needed« label). I will do my best soon.

-------------

[OLD] This PR introduces constructor functions for type declarations – a feature for which, surprisingly, we already have a test case, [itemTypeDecl-026](https://github.com/qt4cg/qt4tests/blob/c0183a06487621e5f218b2e44f69acd02b39a8a4/prod/ItemTypeDecl.xml#L381-L394)… (we certainly need more)

I decided to restrict the creation of constructor functions to declarations of enums and atomic types, for the following reasons mainly:

* We avoid different semantics for `record` and `type ... as record` declarations.
* Functions (incl. maps & arrays) can’t be built from a value; it would basically be a simulation of `instance of`/`treat as`.
* Unions and choices are non-trivial: for `declare type n as (xs:double | xs:decimal)`, `n(1.5)` the result depends on the order in which the type was written, which is easy to get wrong.

### DIFF
--- a/specifications/xquery-40/src/query-prolog.xml
+++ b/specifications/xquery-40/src/query-prolog.xml
@@ -2047,7 +2047,14 @@ local:depth(doc("partlist.xml"))
   
   <div2 id="id-item-type-declaration" diff="add" at="A">
     <head>Item Type Declarations</head>
-    
+
+    <changes>
+      <change issue="1506" PR="TODO" date="2026-07-04">
+        Clarified that an item type declaration whose type expands to a generalized atomic type
+        or to a record type implicitly declares a constructor function.
+      </change>
+    </changes>
+
     <p>An item type declaration defines a name for an <termref def="dt-item-type"/>. Defining a name for an item type
     allows it to be referenced by name rather than repeating
     the <termref def="dt-item-type-designator"/> in full.</p>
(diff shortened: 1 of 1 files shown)


ndw:

PR #2760, opened 2026-07-04
2521 URIs: unify handling of fragment identifiers

### LINKED ISSUE #2521: File retrieval, URIs: fragment identifiers
Fragment identifiers are now allowed in `fn:doc` calls, but they are still illegal in other functions (like `fn:unparsed-text`), for no apparent reason. We should generally allow them.

While reviewing qt4cg/qt4tests#340, I noticed again that the error codes for `fn:unparsed-text` are a bit of a mess, and difficult to adhere to when using low-level I/O:

* A dynamic error is raised [err:FOUT1170] if the $source argument contains a fragment identifier, or if it cannot be resolved to an absolute URI (for example, because the base-URI property in the static context is absent), or if it cannot be used to retrieve the string representation of a resource.
* A dynamic error is raised [err:FOUT1190] if the value of the encoding option is not a valid encoding name, if the processor does not support the specified encoding, if the string representation of the retrieved resource contains octets that cannot be decoded into Unicode characters using the specified encoding, or if any resulting character is not a permitted character.
* A dynamic error is raised [err:FOUT1200] if the encoding option is absent and the processor cannot infer the encoding using external information and the actual encoding is not UTF-8.

Things don’t get better with the introduction of the fallback option. I would prefer to see them merged to single error code, preferably `FOUT1170`, and to simplify the description to…

> A dynamic error is raised [err:FOUT1170] if the the string representation of a resource cannot be retrieved.

…or to at least drop FOUT1200 and merge it with FOUT1190.

ChristianGruen:
This PR proposes to treat fragment identifiers identically for `fn:doc`, `fn:unparsed-text`, and `fn:unparsed-binary`. I have added a reference to RFC5147.

Closes #2521

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -19880,8 +19880,7 @@ return tokenize(normalize-space($s), ' ')[. castable as xs:IDREF]</eg>
          <p diff="add" at="2026-07-04">The URI may include a fragment identifier. The effect of a
             fragment identifier is <termref def="implementation-defined"
             >implementation-defined</termref>; one possible interpretation for text resources
-            is described in <bibref ref="rfc5147"/>. Implementations that associate no semantics
-            with the fragment identifier <rfc2119>should</rfc2119> ignore it.</p>
+            is described in <bibref ref="rfc5147"/>.</p>
          <p>If <code>$source</code> is a relative URI reference, it is resolved relative to the value
             of the <xtermref spec="XP40" ref="dt-executable-base-uri">executable base URI</xtermref> property 
             from the dynamic context of the caller. The resulting absolute URI is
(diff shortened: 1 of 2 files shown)


ndw:
I'm very uneasy about making fragment identifiers implementation defined and giving the implementation freedom to ignore them completely. To my mind, that's roughly analogous to saying that the `fn:doc` function can arbitrarily return any document the implementation chooses, without regard to the specified URI. 

(I recognize that this is in the draft now, so I failed to notice the change when it was made. Apologies for slightly hijacking this PR.)

Ignoring the fragment identifier silently has the potential to completely change the result of the expression. In a publishing scenario with large documents, this could easily go unnoticed and have serious implications.

I think it would be much safer and more reliable to say that it is an error if the processor encounters a fragment identifier in a scheme that it does not recognize or cannot support.

ChristianGruen:
Right, the current spec already says for `fn:doc`:

> The effect of a fragment identifier in the supplied URI is [implementation-defined](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#implementation-defined). One possible interpretation is to treat the fragment identifier as an ID attribute value, and to return a document node having the element with the selected ID value as its only child.

My main motivation would be to align the rules. We could reject them for `fn:doc` again, but I think we shouldn’t preserve a different behavior for `fn:unparsed-binary` and `fn:unparsed-text`.

ndw:
Yes. It's an orthogonal point and I should raise a separate issue. I'll try to remember to bring this up in passing at least tomorrow and get the temperature of the group. If there' so support, I'll quietly give up. But I will be taking an "I told you so" token for future use.

Arithmeticus:
Is this a case where we need to specify that implementations **must** support a handful of select fragid schemes (e.g., XML, SVG), then indicate that support for other fragid schemes are implementation defined? Similar to what we did for `fn:hash()`.

PR #2766, opened 2026-07-05
2765 Destructuring let: records, array fix

### LINKED ISSUE #2765: Destructuring let: records
The current spec says:

```xquery
let ${ $a, $b as xs:integer, $local:c } := { "a": 2, "b": 4, "c": 6, "d": 8 }
return $a + $b + $local:c
```

…is expanded to:

```xquery
let $temp := { "a": 2, "b": 4, "c": 6 }
let $a := map:get($temp, "a", ())
let $b as xs:integer := map:get($temp, "b", ())
let $local:c := map:get($temp, "c", ())
return $a + $b + $local:c
```

We should use the lookup operator or a dynamic function call, so that `let ${unknown}` results in an error.

A side note, the rules for arrays say:

> … replaced by a LetValueBinding of the form `let Ai as Ti := array:get(V, i)`

The subsequent example uses `array:get($temp, 1, ())`.

ChristianGruen:
Closes #2765

### DIFF
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -19393,8 +19393,13 @@ for member $y in $expr2]]></eg>
                to the declared type by applying the coercion rules.
                </change>
                <change issue="37" PR="2055" date="2025-06-17">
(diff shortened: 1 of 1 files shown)


ndw:

PR #2768, opened 2026-07-05
2740 Drop constructor functions from built-in records

### LINKED ISSUE #2740: Built-in records, constructor functions: optional fields
The record type definitions in the specs contain outdated `extensible` and `required` attributes, which I believe can be dropped:

```xml
<fos:record-type id="element-conversion-plan-record" extensible="false">
  <fos:field name="layout" type="enum('empty', ...)" required="true">
     ...
  </fos:field>
  <fos:field name="child" type="xs:string?" required="false">
     ...
  </fos:field>
  <fos:field name="type" type="enum('integer', ...)" required="false">
     ...
  </fos:field>
</fos:record-type>
```

But the more interesting question is the definition of the constructor functions: If the type of a record field indicates that the entry is optional, we could assign `()` as the default argument. However, this would require all subsequent parameters to be optional as well:

```
declare record fn:element-conversion-plan-record(
   layout  as enum('empty', ...),
   child   as xs:string? := (),
   type    as enum('integer', ...) (: ↯ :)
)
```

I can think of the following solutions:
1. make all function arguments in the constructor function mandatory;
2. reorder fields of built-in record types in the spec, by placing optional fields last;
3. drop constructor functions for built-in record types from the spec.

Option 3 would be my favorite (unless we add at least one example in the prose how to benefit from the built-in functions. Otherwise, they will go wildly unnoticed anyway).

ChristianGruen:
This PR drops constructor functions for built-in record types. My rationale:

1. We avoid a bunch of rarely-used functions that would mostly go unnoticed anyway. If keeps the `fn:` namespace clean.
2. We don’t have to deal with a broken argument order (some records list optional fields before required ones).
3. Records can be built with a simple map. If required, coercion is available to turn them into the record.

If we want to keep the constructor functions, we should add at least one example how to benefit from them.

Closes #2740

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -38,57 +38,57 @@
       <fos:summary>
          <p>This record type represents the components of a URI.</p>
       </fos:summary>
-      <fos:field name="uri" type="xs:string?" required="false">
+      <fos:field name="uri" type="xs:string?" required="true">
          <fos:meaning>
             <p>The original URI. This element is returned by <function>fn:parse-uri</function>,
             but ignored by <function>fn:build-uri</function>.</p>
          </fos:meaning>
       </fos:field>
-      <fos:field name="scheme" type="xs:string?" required="false">
+      <fos:field name="scheme" type="xs:string?" required="true">
          <fos:meaning>
             <p>The URI scheme (e.g., “https” or “file”).</p>
          </fos:meaning>
       </fos:field>
-      <fos:field name="absolute" type="xs:boolean?" required="false">
+      <fos:field name="absolute" type="xs:boolean?" required="true">
          <fos:meaning>
             <p>The URI is an absolute URI.</p>
          </fos:meaning>
       </fos:field>
-      <fos:field name="hierarchical" type="xs:boolean?" required="false">
+      <fos:field name="hierarchical" type="xs:boolean?" required="true">
          <fos:meaning>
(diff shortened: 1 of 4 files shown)


ChristianGruen:
The upcoming editorial revision of this PR will replace the obsolete syntax at some places, for example:

* `declare record my:list (value as item()*, next? as my:list);`
* `typed-node? as node(), error-details? as map(*)*)`

ChristianGruen:
I tried to completely remove the `extensible` and `required` attributes, but as I failed to recreate the `.scm` file, I restricted myself to the obvious cleanups.

ChristianGruen:
Conflicts resolved; hopefully correct.

michaelhkay:
Simply saying that these are named item types rather than named record types creates possible problems for the random-number-generator record, since named item types are not allowed to be recursive.

I think we need to make constructor functions optional for all named record types, and then say that these particular record types don't have one.

michaelhkay:
What is the status of this?

I see that quite recently tests have been added e.g. to fn-sort-by that introduce new constructor functions for new record types.

ChristianGruen:
> What is the status of this?

I believe to remember #2787 was a reaction on this PR. But I’ll be glad to resolve the conflicts on this PR so that we can discuss it again.

ndw:
At [meeting 178](https://qt4cg.org/meeting/minutes/2026/09-01.html), the CG agreed to merge this PR.

ndw:
Please resolve the merge conflicts and merge this PR.

PR #2784, opened 2026-07-12
2783 fn:pad-string

### LINKED ISSUE #2783: Padding strings
A simple one:

XPath can format numbers and dates, but it cannot pad a string to a fixed width, although it would often be helpful when creating textual output (reports, log lines, etc.). That’s something most other languages provide.

Users must write it themselves, and the usual workarounds are easy to get wrong. `substring($s || '     ', 1, 5)` only works if the literal is long enough, and it truncates. `$s || string-join(replicate(' ', 5 - string-length($s)))` needs two functions and does not feel very intuitive.

The signature, as I would draft it:

```xquery
fn:pad-string(
  $value    as item()?,
  $length   as xs:integer,
  $options  as map(*)? := {}
) as xs:string
```

The options could be `fill` (default `" "`) and side (`end` by default, `start`, `both`). Examples:

```xquery
pad-string('abc', 10)                                  (: "abc       " :)
pad-string(42, 6, { 'fill': '0', 'side': 'start' })    (: "000042" :)
pad-string('Chapter 1', 20, { 'fill': '.' })           (: "Chapter 1..........." :)
```

ChristianGruen:
Closes #2783

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -34243,6 +34243,125 @@ return every($dl/*, fn($elem, $pos) {
       </fos:changes>
    </fos:function>
 
+   <fos:function name="pad-string" prefix="fn">
+      <fos:signatures>
+         <fos:proto name="pad-string" return-type="xs:string">
+            <fos:arg name="value" type="xs:anyAtomicType?"/>
+            <fos:arg name="length" type="xs:integer"/>
+            <fos:arg name="options" type="map(*)?" default="{}" note="default-on-empty"/>
+         </fos:proto>
+      </fos:signatures>
+      <fos:properties>
+         <fos:property>deterministic</fos:property>
+         <fos:property>context-independent</fos:property>
+         <fos:property>focus-independent</fos:property>
+      </fos:properties>
+      <fos:summary>
+         <p>Returns a string created by extending a value to a given length with fill
+            characters.</p>
+      </fos:summary>
+      <fos:rules>
+         <p>The coercion rules ensure that the supplied <code>$value</code> argument is first
+            converted to an atomic item by applying atomization. If <code>$value</code> is the empty
+            sequence, the <var>input string</var> is the zero-length string; otherwise it is the
+            result of casting the atomized item to an <code>xs:string</code>.</p>
+         <p>The <code>$options</code> argument controls the padding.
+            The <termref def="option-parameter-conventions"/> apply.</p>
+         <fos:options>
+            <fos:option key="fill">
+               <fos:meaning>The string that is repeated to form the fill characters.
+                  It must not be a zero-length string.</fos:meaning>
+               <fos:type>xs:string</fos:type>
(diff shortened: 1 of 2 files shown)


michaelhkay:
(1) if the supplied value is a node, it seems more intuitive to apply the function to its string-value, not to the result of atomizing and then casting the result to a string. As specified, for example, it will lose insignificant leading or trailing zeroes from a number.

(2) Perhaps use `justify = left | right | center` instead of the proposed `side` option?

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -34243,6 +34243,125 @@ return every($dl/*, fn($elem, $pos) {
...
+         <p>Padding a number to a fixed width, with leading zeroes or otherwise, can also be
+            achieved with <function>fn:format-integer</function> and
+            <function>fn:format-number</function>, which additionally allow the digits, the sign,
+            and any grouping separators to be controlled.</p>
+      </fos:notes>
+      <fos:examples role="wide">
+         <fos:example>
+            <fos:test>
+               <fos:expression>pad-string("abc", 6)</fos:expression>
+               <fos:result>"abc   "</fos:result>
```
Arithmeticus, line 34310: In the HTML version, the three spaces have been collapsed into only one space.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -34243,6 +34243,125 @@ return every($dl/*, fn($elem, $pos) {
...
+            </fos:test>
+         </fos:example>
+         <fos:example>
+            <fos:test>
+               <fos:expression>pad-string("abc", 6, { "side": "start" })</fos:expression>
+               <fos:result>"   abc"</fos:result>
```
Arithmeticus, line 34316: In the HTML version, the three spaces have been collapsed into only one space.

specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -34243,6 +34243,125 @@ return every($dl/*, fn($elem, $pos) {
...
+      </fos:summary>
+      <fos:rules>
+         <p>The coercion rules ensure that the supplied <code>$value</code> argument is first
+            converted to an atomic item by applying atomization. If <code>$value</code> is the empty
+            sequence, the <var>input string</var> is the zero-length string; otherwise it is the
+            result of casting the atomized item to an <code>xs:string</code>.</p>
+         <p>The <code>$options</code> argument controls the padding.
+            The <termref def="option-parameter-conventions"/> apply.</p>
+         <fos:options>
+            <fos:option key="fill">
```
Arithmeticus: I recommend the name `padding`, to match the name of the function, and to differentiate verb from noun. Alternatively, the function might be named `fill-string`, in which case the map key should be renamed `filling`.

Arithmeticus (commented):
This function seems useful. I personally prefer `side` to `justify`, because the act of padding/filling is not coterminous with that of justification (e.g., padding a string with itself).

PR #2792, opened 2026-07-16
2790 fn:pad-string, truncation

### LINKED ISSUE #2790: Truncation of fill string in fn:pad-string()
Given `pad-string("x", 5, {"fill": "0123456789", "side":"start"})`

I don't think it's obvious from the spec whether the expected result is `0123x` or `6789x` (the latter seems preferable to me).

It's even less obvious with `side:both`.

There don't seem to be any examples that use a multi-character fill string.

ChristianGruen:
I hope it’s less ambiguous now. Closes #2790
No extra QT4 tests needed (the PR contains two new examples).


ndw:
At [meeting 174](https://qt4cg.org/meeting/minutes/2026/07-28.html), the CG agreed to merge this PR without discussion.

PR #2796, opened 2026-07-17
2352 Trusted execution: clarifications

### LINKED ISSUE #2352: Trusted execution: questions and observations
I am trying to understand the exact implications of the trusted execution mode that we have recently introduced. The spec says:

### [2.3 External Resources and Security](https://qt4cg.org/specifications/xquery-40/xpath-40.html#id-security-resources)

> [Definition: The static context includes a boolean property called trusted that determines whether external resources are available.] This may take the following values:
> * **false**: No external resources are available other than resources explicitly made available by the caller through some trusted implementation-defined mechanism.
> 
> In the interests of security, the default for these options is **false**.

How will the rules be interpreted by Saxon? For example,
1. What is meant by »external resources«? Does it refer to remote resources (`https://abc.de/fgh.xml`) or also to local files?
2. How would the behavior change if the default is changed to `true`?
3. What could be done to allow or disallow access to local files?

ChristianGruen:
Closes #2352

The security feature “trusted” controls whether running code is allowed to reach external resources (like files or web documents). The same word trusted was being used for two different things, which was confusing:

1. Who is running: is the code itself trusted? (This one stays unchanged)
2. What a parse may fetch: when reading an XML document, may it also pull in extra linked things like external entities, a DTD, or XInclude content? (I renamed this one to `trust-external`).

In addition, I tried to clear up loose ends:

1. It says plainly that “code” is trusted (not “documents”).
2. it adds the missing rule for who is trusted by default: a query you run yourself is trusted; code loaded indirectly (via `fn:load-xquery-module`, etc.) is not.
3. I clarified that untrusted code cannot grant itself access just by setting the option (access needs both trusted code and the option).
4. Cleanups

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -19403,7 +19403,7 @@ return tokenize(normalize-space($s), ' ')[. castable as xs:IDREF]</eg>
             are as follows:</p>
          
          <fos:options>
-            <fos:option key="trusted">
+            <fos:option key="trust-external">
                <fos:meaning>Indicates whether processing the document may cause other
                   external resources to be fetched (including, for example, external entities, an external DTD,
                   or documents referenced using <code>xsi:schemaLocation</code> or
@@ -19687,7 +19687,7 @@ return tokenize(normalize-space($s), ' ')[. castable as xs:IDREF]</eg>
                /> if XSD validation is
             carried out during XML parsing, and the supplied content is not valid against the relevant XSD schema.</p>
          <p>A dynamic error is raised <errorref class="DC" code="0016"/> if external resources are required 
-            (for example, external entities) that are not available because the <code>trusted</code>
+            (for example, external entities) that are not available because the <code>trust-external</code>
             option is not set to <code>true</code>.</p>
       </fos:errors>
       <fos:see-also prefix="fn" name="parse-xml"/>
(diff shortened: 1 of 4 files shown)


michaelhkay:
The XSLT document() function should align with fn:doc().

I wonder if `allow-external` would be a better keyword than `trust-external`.

I agree that separating the concepts of trusted code from trusted documents helps to make things clearer.

ChristianGruen:
I tried to keep `trust` in the option name, to indicate that the concepts are related (but this is certainly debatable).

michaelhkay:
Reviewing this again, I'm OK with the changes, but I wonder if the option on transform(), load-xquery-module(), and xsl:evaluate could be a predicate callback applied to a URI that grants or rejects access to that URI?

ChristianGruen:
> I wonder if the option on transform(), load-xquery-module(), and xsl:evaluate could be a predicate callback applied to a URI that grants or rejects access to that URI?

For these functions, `trusted` currently says whether the invoked code uses the caller’s privileges. A URI predicate would not cover it. But I guess you think of it as a replacement for `trust-external`, which e.g. is offered by `fn:doc` – or would you prefer to equip `fn:load-xquery-module` et al. with both options?

Maybe we could tackle this in a separate issue?

ndw:
At [meeting 178](https://qt4cg.org/meeting/minutes/2026/09-01.html), the CG agreed to merge this PR.

PR #2797, opened 2026-07-17
2699 fn:element-to-map: validation

### LINKED ISSUE #2699: fn:element-to-map: validation
In #1949, I suggested to strictly follow the type rules supplied by users. Specifically, I believe a value that is specified to be a number should be output as such, or an error should be raised.

If we decide to stick with a more lax approach (as proposed by the PR for this issue), maybe we could introduce a `mode` option instead:

```xquery
(: would raise an error, as 'haha' is no boolean :)
element-to-map(
  <a>haha</a>,
  { 'plan': { 'a': { 'layout': 'simple', 'type': 'boolean' } }, 'strict': true() }
)
```

I would prefer `strict` to be the default. Otherwise, when you design and apply a plan, you will never know which rules are observed and which are ignored.

ChristianGruen:
Closes #2699

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -28851,6 +28851,20 @@ element-to-map-plan((<a><b/><b/></a>, <a><b/><c/></a>))
                      as distinct from a descendant of such an element.</fos:value>
                </fos:values>
             </fos:option>
+            <fos:option key="validation">
+               <fos:meaning>Determines how a content type prescribed by the conversion plan is
+                  enforced. In <code>strict</code> mode, a value that cannot be cast to its prescribed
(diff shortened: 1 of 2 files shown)


michaelhkay:
I'm not sure about the choice of option name: `validation=strict|lax` has strong connotations of XSD validation.

I'm not sure about the phrase "if the plan explicitly prescribes a type". The type is an enum(), it can't be omitted, and we say the default is string.

ChristianGruen:
Perhaps `liberal`, in analogy with `fn:json-to-xml`?

ChristianGruen:
I have revised the PR and renamed the `validation` to `liberal`. The type string no longer counts as a prescribed type.

ChristianGruen:
@ndw you can merge this PR today (I’ve incorporated today’s comment).

ndw:
At [meeting 179](https://qt4cg.org//meeting/minutes/2026/09-08.html), the CG agreed to merge this PR.

PR #2800, opened 2026-07-20
2799 fn:slice: negative steps

### LINKED ISSUE #2799: `fn:slice`: negative step with default start/end returns empty sequence
Users from other Python may want to use `fn:slice` to revert a sequence. Reversal works if you supply both bounds:

```xquery
slice((1, 2, 3, 4, 5), 5, 1, -1)
```

…but if you omit start and end, you get the empty sequence:

```xquery
slice((1, 2, 3, 4, 5), step := -1)
```

I think this can easily be fixed without changing the existing behavior of the function. I’ll write a PR.

ChristianGruen:
Closes #2799


ndw:

PR #2804, opened 2026-07-20
2780 Functions on CSV Data: comments

### LINKED ISSUE #2780: Comments on 'Functions on CSV Data' in XPath and XQuery Functions and Operators 4.0
I have some comments on [17.5 Functions on CSV Data](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#csv-functions)

The paragraph starting: 
>The other two functions recognize column names.

 There are 3 other functions not 2.

[17.5.4 fn:csv-to-arrays](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-csv-to-arrays)
In the examples, the results are sometimes shown as sequence of arrays enclosed in `()` and sometimes not. I know it's not that important, but is it a deliberate decision?

[17.5.6 Record fn:parsed-csv-structure-record](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#parsed-csv-structure-record)
`columns` description says:
> The sequence of strings will potentially be truncated if the number-of-columns option is specified, and it will potentially be reordered if the filter-columns option is specified.

I assume this is left over from a previous version and should talk about `select-columns`? 

[17.5.4 fn:csv-to-arrays](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-csv-to-arrays) and 
[17.5.7 fn:parse-csv](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-parse-csv)
Rules section says:
> The input supplied in $value is CSV data, as defined in [RFC 4180].

Should this refer to the extended grammer defined in 17.5, or just not have this sentence at all. Section 17.5.10 doesn't have it.

[17.5.7 fn:parse-csv](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#func-parse-csv)
trim-whitespace section:
If you have `{'header':true(),'trim-whitespace':false()}` the 'header' fields are trimmed but not the data fields.
The description of `trim-whitespace` needs to say something about this case.

trim-rows section:
It might be better if the `false` part started 'No padding or trimming of fields in the rows takes place' so that it is explicit this is about field padding/trimming.

[17.5.11 Illustrative examples of processing CSV data](https://qt4cg.org/specifications/xpath-functions-40/Overview.html#illustrative-csv-examples)
Example: Converting a CSV into an HTML-style table using fn:parse-csv
I have to say that it doesn't look like anyone has tested this. Some comments are: 
1. For the `$crlf` variable could you make the value XSLT/XQuery agnostic by just having the part after ':='? 
2. No data is output because the default option for `header`, `false()` applies to `fn:parse-csv` and `fn:csv-to-xml`.
3. In the XSLT versions some `<xsl:template>` elements have a closing tag of `</xsl:function>`.
4. In the first example the code doesn't reflect the record structure.
5. In the first example why is the namespace prefix `eg` declared and excluded when it isn't used. 
6. In the second XQuery example it would help if the references to the CSV XML elements used the correct namespace.

ChristianGruen:
…and other editorial fixes.

Closes #2780

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -450,10 +450,9 @@
                      <item><p>With <code>"header":true()</code>, the value is a sequence
                      of strings taken from the first row of the data. The strings have
                      leading and trailing whitespace trimmed, regardless of the value of the
-                     <code>trim-whitespace</code> option. The sequence
-                     of strings will potentially be truncated if the <code>number-of-columns</code>
-                     option is specified, and it will potentially be reordered if the
-                     <code>filter-columns</code> option is specified. Any strings that are
+                     <code>trim-whitespace</code> option and of whether the field is quoted. The sequence
+                     of strings will potentially be truncated and reordered if the
+                     <code>select-columns</code> option is specified. Any strings that are
                      zero-length or duplicated are retained <emph>as-is</emph>.</p>
                     </item>
                      <item><p>If the value of the <code>header</code> option is a sequence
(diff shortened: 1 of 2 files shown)


ndw:
At [meeting 174](https://qt4cg.org/meeting/minutes/2026/07-28.html), the CG agreed to merge this PR without discussion.

PR #2805, opened 2026-07-20
2802 Function identity removed

### LINKED ISSUE #2802: Remove function identity?
I think we should reconsider whether it was a good idea to give functions public identities:

* It’s a feature you can see but are told not to trust: The result of `fn:function-identity` will depend completely on the implementation.
* If we keep it, all built-in functions that create functions will need to become non-deterministic (see #1920), with the drawback that the effects of nondeterminism again depend on the implementation. Among others, this includes `fn:parse-json`, `fn:apply`, or `fn:op`.
* It could simplify the prose of `fn:jtree`. The current notes revolve alot around the function identity, but it is not clear why: Since the identity is unpredictable, the behavior of `fn:jtree` will be too.

`fn:deep-equal` could be equipped with an option to ignore functions other than maps and arrays.

I may need a memory refresher: Are there are language features that would suffer from the reversal?

ChristianGruen:
The attempt… which closes #2802 if accepted. Much of the prose is derived from the original version of the spec.

I revived the strict type error for `fn:deep-equal($fn1, $fn)`. I believe we shouldn’t re-introduce implementation-dependent behavior; I don’t think it would have enough added value.

### DIFF
--- a/specifications/xpath-datamodel-40/src/xpath-datamodel.xml
+++ b/specifications/xpath-datamodel-40/src/xpath-datamodel.xml
@@ -5906,6 +5906,9 @@ is returned.</p>
     <div2 id="function-items">
 <head>Function Items</head>
   <changes>
+    <change issue="1920" date="2026-07-20">
+      Reverts issue 520: function items again cannot be compared for identity.
+    </change>
     <change issue="520" PR="525" date="2023-05-30">
       Introduced the concept of function identity.
     </change>
@@ -5920,10 +5923,14 @@ is returned.</p>
   A <term>function item</term> is an <termref def="dt-item"/> that can be called with zero or more
   arguments to return a result.
 </termdef>
-Function items <phrase diff="del" at="2023-05-25">cannot be compared for
+Function items <phrase diff="add" at="2026-07-20">cannot be compared for
 identity, equality, or otherwise, and</phrase> have no serialization.
 </p>
-  
+
+<note diff="add" at="2026-07-20"><p>Maps and arrays are function items, but unlike other function
+items they can be compared by value; see <function>fn:deep-equal</function> and
+<xspecref spec="FO40" ref="dt-identical"/>.</p></note>
+
   <note diff="add" at="2023-03-11"><p>XDM 4.0 uses the term <term>function item</term> where XDM 3.1 used
     <term>function</term>. There is no distinction
     in meaning, but <term>function item</term> is preferred for clarity, because the unqualified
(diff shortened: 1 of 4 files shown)


michaelhkay:
I don't think it's acceptable for `deep-equal($X, $X)` ever to return anything other than true. Remember it's not just the immediate case where $X is a function item, it also affects the case where $X is a large map or array with a function item deep within it. It also affects all operations that depend on deep-equal.

For example, given:

```
declare type rectangle as record (width, height, area:=fn{?width * ?height})
let $r1 := rectangle(2, 3)
let $r2 := rectangle(2, 3)
```

I think we should try to ensure that deep-equal($r1, $r2) is true. It's no problem to achieve this in an implementation because the two rectangles "obviously" have the same area method, so it's very unhelpful for users to say you can't do this.

ChristianGruen:
> I don't think it's acceptable for `deep-equal($X, $X)` ever to return anything other than true. Remember it's not just the immediate case where $X is a function item, it also affects the case where $X is a large map or array with a function item deep within it. It also affects all operations that depend on deep-equal.

A processor is allowed to rewrite `fn { 1 } ! deep-equal(., .)` to `deep-equal(fn { 1 }, fn { 1 })`, and it is legal for every occurrence of `fn { 1 }` to be a different internal instance.

I think the only clean way out is to raise an error as before, and offer an explicit option to ignore functions. Everything else will just be random behavior, or will require a deep understanding of a particular implementation. That’s what we have today, and the long discussion before we ended up here showed it was not satisfactory.

ChristianGruen:
Function equivalence is now part of this PR. Looking forward to feedback.

michaelhkay:
This is a good step forward and I've abandoned my alternative attempt.

Ideally we should try to define function equivalence in terms of the properties of the function. I think that the name and annotation properties should probably be ignored since they don't affect the result of evaluating the function. I think it's necessary that the signatures be identical - that's a precondition for `F1(A, B, C,,,)` always delivering the same result as `F2(A,B,C,,,)`. 

It's necessary that the captured context be the same to the extent that the captured context is actually used. Not quite sure how to define that precisely.

And then it's down to comparing the function bodies, where I think we have to fudge things a bit. It comes down to "if the processor can establish that the function bodies are equivalent.

Somewhere in your text you talk about $x and $y being "the same function item". I think we should try to clarify that, otherwise it simply begs the question. I think what we're trying to say here is that operations like adding a function item to a sequence or array, binding it to a variable, extracting it from a sequence or array, derefer encing the variable, don't change the function item in any way, and operations that return an existing function item are somehow different from operations that construct a new function item. When two function items share a common history like that, we expect them to be recognized as equivalent: but I'm not sure how to express that.

ChristianGruen:
A revised version is online.

ndw:

PR #2812, opened 2026-07-23
2811 fn:decode-from-uri: to +, or not to +

### LINKED ISSUE #2811: fn:decode-from-uri: to +, or not to +
We decided that `fn:decode-from-uri` should turn a plus character into a space, so that `a+b.xml` is decoded as `a b.xml`.

That rule actually comes from HTML form data (as I was informed), not from URIs, so the function does not really do what its name promises. It does not match its own counterpart `fn:encode-for-uri` either, which writes `+` as `%2B`. In addition, it seems to break `fn:parse-uri`, which reuses the same decoding for path segments.

We could add an option to control the behavior for `+`, but it may be over the top.

ChristianGruen:
Closes #2811


ndw:
At [meeting 175](https://qt4cg.org/meeting/minutes/2026/08-11.html), the CG agreed to merge this PR.

PR #2831, opened 2026-08-02
2809 fn:deep-equal, debug option: nondeterminism

### LINKED ISSUE #2809: fn:deep-equal, debug option
If the `debug` option is enabled, `fn:deep-equal` should be treated as nondeterministic function.
Otherwise, debugging may be skipped.

ChristianGruen:
Closes #2809

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -17560,6 +17560,13 @@ else error(#Q{http://www.w3.org/2005/xqt-errors}FORG0005))
          <note><p>As a general rule for boolean options (but not invariably), the value <code>true</code> indicates 
             that the comparison is more strict. </p></note>
 
+         <p diff="add" at="issue2809">The determinism of this function depends on the value of the
+            <code>debug</code> option. If the option is <code>false</code> (the default), the function
+            is <termref def="dt-deterministic">deterministic</termref>. If it is <code>true</code>, the
+            function is <termref def="dt-nondeterministic">nondeterministic</termref>, on the grounds
+            that the diagnostics it outputs are a side effect, even though this side effect has no
+            consequences on the result of the function.</p>
+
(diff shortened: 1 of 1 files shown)


michaelhkay:
I'm a little uneasy about this change. We've defined "deterministic" as being a property of a function, not a property of a particular run-time evaluation of a function call. The new text is therefore something of a fudge.

In addition, I can't see that the rules for calling nondeterministic functions are really appropriate here. I can't see, for example, that an optimizer would want to avoid loop-lifting a call to deep-equal() just because the debug option is set (or because it isn't known statically whether it is set.)

ChristianGruen:
This came up in an existing use case where the values to be compared were provided by external variables. As our optimizer decided to evaluate the code at compile time, the debug info was never output.

michaelhkay:
We need to ask, what actual effect does it have to say that a function is nondeterministic?

The definition is:

> [Definition: A function that is guaranteed to produce [identical] results from repeated calls within a single [execution scope] if the explicit and implicit arguments are identical is referred to as deterministic.]

On that definition, I think that trace(), message(), and deep-equal() with debug="true" are deterministic. The diagnostic output to my mind is not part of the result of the function, any more than its observable execution time is part of its result. (Resource access functions with stable="no" fall into a different category...)

Optimisers might like to take account of the special nature of these functions to improve usability, but I don't think the spec needs to get into that, it's a "quality of implementation" issue.

ChristianGruen:
`fn:trace` became nondeterministic through #2560, with the note explaining that the output “can be regarded as a side effect” and that the classification exists to encourage optimizer restraint. `fn:deep-equal` with debug set is the same case.

As you say, the definition does not reflect this yet. We could change it to:

> [Definition: A function is deterministic if a call on the function within a single execution scope can be replaced by the result of another call with identical explicit and implicit arguments, without changing the observable behavior of the query or transformation. Repeated calls with identical arguments are guaranteed to produce identical results.]

ChristianGruen:
I have added the `Revise` label: We should resolve #2848 first.

PR #2832, opened 2026-08-02
2808 strip-space option: fn:doc vs. fn:parse-xml

### LINKED ISSUE #2808: strip-space option: fn:doc vs. fn:parse-xml
* The function `fn:doc` has the option `strip-space` with the enum values `all`, `none`, and `conditional`.
* The functions `fn:parse-xml` and `fn:parse-xml-fragment` have an option with the same name, but a Boolean value.

Proposal:
* unify on `enum("all", "none", "conditional")` for all four (including `fn:doc-available`), defining `conditional` as `none` for `fn:parse-xml(-fragment)`.
* If we prefer to keep different types, we should add an explanatory note.

ChristianGruen:
Closes #2808

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -20964,21 +20964,26 @@ return { "width": $int16-at($loc + 5),
                   </fos:value>
                </fos:values>
             </fos:option>
-            <fos:option key="strip-space">
+            <fos:option key="strip-space" diff="chg" at="issue2808">
                <fos:meaning>Determines whether whitespace-only text nodes are removed
-                  from the resulting document. (Note: in XSLT, the <code>xsl:strip-space</code>
-                  and <code>xsl:preserve-space</code> declarations are ignored.)</fos:meaning>
-               <fos:type>xs:boolean</fos:type>
-               <fos:default>false</fos:default>
+                  from the resulting document.</fos:meaning>
+               <fos:type>enum("all", "none", "conditional")</fos:type>
+               <fos:default>"none"</fos:default>
                <fos:values>
-                  <fos:value value="true">All whitespace-only text nodes are stripped,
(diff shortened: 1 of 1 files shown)


specifications/xpath-functions-40/src/function-catalog.xml
```diff
@@ -20964,22 +20964,27 @@ return { "width": $int16-at($loc + 5),
...
                   or (b) XSD validation identifies that the parent element has a simple type or a complex
                   type with simple content.
                   </fos:value>
-                  <fos:value value="false">All whitespace-only text nodes are preserved,
+                  <fos:value value="none">All whitespace-only text nodes are preserved,
                   unless either (a) DTD validation marks them as ignorable, or (b) XSD validation recognizes
                   the containing element as having element-only or empty content.
                   </fos:value>
+                  <fos:value value="conditional">Equivalent to <code>"none"</code>. Unlike the
+                     <function>fn:doc</function> function, this function is not subject to external
```
Arithmeticus, line 20983: I am confused. The placement of this proviso makes me think that I will learn what differentiates `none` from `conditional` but the prose that follows seems to hold true for both values. Or maybe not. Because it refers to "this function" which makes me think that it applies more generally to `strip-space` or the function as a whole. I think either the note needs to be moved somewhere else, or the wording needs to be adjusted. Or maybe I have misread the material.

Arithmeticus (commented):
Minor suggestion. I am ambivalent about whether to go in this direction or retain the inconsistency. I look forward to discussion.

specifications/xpath-functions-40/src/function-catalog.xml
ChristianGruen: Thanks. Maybe we could also restrict the enum to `all` and `none`.

In general, it seems confusing to have an option which is typed as `xs:boolean` for one function and `xs:string` for a closely related function.

PR #2833, opened 2026-08-02
2263 New named record types

### LINKED ISSUE #2263: Named record types: TODOs
Resulting from the conversation in #2101:

1. The following records have not been named yet:

```xquery
(: return types :)
fn:divide-decimals
fn:xsd-validator
array:of-members
array:members

(: for parameter types as well? :)
fn:sort-by
```

2. We should add at least one example that demonstrates the usage of a built-in-record constructor.

3. @line-o suggested we could have records for all options.

ChristianGruen:
Closes #2263

Various of the comments in the issue were outdated. Worth discussing:

* Do we want record types for parameter types (specifically, `fn:sort-by` and `array:sort-by`)?
* Do we want to keep the rule alive that “all built-in record types live in the `fn` namespace”? Otherwise, we could change `fn:array-sort-key-record` to `array:sort-key-record`.
* I have skipped `bin:infer-encoding` as I didn’t know how to extend the record infrastructure to the EXPath modules.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -703,6 +703,100 @@
          </fos:meaning>
       </fos:field>
    </fos:record-type>
+
+   <fos:record-type id="division-record" extensible="false" diff="add" at="issue2263">
+      <fos:summary>
+         <p>This record type is used to represent the result of a call on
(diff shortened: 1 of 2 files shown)


ndw:
Hi Christian, can you fix whatever the conflicts are and then merge this, please?

PR #2851, opened 2026-08-20
2703 Dynamic selectors: consistent error handling

### LINKED ISSUE #2703: Dynamic selectors: consistent error handling
The enclosed-expression selector `axis::{ E }` coerces `E` to `(xs:QName | xs:string)*` for XNodes, but to `xs:anyAtomicType*` for JNodes, where any non-matching key simply selects nothing. This asymmetry means the same selector `$x/child::{ $keys }` raises a type error when `$x` is an element, yet silently matches nothing when `$x` is a map or array.

I think we should handle the XNode case leniently as well: atomic items in the selector value that are not `xs:QName` or `xs:string` values are simply ignored (they match nothing), as was already stated in a note for the former `get()` selector. This keeps the two node kinds parallel and avoids surprising, kind-dependent failures when the key sequence is computed generically. Raising no error would also live up to the XPath 1.0 tradition.

If we want to keep the current error handling, we should also reject other cases in which the resulting selector makes no sense:

```xquery
$node/child::{ 1 }      (: ↯ :) 
$node/child::{ "x y" }  (: ✓ currently accepted, even though the string is an invalid local name :) 
$array/child::{ x }     (: ✓ currently accepted, even though arrays cannot have string keys :) 
```

ChristianGruen:
Closes #2703

### DIFF
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -13264,26 +13264,29 @@ return if (every $r in $R satisfies $r instance of gnode())
                   <code>{E}</code>
                   where <var>E</var> is an <nt def="Expr">Expr</nt>.
                The contained expression <var>E</var> is evaluated with an absent focus and coerced
-                  to a required type of <code>(xs:QName | xs:string)*</code>;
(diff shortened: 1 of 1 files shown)


ndw:

PR #2854, opened 2026-08-20
2807 Dynamic node tests: avoid static errors

### LINKED ISSUE #2807: Dynamic node tests: static errors?
Copied from https://github.com/qt4cg/qtspecs/pull/2712#pullrequestreview-4615985800, something we should at least discuss:

--------

The rules of this PR combined with #2734 have the potential to break ordinary XML paths. A parser may rewrite `<xml/>/data()` to:

```xquery
<xml/>/(
  if (. instance of node()) then name() else (
    if (. instance of node()) then (
      let $A as ... := fn(){ data() }() return ...
    ) else (
      let $A as ... := fn(){ data() }() return ...
  )
)
```

As `XPTY0004` can be raised statically, `fn(){ data() }` may result in a static error, and the full query will be rejected although the paths in question would never be executed.

Maybe we should stress that the presented equivalent code is to be understood as code that cannot be simply pasted by the parser (?). Or we simply drop the equivalent representation and limit ourselves to formal rules.

ChristianGruen:
Closes #2807


michaelhkay:
In 4.7.3 you use the phrase "the step" without explanation to mean E2. I think it would be clearer to use "E2".

ChristianGruen:
@ndw One more issue I may not be able to fix by myself (sorry): https://github.com/qt4cg/qtspecs/actions/runs/32813250050/job/97701183125

ndw:

PR #2855, opened 2026-08-20
2818 Generalized casts: wildcards, error codes

### LINKED ISSUE #2818: Generalized casts: wildcards, error codes
#### 1. Wildcards

The current spec disallows casts to `array(*)`, `map(*)`, and `record(*)`:

```xquery
[ 1 ] cast as array(item()*)  (: ✓ :)
[ 1 ] cast as array(*)        (: ↯ :)
```

I would suggest also making the wildcard casts legal, in analogy with currently valid `instance of` checks.

#### 2. `XPTY0155` → `XPTY0004`

If I read the spec correctly, different error codes are to be raised, depending on the cast target:

```xquery
1 cast as map(xs:string, xs:integer)  (: XPTY0004 :)
1 cast as record(a)                   (: XPTY0155 :)
```

To avoid special-casing, I would prefer to always get `XPTY0004`.

…both resulting from PR #2777.

ChristianGruen:
Closes #2818


ndw:

PR #2856, opened 2026-08-20
2819 A step too far: notes

### LINKED ISSUE #2819: A step too far: notes
@michaelhkay A little summary of the pending comments from #2734:

1. “If S1 is neither a sequence of XNodes nor a sequence of JNodes” → “If any item in S1 is not a GNode”
2. `descendant-or-self::*` → `descendant-or-self::gnode()`
3. Add `def="dt-non-trivial"` for `<nt def="IfExpr">IfExpr</nt>` and subsequent expressions
4. Add `try/catch/finally`
5. Add `otherwise`
6. Add absolute path expressions? like `E1/(/a)`
7. …and three comments from @Arithmeticus, https://github.com/qt4cg/qtspecs/pull/2734#pullrequestreview-4640744699

ChristianGruen:
Closes #2819

### DIFF
--- a/specifications/xquery-40/src/expressions.xml
+++ b/specifications/xquery-40/src/expressions.xml
@@ -131,8 +131,8 @@ declaration contains construction of new nodes.)</phrase>
                but it is not a <term>range expression</term> under this definition.</p>
             
             <p><termdef id="dt-non-trivial" term="non-trivial">A construct is said to be a
-            <term>non-trivial</term> instance of a grammatical production if it is not also
-            an instance of one of its sub-productions.</termdef></p>
+            <term>non-trivial</term> instance of a grammatical production if it uses at least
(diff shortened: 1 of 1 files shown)


michaelhkay:
In 4.7.3, applying the adjective "non-trivial" to constructs such as ForExpr, LetExpr, IfExpr, and ParenthesizedExpr in incorrect. The way "non-trivial" is defined, it only really applies to constructs with an infix operator.

ChristianGruen:
Thanks; fixed. Your comment let me think about (non-)triviality:
* In its definition, a “sub-production” was defined nowhere, so I changed “it is not also an instance of one of its sub-productions.” to “it uses at least one terminal symbol contributed by that production itself.”.
* I also removed *non-trivial* from `FLWORExpr`, `FilterExpr` and other constructs (which can never be *trivial*).

ndw:

PR #2857, opened 2026-08-20
2845 csv-to-xml: pseudo-code fixed

### LINKED ISSUE #2845: csv-to-xml: prose contradicts (broken) pseudo-code
The CSV rules seem to contradict the pseudo-code. In the pseudo-code, `<columns>` is emitted whenever `exists($colNames)`. The prose says…

> If non-empty column names are available for some columns but not for others, then (a) the empty column element is included within the `columns` element if and only if there is a subsequent column with a non-empty name, […]

The text was introduced with [be078f51](https://github.com/qt4cg/qtspecs/commit/be078f51) as a resolution for https://github.com/qt4cg/qtspecs/issues/1675.

I think the prose should win. The pseudo-code is broken anyway (`$colNames` is declared, `$colnames` is referenced; the rule from above is not implemented at all), perhaps:

```xquery
let $parsed := parse-csv($value, $options)
let $names  := $parsed?columns
let $last   := foot(index-where($names, fn { . != '' }))
return document {
  <csv xmlns="http://www.w3.org/2005/xpath-functions"> {
    if (exists($last)) {
      <columns>{ subsequence($names, 1, $last) ! <column>{ . }</column> }</columns>
    },
    <rows>{
      for $row in $parsed?rows
      return <row>{
        for member $field at $col in $row
        return <field>{
          if ($names[$col] != '') { attribute column { $names[$col] } },
          $field
        }</field>
      }</row>
    }</rows>
  }</csv>
}

```

ChristianGruen:
Closes #2845

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -30208,24 +30208,27 @@ return csv-to-arrays(
          
          <eg><![CDATA[let $parsedCSV := parse-csv($value, $options)
 let $colNames := $parsedCSV?columns
+let $lastCol := foot(index-where($colNames, fn { . != '' }))
 return document {
    <csv xmlns="http://www.w3.org/2005/xpath-functions"> {
-     if (exists($colNames)) {
-       <columns>{ $colNames ! <column>{ . }</column> }</columns>
+     if (exists($lastCol)) {
+       <columns>{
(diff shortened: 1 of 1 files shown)


ndw:
At [meeting 177](https://qt4cg.org/meeting/minutes/2026/08-25.html), the CG agreed to merge this PR without discussion.

PR #2868, opened 2026-08-26
2867 fn:current everywhere

### LINKED ISSUE #2867: `fn:current` everywhere
With #2858, the custom syntax `context value` was introduced for defaults of function parameters.

It may be difficult for users to understand the difference between `.` and `context value`. Second, `context value` can only be used on top level (we cannot write `data(context value)`.

If a path step or a predicate changes the focus, XQuery gives you no way back to the value the query was started with; you have to remember to bind it to a variable instead, which is often a bit tedious, even more with the new pipeline operator `->`. This is why I think we should make `fn:current` available everywhere, and not restrict it to XSLT anymore.

ChristianGruen:
Closes #2867


michaelhkay:
Should add a note to the backwards compatibility appendix of XSLT to say that the error code for "no context value" has changed.

michaelhkay:
I think the right way to specify this (and I'm surprised to discover XSLT doesn't already do this) is to define a new component in the dynamic context, called perhaps the "current value" or the "preserved context value"; define the current() function to return the value of this component, and say that setting the value of the component is host-language dependent. Then define rules in XPath, XQuery, and XSLT to say how the "current value" (or whatever we call it) is set initially and is changed by particular constructs.

In XSLT this would go in §5.3.

ChristianGruen:
Makes sense; I had similar thoughts. I have revised the prose: as usual, I am particularly interested in the assessment of the XSLT part: https://github.com/qt4cg/qtspecs/pull/2868/changes/32e1e66bcc6476cbd1b5f6af55817ff9dd33c194

specifications/xslt-40/src/xslt.xml
```diff
@@ -20096,9 +20111,10 @@ return $tree =?> depth()]]></eg>
                   value is accepted.</p>
                
                <p>The default value for an optional function parameter is evaluated in the same way as the initial value
-                  of a global variable declaration, with one exception: a call on <function>fn:current</function> within
-                  the <code>select</code> expression — but not within a contained sequence constructor — is interpreted
-                  as a reference to the context value from the dynamic context of the function caller.
+                  of a global variable declaration, with one exception: while the <code>select</code> expression is
+                  evaluated, the <xtermref spec="XP40" ref="dt-current-value"/> is set to the context value from the
+                  dynamic context of the function caller, and a call on <function>fn:current</function> returns it.
+                  This does not apply within a contained sequence constructor, which is evaluated by the function itself.
```
michaelhkay: I suggest: "This does not apply within a contained sequence constructor, which follows the usual rule that for any XPath expression evaluated within the sequence constructor, the current value is set to the context value at that point, which will typically be the value of the global context item."

michaelhkay (commented):
Fine except for the suggested change in XSLT

michaelhkay:
I think that the way you have defined it, in XQuery, the current() function can be called within a function body (of either a global function or an inline function) to refer to the global context value. Is that intended, and is it a good idea? And if so, should we change XSLT to work that way? Currently in XSLT, a function call sets current() to absent just as it sets "." to absent.

ChristianGruen:
> Is that intended, and is it a good idea?

Yes, it was inteded. It seemed a good idea to me because we can also access global variables in function bodies.

johnlumley:
The example 

> A simple map expression does not affect `fn:current` either:
> ```  
>     item ! current()/@default-category
> ```
> The expression returns two attributes, one for each item.


should perhaps be more

> A simple map expression does not affect `fn:current` either:
> ```  
>        item ! current()/@default-category
> ```
> The expression returns a sequence of two identical attributes (`@default-category[.='fiction']`), one for each `item`. Note that these are the same attribute from the `catalog` element - in some circumstances, such as applying the union operator, duplicates would be eliminated.

ChristianGruen:
@johnlumley Thanks – I‘ve incorporated your wording.

ndw:
At [meeting 180](https://qt4cg.org/meeting/minutes/2026/09-15.html), the CG agreed to merge this PR without discussion.

PR #2877, opened 2026-08-27
2827 Editorial Notes (August 2026 edition)

### LINKED ISSUE #2827: Editorial Notes (August 2026 edition)
`fn:map-to-element`:
 
* [x] example: `dates` → `years`
* [x] Value conversions: item 5 contradicts the `xml` layout. It says the string “is parsed as an XML fragment and the resulting node is used as the element content”. But the forward layout serializes the element itself. It should say instead that the parsed element is the result, and that the map key is then redundant.

ChristianGruen:
Closes #2827


ndw:
At [meeting 178](https://qt4cg.org/meeting/minutes/2026/09-01.html), the CG agreed to merge this PR without discussion.

Issue #2873, opened 2026-08-27
`fn:update`: another attempt


ChristianGruen:
I thought I’d give the pending JNode update issue another attempt (@joewiz’s #77 is the oldest one that’s still unresolved). This one tries to…

* define a single XPath update function;
* work for both JNodes and XNodes; and
* allow recursive updates, to for example rename all keys `a` as `b`.

The signature:

```
fn:update(
  $input        as gnode(),
  $select       as fn($node as gnode()) as gnode()*,
  $replacement  as item()* := ()
) as item()*
```

* `$select` is applied once to `$input` and returns the nodes to replace. They must belong to `$input`.
* `$replacement` is what a selected node becomes. If it is a function item, it is called with the node.
* Replacements are applied bottom-up, so a node reaches `$replacement` with its descendants already updated.
* A map or array supplied where `gnode()` is expected becomes a root JNode.
* A root node in substitution position contributes its children, exactly as a document node does in content position.
* Inside an array, the key of a substituted entry is ignored, since positions are renumbered.

The replacement, or the result of the replacement function:

Result | Description
--- | ---
`()` | delete the node (default)
one atomic item | replace the node’s value, keeping its key or name
one node, map or array | replace the node, key or name included
several items | replace and insert, each item read as above

An atomic item replaces the value, everything else replaces the node. To assign a map, an array or a sequence as a value, name the key: `{ jkey($n): [] }`.

### Input

```xquery
let $store := {
  'store': {
    'books': [
      { 'title': 'XQuery', 'price': 39.95 },
      { 'title': 'JSON',   'price': 24.50, 'draft': true() }
    ],
    'open': true()
  }
}
```

#### Delete `draft` nodes

```xquery
update($store, fn { .//draft })
```

#### Increase a value

```xquery
update($store, fn { .//price }, fn { . * 2 })
```

#### Insert a sibling entry

```xquery
update($store, fn { .//price }, fn { ., { 'currency': 'EUR' } })
```

#### Rename `books` recursively

Nested occurrences are covered: the inner nodes are replaced first, so `jvalue($n)` already carries the renamed subtree.

```xquery
update($store, fn($n) { $n//books }, fn($n) { { 'items': jvalue($n) } })


```

#### Add a book

```xquery
update(
  $store,
  fn($n) { $n//books },
  fn($n) { { 'books': array:append(jvalue($n), { 'title': 'Update', 'price': 9.95 }) } }
)
```

#### Single-pass update

```xquery
update(
[...]

PR #2881, opened 2026-08-31
2873 fn:update: another attempt

ChristianGruen:
I deliberately ignored Sequence JNodes and the ·jposition· property. I will look at it once #2878 is done.

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -15864,21 +15864,18 @@ else $node
          <fos:example>
             <fos:test>
                <fos:expression><eg>
-update(
-  { 'title': 'Sojourner Truth', 'draft': true() },
-  fn { .//draft }
-) => jvalue()
+{ 'title': 'Sojourner Truth', 'draft': true() }
+=> update(fn { .//draft })
+=> jvalue()
                </eg></fos:expression>
                <fos:result>{ "title": "Sojourner Truth" }</fos:result>
                <fos:postamble>Selected nodes are deleted if no replacement is supplied.</fos:postamble>
             </fos:test>
             <fos:test>
                <fos:expression><eg>
-update(
-  { 'title': 'Ida Wells', 'price': 39 },
-  fn { .//price },
-  fn { . * 2 }
-) => jvalue()
+{ 'title': 'Ida Wells', 'price': 39 }
+=> update(fn { .//price }, fn { . * 2 })
+=> jvalue()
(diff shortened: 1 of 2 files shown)


michaelhkay:
A couple of specific comments from today's discussion, before I forget them:

(a) Type annotations on XNodes: the simplest approach is probably to say that all new nodes are untyped. But one might do better by saying that type annotations are retained on nodes if their (recursive) content doesn't change in any way.

(b) Namespace fixup: operations like renaming create a need for namespace fixup. I think they also create a need (which already exists....) for dropping namespace nodes, and I wonder if we therefore need to put the namespace axis back on the XQuery agenda?

(c) In my earlier attempts to define this, I hit problems of formalisation because of node identity. I think you've dodged these issues by describing things fairly informally. For example, if you make a change to the content of an array, then you end up with a new array, but since you're applying updates bottom-up, you need to recognize that this new array is a replacement for one which might itself have been selected for updating. Similarly, if two subtrees of a single array are both updated, these updates need to be merged into one new array.

There may be a case for two levels of description here, one relatively informal and one much more formal. The formal description might perhaps be in terms of equivalent XQuery or XSLT constructs?

ChristianGruen:
I have reworked the rules, and I hope they are better accessible now.

> (a) Type annotations on XNodes: […]

Done: an XNode now keeps its type annotation if neither it nor any of its attributes or descendants was selected. That matches XQUF (with revalidation skip). Node identity stays strict, because once anything is replaced the ancestors are rebuilt and everything below them has to be copied anyway.

> (b) Namespace fixup: operations like renaming create a need for namespace fixup. […]

The function has no rename primitive: renaming an element, and dropping a declaration, are both done by the caller with an ordinary constructor, which already carries fixup and does not carry unused in-scope namespaces over. So I would not reopen the namespace axis for this.

> (c) In my earlier attempts to define this, I hit problems of formalisation because of node identity. I think you've dodged these issues by describing things fairly informally. […]

Both problems come from formulating updates as a set of primitives. Here `$select` is called once, before any node is constructed, and each node is processed once with all of its children, so nothing is re-selected and there is nothing to merge. I have added a sentence saying this.

> There may be a case for two levels of description here, one relatively informal and one much more formal. The formal description might perhaps be in terms of equivalent XQuery or XSLT constructs?

The pseudo-code can be found at https://github.com/qt4cg/qtspecs/issues/2873#issuecomment-5475472203; I think it’s fairly close to what the rules say. Unfortunately, it’s far from compact, so I tend to believe it’s better to not add it to the spec.

PR #2886, opened 2026-09-02
1459 Function properties and arities

### LINKED ISSUE #1459: Function properties and arities

ChristianGruen:
Summary of the most important changes:

* I have removed the explicit listing of the default function properties both in the code as well as in the rendered representation (deterministic, context-independent, focus-independent); they are stated once in a new sectin §1.5.4.
* The arity prose has been changed to a single sentence naming the omitted argument, for example: “If `$node` is absent, this function is context-dependent and focus-dependent.”
* Various wrong statements were fixed (`fn:format-integer`, `fn:format-number`, `fn:civil-timezone`, fn:contains-subsequence`, `fn:load-xquery-module`, `fn:in-scope-prefixes`, and others).
* Functions like `fn:transform` now name the component they depend on.

Closes #1459


ChristianGruen:
…ready to merge

ndw:

PR #2903, opened 2026-09-11
2901 element-to-map: errors in the conversion plan

### LINKED ISSUE #2901: element-to-map: errors in the conversion plan
F+O §14.6.5, "Structure of the conversion plan" makes some statements using "must" (without special markup), for example

> the corresponding value must be of type fn:attribute-conversion-plan-record

and makes other statements of the form:

> The layout entry is always non-empty.

It's not clear how violations of these constraints are to be handled. There's a general rule:

> Any error in the conversion plan is treated as a type error [[err:XPTY0004](https://www.w3.org/TR/xpath20/#ERRXPTY0004)]XP

But it's not 100% clear whether violations of these constraints are errors in this sense; and this is particularly true when there is a fallback layout.

There are two cases where non-conformance with the rules is explicitly tolerated:

> Any entries whose keys are not in this format will be ignored.

> If additional entries (beyond those described above) are present in any of the maps, they are ignored, provided that the map is coercible to the given type definition.

Does "entries (beyond those described above)" include entries that violate the rules described above, and thus imply that entries violating the rules should be ignored?

Clarification is needed.

ChristianGruen:
Closes #2901

### DIFF
--- a/specifications/xpath-functions-40/src/function-catalog.xml
+++ b/specifications/xpath-functions-40/src/function-catalog.xml
@@ -116,7 +116,7 @@
             <p>Defines whether attributes should be represented
                as numbers or booleans instead of the default representation as strings;
                alternatively, indicates that the attribute should be omitted.
-               The field may be empty, which is equivalent to <code>string</code>.</p>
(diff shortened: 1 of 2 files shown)


ndw:

PR #2905, opened 2026-09-11
2900 Map to Record Coercion

### LINKED ISSUE #2900: Map to Record Coercion
I know this has been raised before, but I can't find an open issue on it.

In the rules for coercing a map to a record type, the Notes say:

> But if the supplied value is { "latitude": 53.2, "longitude": 0, "altitude": 3_500 }, then a type error occurs because the field "altitude" is not defined in the record type.

However, there is nothing in the normative rules to justify this.

In my view we should retain the rules as they are. But I know there is an alternative view that coercion should fail if the map contains excess entries. Either view is workable; they have different pros and cons. I'm inclined to the current rules partly because they are easier to specify, easier and more efficient to implement. However, now that users have the option of casting the map to the record type (which does drop unwanted entries), there's a workaround when that's the behaviour that users want.

ChristianGruen:
…like this?

Closes #2900


ndw:

PR #2910, opened 2026-09-14
2909 Binary functions: Base64 URL en-/decoding

### LINKED ISSUE #2909: Binary functions: Base64 URL en-/decoding
The encoding behind JWT or PKCE is called `base64url` (see [RFC 4648, §5](https://datatracker.ietf.org/doc/html/rfc4648)), and our languages offer no way to produce or consume it: `xs:base64Binary` only speaks the standard alphabet with mandatory padding, so every implementation of these protocols hand-rolls translate and replace chains.

Two functions, `bin:from-base64url` and `bin:to-base64url`, would close that gap alongside `bin:hex`, `bin:octal` and `bin:bin`, which already cover the other textual octet encodings.

ChristianGruen:
Closes #2909


ndw:
At [meeting 181](https://qt4cg.org/meeting/minutes/2026/09-22.html), the CG agreed to merge this PR.

